PR #5, opened 2020-09-14
Add Batch step

talIguaz:


storey/flow.py
```diff
@@ -649,6 +649,47 @@ async def _do(self, event):
...
+                await self.emit_batch()
+
+    async def emit_batch(self):
+        if len(self._batch) > 0:
+            await self._do_downstream(Event([self._get_safe_event_or_body(event) for event in self._batch], time=self._batch[0].time))
+            self._batch = []
+            self._events_count = 0
+
+            if self._timeout_task and not self._timeout_task.cancelled():
+                self._timeout_task.cancel()
```
gtopper: I think you have to cancel it before `self. _do_downstream` in order to avoid a race condition where emit_batch is executed twice and emits duplicate batches.

talIguaz: hmm i guess i can do that, although i do check if there is something to emit right before i emit but fine

storey/flow.py
```diff
@@ -649,6 +649,47 @@ async def _do(self, event):
...
+            return await self._do_downstream(_termination_obj)
+        else:
+            if len(self._batch) == 0 and self._timeout_secs:
+                self._timeout_task = asyncio.get_running_loop().create_task(self.sleep_and_emit())
+
+            self._events_count = self._events_count + 1
+            self._batch.append(event)
+
+            if self._events_count == self._max_events:
+                await self.emit_batch()
```
gtopper, line 681: It's better not to create the task at all if you're going to emit right away.

storey/flow.py
```diff
@@ -649,6 +649,47 @@ async def _do(self, event):
...
+
+    async def emit_batch(self):
+        if len(self._batch) > 0:
+            await self._do_downstream(Event([self._get_safe_event_or_body(event) for event in self._batch], time=self._batch[0].time))
+            self._batch = []
+            self._events_count = 0
```
gtopper: I think these need to precede the await as well, or you risk another `emit_batch` being triggered before this method resumes.

tests/test_flow.py
```diff
@@ -338,3 +340,54 @@ def mapf(x):
     assert event.body == 'original body'
     assert result.key == 'new key'
     assert result.body == 'new body'
+
+
+def test_batch():
+    controller = build_flow([
+        Source(),
+        Batch(4, 5),
```
gtopper: If you don't want timeout to be reached, set it very high. 5s is just low enough that it might happen under some strange circumstances (e.g. in a highly virtualized environment).

tests/test_flow.py
```diff
@@ -338,3 +340,54 @@ def mapf(x):
...
+def test_batch_with_timeout():
+    controller = build_flow([
+        Source(),
+        Batch(4, 2),
+        Reduce([], lambda acc, x: append_and_return(acc, x)),
+    ]).run()
+
+    for i in range(10):
+        if i == 3:
+            time.sleep(3)
```
gtopper, line 389: Would be safer to use `Complete` here instead of sleep.

storey/flow.py
```diff
@@ -649,6 +649,47 @@ async def _do(self, event):
...
+        if self._timeout_secs <= 0:
+            raise ValueError('Batch timeout cannot be 0 or negative')
+
+    async def sleep_and_emit(self):
+        await asyncio.sleep(self._timeout_secs)
+        await self.emit_batch()
+
+    async def _do(self, event):
+        if event is _termination_obj:
+            self._terminate_worker = True
```
gtopper: This field is not used.

storey/flow.py
gtopper: But you reset `self._batch` after the async def call, where the event loop can pick up another task.

storey/flow.py
gtopper: Also, `_events_count` -> `_event_count`

tests/test_flow.py
talIguaz: `Complete` does not work with `Batch`. Topper will add a trello ticket

gtopper (approved):

PR #10, opened 2020-09-17
Use v3io-py for v3io calls.

gtopper:


storey/flow.py
```diff
@@ -1015,50 +973,63 @@ def _get_attributes_as_blob(self, aggregation_element):
...
+        if response.status_code != 200:
+            raise V3ioError(f'Failed to get number of shards. Got {response.status_code} response: {response.body}')
+        return response.output
+
+    async def _put_records(self, container, stream_path, payload):
+        self._lazy_init()
+
+        return await self._v3io_client.stream.put_records(container, stream_path, payload)
+
+    async def _get_item(self, container, stream_path, key, attributes):
```
talIguaz: `stream_path` -> `table_path`

storey/flow.py
```diff
@@ -1015,50 +973,63 @@ def _get_attributes_as_blob(self, aggregation_element):
...
+            parsed_response = response.output.item
+            for name in parsed_response.items():
                 if name.startswith(self._aggregation_attribute_prefix):
                     del parsed_response[name]
             return parsed_response
         else:
-            raise V3ioError(f'Failed to get item. Response status code was {response.status}: {body}')
+            raise V3ioError(f'Failed to get item. Response status code was {response.status_code}: {response.body}')
+
+    async def _describe(self, container, stream_path):
```
talIguaz, line 969: Why half of the functions accept `container` and `path` and some a combined `path`?

gtopper: Because once again we're forcecd to make this separation 😿

storey/flow.py
```diff
@@ -941,27 +942,44 @@ async def _load_schema(self, table_path):
...
+    def _build_update_expression(self, aggregation_element, additional_data):
+        expressions = []
         for name, bucket in aggregation_element.aggregation_buckets.items():
             # Only save raw aggregates, not virtual
             if bucket.should_persist:
                 blob = pickle.dumps(bucket.to_dict())
-                data[f'{self._aggregation_attribute_prefix}{name}'] = blob
-
-        return data
+                base64_blob = base64.b64encode(blob).decode('ascii')
```
talIguaz, line 981: Shouldn't that be without the base64encode?

storey/flow.py
talIguaz: Why is that? it's redundant to split the path in the "calling" flows if it's only used here anyways?
Or alternatively if you want you can split that path always outside the driver.. no?

talIguaz (approved):

integration/integration_test_utils.py
```diff
@@ -9,12 +10,76 @@
 import string
 import pytest
 
-from storey import NeedsV3ioAccess
 from storey.flow import V3ioError
 
 _non_int_char_pattern = re.compile(r"[^-0-9]")
 
 
+class V3ioHeaders:
```
talIguaz: Shouldn't it inherit from `NeedV3ioAccess` to not duplicate code

PR #15, opened 2020-10-07
Add docstrings.

gtopper:


storey/flow.py
```diff
@@ -78,6 +78,17 @@ def _user_fn_output_to_event(self, event, fn_result):
 
 
 class Choice(Flow):
+    """Redirects each input element into at most one of multiple downstreams.
+
+    :param choice_array: a list of (downstream, condition) tuples, where downstream is a step and condition is a function. The first
+    condition in the list to evaluate as true for an input element causes that element to be redirected to that downstream step.
+    :type choice_array: tuple of (Flow, Function (Event=boolean))
```
talIguaz: `Event=boolean` ?
I think the syntax for a function type should be something like `callable[[args..], result]`

storey/flow.py
```diff
@@ -105,6 +116,26 @@ async def _do(self, event):
 
 
 class Event:
+    """The basic unit of data in storey. All steps receive and emit events.
+
+    :param body the event payload, or data
```
talIguaz: body:
Also maybe explicitly write that it's type is "Any"

storey/flow.py
gtopper: Was meant to be => like everywhere else.

storey/flow.py
```diff
@@ -105,6 +116,26 @@ async def _do(self, event):
...
+    :param id: Event identifier.
+    :type id: string
+    :param key: Event key.
```
talIguaz: Explain more about the logical difference and the usage difference between `id` and `key`.
Also, i'd put `key` first since it's more "interesting" in a flow

gtopper: It matches the order of the constructor params.

storey/flow.py
gtopper: `object` in Python

storey/flow.py
talIguaz: From a user perspective, "Event key" and "Event identifier" are the same thing.. you have to explain what each of the does and how "we" use each of them.
regarding the order.. well in my opinion better order by significance, but as long as you explain both it's fine. BTW i would change the constructor params order to have key first from the same reasons...

storey/flow.py
```diff
@@ -132,11 +163,28 @@ def _set_error(self, element):
...
+    """
+
     def __init__(self, emit_fn, await_termination_fn):
         self._emit_fn = emit_fn
         self._await_termination_fn = await_termination_fn
 
     def emit(self, element, key=None, event_time=None, return_awaitable_result=False):
+        """Emits an event into the associated flow.
+
+        :param element: The event data, or payload.
```
talIguaz: worth specifying that you can also pass an Event object here

storey/flow.py
```diff
@@ -170,6 +220,14 @@ def await_termination(self):
 
 
 class Source(Flow):
+    """
+    Synchronous entry point into a flow. Produces a FlowController when run, for use from inside a synchronous context. See AsyncSource
+    for use from inside an async context.
```
talIguaz, line 229: Weird line break

storey/flow.py
```diff
@@ -238,11 +296,29 @@ async def _set_error(self, ex):
...
+        :param element: The event data, or payload.
+        :type element: object
+        :param key: The event key (optional)
+        :type key: string
+        :param event_time: The event time (default to current time, UTC).
+        :type event_time: datetime
+        :param await_result: Whether to await a result from the flow (as signaled by the Complete step). Defaults to False.
+        :type await_result: boolean
+
+        :returns: The result received from the from if await_result is True. None otherwise.
```
talIguaz: from the from

talIguaz:
You never document `full_event` and `termination_result_fn` which are passed as the `kw` args to every flow.
Not sure about the termination but the `full_event` indication should be documented somewhere i think

storey/flow.py
gtopper: Yeah, need to document it *everywhere* I guess...

talIguaz:
You are referring to `V3ioDriver` in your docs, but haven't documented this class.. so add docs to the driver

talIguaz (approved):
✅

PR #21, opened 2020-10-15
Add dataframe source.

gtopper:


storey/flow.py
```diff
@@ -47,6 +47,9 @@ def run(self):
         for outlet in self._outlets:
             outlet.run()
 
+    def run_async(self):
```
talIguaz: async def run_async?

storey/flow.py
```diff
@@ -147,15 +150,23 @@ class Event:
...
         self.path = path
         self.content_type = content_type
         self._awaitable_result = awaitable_result
 
+    def __eq__(self, other):
+        if not isinstance(other, Event):
+            return False
+
+        result = self.body == other.body and self.time == other.time and self.id == other.id and self.headers == other.headers
+        result = result and self.method == other.method and self.path == other.path and self.content_type == other.content_type
```
talIguaz, line 167: why break it to two statements just break the line

storey/flow.py
gtopper: 🙈

storey/flow.py
gtopper: Actually pycharm and the linter disagree on the indentation level in that case, so this is my workaround...

storey/flow.py
```diff
@@ -440,81 +496,97 @@ def __init__(self, paths, with_header=False, build_dict=False, key_field=None, t
...
+async def _aiter(iterable):
+    for x in iterable:
+        yield x
 
-    def run(self):
-        super().run()
 
-        thread = threading.Thread(target=self._loop_thread_main)
-        thread.start()
+class DataframeSource(FileSource):
```
talIguaz: if you use it here, not sure `FileSource` is the right terminology

storey/flow.py
```diff
@@ -440,81 +496,97 @@ def __init__(self, paths, with_header=False, build_dict=False, key_field=None, t
...
+        :param time_column: column to be used as time for events.
+        :type time_column: datetime
+        :param id_column: column to be used as ID for events.
+        :type id_column: string
+    """
 
-        def raise_error_or_return_termination_result():
-            self._raise_on_error(self._termination_q.get())
-            return self._termination_future.result()
+    def __init__(self, dfs, key_column=None, time_column=None, id_column=None, **kwargs):
```
talIguaz, line 565: isn't key and time are mandatory?
Even if `time_column` is optional ill set the time of the event to current time rather than `None`

storey/flow.py
```diff
@@ -440,81 +496,97 @@ def __init__(self, paths, with_header=False, build_dict=False, key_field=None, t
...
+        self._id_field = id_column
 
-        return FlowAwaiter(raise_error_or_return_termination_result)
+    async def _run_loop(self):
+        async for df in _aiter(self._dfs):
+            async for _, row in _aiter(df.iterrows()):
+                body = row.to_dict()
+                key = None
+                if self._key_field:
+                    key = body.pop(self._key_field, None)
```
talIguaz: I would raise KeyError rather than silencing it.
its a user error in he should be aware of that, he might have misspelled the column name or whatever

storey/flow.py
```diff
@@ -440,81 +496,97 @@ def __init__(self, paths, with_header=False, build_dict=False, key_field=None, t
...
+            dfs = [dfs]
+        self._dfs = dfs
+        self._key_field = key_column
+        self._time_field = time_column
+        self._id_field = id_column
 
-        return FlowAwaiter(raise_error_or_return_termination_result)
+    async def _run_loop(self):
+        async for df in _aiter(self._dfs):
+            async for _, row in _aiter(df.iterrows()):
```
talIguaz: Why are you ignoring indexes?
it is very possible that the `key_column` passed is already the index for the DF

tests/test_flow.py
```diff
@@ -118,6 +119,49 @@ def test_csv_reader_as_dict_no_header():
     assert termination_result == 21
 
 
+def test_dataframe_source():
```
talIguaz, line 122: Add tests with DFs containing indexes

storey/flow.py
gtopper: They're not mandatory. `time` will be set to the time of the event when you pass it `None`.

talIguaz (approved):

PR #32, opened 2020-10-29
Add WriteToParquet.

gtopper:


yaronha:
@gtopper i think its a good start, need to add the notion of partitioning by time as we discussed, also option to upload data into object like v3io/s3/.. (there is a method for it in mlrun)

PR #37, opened 2020-11-03
Parallel persist

talIguaz:
* create `_ConcurrentByKeyJobExecution` -> base class for concurrent execution across different keys
* add notion of `closeables` that will be "closed" on flow termination. every step can subscribe his associated closeables to be closed in the end of the flow
* several aggregation saving bugs that came up because of the above changes

### DIFF
--- a/storey/aggregations.py
+++ b/storey/aggregations.py
@@ -32,6 +32,7 @@ class AggregateByKey(Flow):
     def __init__(self, aggregates, table, key=None, emit_policy=_default_emit_policy, augmentation_fn=None):
         Flow.__init__(self)
         self._aggregates_store = AggregateStore(aggregates)
+        self._closeables = [table]
 
         self._table = table
         self._table._set_aggregation_store(self._aggregates_store)
@@ -123,11 +124,6 @@ class AggregateByKey(Flow):
             await self._emit_all_events(next_emit_time * 1000)
             next_emit_time = next_emit_time + seconds_to_sleep_between_emits
 
-    def run(self):
-        closeables = super().run()
-        closeables.append(self._table)
-        return closeables
-
 
 class QueryAggregationByKey(AggregateByKey):
     """
@@ -189,17 +185,13 @@ class Persist(_ConcurrentByKeyJobExecution):
     def __init__(self, table):
         super().__init__()
         self._table = table
+        self._closeables = [table]
 
     async def _process_event(self, event):
         return await self._table.persist_key(event.key)
 
     async def _handle_completed(self, event, response):
         await self._do_downstream(event)
-
-    def run(self):
-        closeables = super().run()
-        closeables.append(self._table)
-        return closeables
 
 
 class AggregatedStoreElement:
--- a/storey/aggregations.py
+++ b/storey/aggregations.py
@@ -124,9 +124,9 @@ class AggregateByKey(Flow):
             next_emit_time = next_emit_time + seconds_to_sleep_between_emits
 
     def run(self):
-        closables = super().run()
-        closables.append(self._cache)
-        return closables
+        closeables = super().run()
+        closeables.append(self._cache)
+        return closeables
 
 
 class QueryAggregationByKey(AggregateByKey):
@@ -197,9 +197,9 @@ class Persist(_ConcurrentByKeyJobExecution):
         await self._do_downstream(event)
 
     def run(self):
-        closables = super().run()
-        closables.append(self._cache)
-        return closables
+        closeables = super().run()
+        closeables.append(self._cache)
+        return closeables
 
 
 class AggregatedStoreElement:
--- a/storey/aggregations.py
+++ b/storey/aggregations.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from .aggregation_utils import is_raw_aggregate, get_virtual_aggregation_func, get_implied_aggregates, get_all_raw_aggregates, \
     get_all_raw_aggregates_with_hidden
 from .dtypes import EmitEveryEvent, FixedWindows, EmitAfterPeriod, EmitAfterWindow, EmitAfterMaxEvent
-from .flow import Flow, _termination_obj, Event
+from .flow import Flow, _termination_obj, Event, _ConcurrentByKeyJobExecution
 
 _default_emit_policy = EmitEveryEvent()
 
@@ -63,7 +63,6 @@ class AggregateByKey(Flow):
     async def _do(self, event):
         if event == _termination_obj:
             self._terminate_worker = True
-            await self._cache.close()
             return await self._do_downstream(_termination_obj)
 
         try:
@@ -88,7 +87,6 @@ class AggregateByKey(Flow):
                     await self._emit_event(key, event)
                     self._events_in_batch[key] = 0
         except Exception as ex:
-            await self._cache.close()
             raise ex
 
     # Emit a single event for the requested key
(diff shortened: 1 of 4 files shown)


storey/aggregations.py
```diff
@@ -125,6 +123,11 @@ async def _emit_worker(self):
             await self._emit_all_events(next_emit_time * 1000)
             next_emit_time = next_emit_time + seconds_to_sleep_between_emits
 
+    def run(self):
+        closeables = super().run()
+        closeables.append(self._cache)
+        return closeables
```
gtopper: Probably easier to have each step declare its closables and have run do the append.

storey/aggregations.py
```diff
@@ -189,14 +190,16 @@ def __init__(self, cache):
...
+    async def _process_event(self, event):
+        return await self._cache.persist_key(event.key)
+
+    async def _handle_completed(self, event, response):
+        await self._do_downstream(event)
+
+    def run(self):
+        closeables = super().run()
+        closeables.append(self._cache)
+        return closeables
```
gtopper: Probably easier to have each step declare its closables and have run do the append.

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+        self._q = None
+        self._pending_by_key = {}
+
+    async def _worker(self):
+        try:
+            while True:
+                job = await self._q.get()
+                if job is _termination_obj:
+                    for key, pending_event in self._pending_by_key.items():
+                        if self._pending_by_key[key].pending and not self._pending_by_key[key].in_flight:
```
gtopper: Isn't `self._pending_by_key[key]` the same as `pending_event`?

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+        else:
+            # Initializing the key with 2 lists. One for pending requests and one for requests that an update request has been issued for.
+            if event.key not in self._pending_by_key:
+                self._pending_by_key[event.key] = _PendingEvent()
+
+            # If there is a current update in flight for the key, add the event to the pending list. Otherwise update the key.
+            if len(self._pending_by_key[event.key].in_flight) > 0:
+                self._pending_by_key[event.key].pending.append(event)
+            else:
+                self._pending_by_key[event.key].pending.append(event)
```
gtopper: Shouldn't `self._pending_by_key[event.key].pending.append(event)` be outside the `if` if we do it either way?

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+
+    async def _worker(self):
+        try:
+            while True:
+                job = await self._q.get()
+                if job is _termination_obj:
+                    for key, pending_event in self._pending_by_key.items():
+                        if self._pending_by_key[key].pending and not self._pending_by_key[key].in_flight:
+                            resp = await self._process_event(pending_event.pending[0])
+                            for event in pending_event.pending:
```
gtopper, line 412: Are we ok with this being O(N)?

storey/aggregations.py
```diff
@@ -188,15 +185,13 @@ class Persist(Flow):
...
-            return await self._do_downstream(_termination_obj)
-        else:
-            # todo: persist keys in parallel
-            await self._table._persist_key(event.key)
-            await self._do_downstream(event)
+    async def _process_event(self, event):
+        return await self._table.persist_key(event.key)
+
+    async def _handle_completed(self, event, response):
+        await self._do_downstream(event)
```
gtopper, line 194: `response` is never used?

gtopper: I guess the idea is that is might be in some other impl?

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+                    await self._handle_completed(event, completed)
+                self._pending_by_key[event.key].in_flight = []
+
+                # If we got more pending events for the same key process them
+                if self._pending_by_key[event.key].pending:
+                    first_event = self._pending_by_key[event.key].pending[0]
+                    self._pending_by_key[event.key].in_flight = self._pending_by_key[event.key].pending
+                    self._pending_by_key[event.key].pending = []
+
+                    task = self._process_event(first_event)
```
gtopper, line 433: Don't this code only process the first pending event? But it moves all the events from pending to in_flight?

storey/drivers.py
```diff
@@ -191,19 +194,31 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
         update_expression = ';'.join(expressions)
         return update_expression, condition_expression, pending_updates
 
+    def _discard_old_pending_items(self, pendings, max_window_millis):
+        res = {}
+        if pendings:
```
gtopper: pendings?

storey/drivers.py
```diff
@@ -191,19 +194,31 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
...
+            last_time = int(list(pendings.keys())[-1] / max_window_millis) * max_window_millis
+            min_time = last_time - max_window_millis
+            for time, value in pendings.items():
+                if time > min_time:
+                    res[time] = value
```
gtopper, line 204: Consider deleting in place for performance?

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+                        break
+                    else:
+                        await self._q.put(_termination_obj)
+                        continue
+
+                event = job[0]
+                completed = await job[1]
+
+                for event in self._pending_by_key[event.key].in_flight:
+                    await self._handle_completed(event, completed)
```
gtopper, line 424: It looks like `in_flight` may change on us while we `await` here?

storey/flow.py
```diff
@@ -386,6 +388,104 @@ async def _do(self, event):
...
+                # If we got more pending events for the same key process them
+                if self._pending_by_key[event.key].pending:
+                    first_event = self._pending_by_key[event.key].pending[0]
+                    self._pending_by_key[event.key].in_flight = self._pending_by_key[event.key].pending
+                    self._pending_by_key[event.key].pending = []
+
+                    task = self._process_event(first_event)
+                    await self._q.put((event, asyncio.get_running_loop().create_task(task)))
+                else:
+                    del self._pending_by_key[event.key]
```
gtopper, line 436: Can't we end up here because `_do` reset `pending` (and set `in_flight` to its previous value) while we were awaiting?

storey/flow.py
talIguaz: the `_handle_completed`? how can it not be O(n) if we want to send each of them downstream

storey/aggregations.py
talIguaz: don't think there is any response for the `_do_downstream` other than on termination. which is handled elsewhere

storey/flow.py
talIguaz: well for the current implementation it's fine since the data i actually want to save is in the cache object.
But maybe for other implementations, other drivers might want the raw events in addition to the cached data.. 🤷‍♂️

storey/drivers.py
talIguaz: i can't delete something i iterating over 🤔

storey/flow.py
talIguaz: it shouldn't cause it's per key and only gets set when the in_flight is empty

gtopper (approved):

PR #42, opened 2020-11-04
Fuse ToDataFrame and Batch into parquet writer.

gtopper:


talIguaz:
Inherit from `_Batching`

storey/flow.py
```diff
@@ -426,57 +426,80 @@ async def _handle_completed(self, event, response):
...
+    async def _emit_batch(self):
         if len(self._batch) > 0:
             batch_to_emit = self._batch
+            batch_time = self._batch_time
             self._batch = []
+            self._batch_time = None
             self._event_count = 0
 
-            await self._do_downstream(Event([self._get_safe_event_or_body(event) for event in batch_to_emit], time=batch_to_emit[0].time))
+            await self._emit(batch_to_emit, batch_time)
```
talIguaz, line 480: why save the batch_time separately, what's wrong with `batch_to_emit[0].time`?

gtopper: The fast that `batch_to_emit[0]` is not an Event unless you pass in `full_event=True`.

talIguaz (approved):

PR #49, opened 2020-11-09
better handle of flow closeables

talIguaz:


storey/flow.py
```diff
@@ -194,6 +194,8 @@ def __init__(self, initial_state, fn, group_by_key=False, **kwargs):
         self._state = initial_state
         self._fn = fn
         self._group_by_key = group_by_key
+        if hasattr(initial_state, 'close'):
+            self._closeables = [initial_state]
```
gtopper, line 198: Shouldn't this be appended instead?

storey/sources.py
```diff
@@ -115,15 +115,17 @@ async def _run_loop(self):
...
             except BaseException as ex:
                 self._ex = ex
                 if not self._q.empty():
                     self._q.get()
                 self._termination_future.set_result(None)
                 break
+            finally:
+                if event is _termination_obj or self._ex:
+                    for closeable in self._closeables:
+                        await closeable.close()
```
gtopper: Hm, just put this after the loop?

storey/sources.py
```diff
@@ -261,6 +261,10 @@ async def _run_loop(self):
                 if not self._q.empty():
                     await self._q.get()
                 return None
+            finally:
+                if event is _termination_obj or self._ex:
+                    for closeable in self._closeables:
+                        await closeable.close()
```
gtopper, line 267: Hm, just put this after the loop?

storey/flow.py
talIguaz: as we discussed, each step sets his `_closeables` list and the Flow.run() is responsible on appending all the closeables into one list

gtopper: Ok, it's just that we already create a list in Flow.__init__ 🤷‍♂️

gtopper (approved):

PR #52, opened 2020-11-11
WriteToCSV API overhaul.

gtopper:


storey/writers.py
```diff
@@ -10,29 +10,38 @@
...
     :param path: path where CSV file will be written.
     :type path: string
-    :param event_to_line: function to transform an event to a CSV line (represented as a list).
-    :type event_to_line: Function (Event=>list of string)
-    :param header: a header for the output file.
-    :type header: list of string
-    :param name: Name of this step, as it should appear in logs. Defaults to class name (WriteCSV).
+    :param columns: fields to be written to CSV. Will be written as the file header if write_header is True. Will be extracted from
+    events when an event is a dictionary (lists will be written as is). Optional. Defaults to None (will be inferred if event is dictionary,
+    metadata_columns must be False).
```
talIguaz: why ` metadata_columns must be False`?

storey/writers.py
```diff
@@ -42,16 +51,30 @@ async def _do(self, event):
         try:
             if not self._open_file:
                 self._open_file = await aiofiles.open(self._path, mode='w')
+            data = event.body
+            if isinstance(data, dict):
+                if self._first_event and not self._columns:
+                    self._columns = list(data.keys())
+                    self._columns.sort()
+                if self._columns:
```
talIguaz, line 59: won't you always have columns by this line?

tests/test_flow.py
```diff
@@ -533,65 +534,104 @@ def test_batch_with_timeout():
...
+
+    expected = "n,n*10\n0,0\n1,10\n2,20\n3,30\n4,40\n5,50\n6,60\n7,70\n8,80\n9,90\n"
+    assert result == expected
+
 
+def test_write_csv(tmpdir):
+    asyncio.run(async_test_write_csv(tmpdir))
+
+
+async def async_test_write_csv_error(tmpdir):
```
talIguaz, line 561: nothing seems to run it

storey/writers.py
```diff
@@ -10,29 +10,38 @@
...
+    :type write_header: boolean
+    :param name: Name of this step, as it should appear in logs. Defaults to class name (WriteToCSV).
     :type name: string
     :param full_event: Whether user functions should receive and/or return Event objects (when True), or only the payload (when False).
     Defaults to False.
     :type full_event: boolean
     """
 
-    def __init__(self, path, event_to_line, header=None, **kwargs):
+    def __init__(self, path, columns=None, write_header=False, metadata_columns=None, **kwargs):
```
talIguaz, line 35: you don't use `metadata_columns` anywhere

talIguaz: Also add tests to these cases

storey/writers.py
gtopper: Not if we get mixed events (list/dict). Funky usage but it's up to users.

storey/writers.py
gtopper: Because metadata columns tell you which columns should get which metadata. If there are no known columns, you can't select which to fill with metadata.

gtopper: Discussed and decided to allow this use case and overwrite in case of naming conflict.

talIguaz (approved):

PR #59, opened 2020-11-17
support passing key as string to joinWithTable

talIguaz:


storey/flow.py
```diff
@@ -708,7 +708,13 @@ def __init__(self, table, key_extractor, attributes=None, join_function=None, **
         self._table = table
         self._closeables = [table]
 
-        self._key_extractor = key_extractor
+        if key_extractor:
```
gtopper, line 711: Need to update docstring.

gtopper (approved):

PR #61, opened 2020-11-18
Select columns to persist in write table

talIguaz:
* the cache will always be persisted (as before)
* the aggregates will always be persisted (if there are any new aggregates) (as before)
* Event's columns will be persisted only of a list of columns was explicitly provided

### DIFF
--- a/storey/writers.py
+++ b/storey/writers.py
@@ -3,7 +3,7 @@ import csv
 import io
 import json
 import random
-from typing import Optional, Union
+from typing import Optional, Union, List
 
 import aiofiles
 import pandas as pd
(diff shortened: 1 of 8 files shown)


storey/writers.py
```diff
@@ -347,3 +347,28 @@ async def _do(self, event):
...
+
+class WriteToTable(_ConcurrentByKeyJobExecution, _Writer):
+    """
+    Persists the data in `table` to its associated storage by key.
+
+    :param table: A table object.
+    :param columns: List of specific columns to write or '*' for all columns. Do not save any event's columns by default.
+    """
+
+    def __init__(self, table: Table, columns: Optional[list] = None,  infer_columns_from_data: bool = False, **kwargs):
```
gtopper: `columns` can also be a string... `Union[str, list, None]`

gtopper: What about `infer_columns_from_data` in docstring?
Extra space, needs reformatting.

gtopper: Actually, there's no reference in the code to '*'... Do you mean for it to be passed inside the list? If so, needs to be clearer.

gtopper (approved):

PR #63, opened 2020-11-19
Docstring and minor API changes.

gtopper:


storey/writers.py
```diff
@@ -228,16 +225,13 @@ class WriteToV3IOStream(Flow):
...
     :param stream_path: Path to the V3IO stream.
-    :type stream_path: string
     :param sharding_func: Function for determining the shard ID to which to write each event.
-    :type sharding_func: Function (Event=>int)
     :param batch_size: Batch size for each write request.
-    :type batch_size: int
     """
 
-    def __init__(self, storage, stream_path, sharding_func=None, batch_size=8, **kwargs):
+    def __init__(self, storage: V3ioDriver, stream_path: str, sharding_func: Optional[Callable[[Event], int]] = None, batch_size: int = 8,
```
talIguaz, line 233: Storage can be a "Driver" type not just V3ioDriver.. but i guess we can change it once we have more drivers 🤷‍♂️

talIguaz (approved):

storey/writers.py
gtopper: This class, WriteToV3IOStream, is pretty V3IO specific.

PR #65, opened 2020-11-22
Writer API: separate index from columns.

gtopper:
And other API changes.

### DIFF
--- a/integration/test_flow_integration.py
+++ b/integration/test_flow_integration.py
@@ -12,7 +12,7 @@ import v3io_frames as frames
 
 from storey import Filter, JoinWithV3IOTable, SendToHttp, Map, Reduce, Source, HttpRequest, build_flow, \
     WriteToV3IOStream, V3ioDriver, WriteToTSDB, Table, JoinWithTable, MapWithState, WriteToTable
-from .integration_test_utils import V3ioHeaders, append_return, test_base_time
+from .integration_test_utils import V3ioHeaders, append_return, setup_kv_teardown_test, setup_teardown_test, test_base_time
 
 
 class SetupKvTable(V3ioHeaders):
@@ -153,28 +153,25 @@ def test_write_to_tsdb():
     date_time_str = '18/09/19 01:55:1'
     for i in range(9):
         now = datetime.strptime(date_time_str + str(i) + ' UTC-0000', '%d/%m/%y %H:%M:%S UTC%z')
-        controller.emit([i + 1, i + 2, now, i])
-        expected.append([float(i + 1), float(i + 2), now, i])
+        controller.emit([now, i, i + 1, i + 2])
+        expected.append([now, f'{i}', float(i + 1), float(i + 2)])
 
     controller.terminate()
     controller.await_termination()
 
     client = frames.Client()
-    res = client.read("tsdb", tsdb_path, start='0', end='now')
+    res = client.read('tsdb', tsdb_path, start='0', end='now', multi_index=True)
     res = res.sort_values(['time'])
-    res['node'] = pd.to_numeric(res["node"])
     df = pd.DataFrame(expected, columns=['time', 'node', 'cpu', 'disk'])
-    df.set_index(keys=['time'], inplace=True)
+    df.set_index(keys=['time', 'node'], inplace=True)
     assert res.equals(df), f"result{res}\n!=\nexpected{df}"
 
 
 def test_write_to_tsdb_with_metadata_label():
     tsdb_path = f'tsdb_path-{int(time.time_ns() / 1000)}'
-    columns = ['cpu', 'disk', 'time', 'node']
-
     controller = build_flow([
         Source(),
-        WriteToTSDB(path=tsdb_path, time_col='time', index_cols='node', columns=['cpu', 'disk', '$time', 'node'], rate='1/h',
+        WriteToTSDB(path=tsdb_path, index_cols='node', columns=['cpu', 'disk'], rate='1/h',
                     max_events=2)
     ]).run()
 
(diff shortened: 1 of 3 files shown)


integration/test_flow_integration.py
```diff
@@ -2,19 +2,17 @@
...
-
-from _datetime import datetime, timedelta
-from datetime import timezone
-import pytz
+import v3io_frames as frames
 
 from storey import Filter, JoinWithV3IOTable, SendToHttp, Map, Reduce, Source, HttpRequest, build_flow, \
     WriteToV3IOStream, V3ioDriver, WriteToTSDB, Table, JoinWithTable, MapWithState, WriteToTable
-from .integration_test_utils import V3ioHeaders, append_return, setup_kv_teardown_test, setup_teardown_test, test_base_time
+from .integration_test_utils import V3ioHeaders, append_return, test_base_time
```
talIguaz: you need both `setup_kv_teardown_test, setup_teardown_test`

storey/writers.py
```diff
@@ -15,58 +15,81 @@
 
 
 class _Writer:
-    def __init__(self, columns: Optional[List[str]], infer_columns_from_data: bool):
-        self._first_event = None
+    def __init__(self, columns: Union[str, List[str], None], infer_columns_from_data: Optional[bool],
+                 index_cols: Union[list, str, None] = None):
```
talIguaz: List[str]

storey/writers.py
```diff
@@ -15,58 +15,81 @@
...
+            self._get_column_data_from_dict(data, event, self._columns, self._metadata_columns, self._rename_columns)
+            self._get_column_data_from_dict(data, event, self._index_cols, self._metadata_index_columns, self._rename_index_columns)
+        elif isinstance(data, list):
+            sub_metadata = bool(self._columns) and bool(self._metadata_columns)
+            sub_index_metadata = bool(self._index_cols) and bool(self._metadata_index_columns)
+            if sub_metadata or sub_index_metadata:
+                data = []
+                cursor = self._get_column_data_from_list(data, event, event.body, self._index_cols, self._metadata_index_columns)
+                self._get_column_data_from_list(data, event, event.body[cursor:], self._columns, self._metadata_columns)
         return data
```
talIguaz, line 97: if it's neither you just return the body and ignore the columns etc the user requested?
Better raise an error saying you can only use the `columns, infer_columns_from_data, index_cols` if your event is either dict or list

storey/writers.py
```diff
@@ -15,58 +15,81 @@
...
-                    metadata_attr = self._metadata_columns[column]
-                    new_value = getattr(event, metadata_attr)
-                    new_data.append(new_value)
-                else:
-                    new_data.append(data[data_cursor])
-                    data_cursor += 1
-            data = new_data
+                self._infer_columns_from_data = False
+            data = []
+            self._get_column_data_from_dict(data, event, self._columns, self._metadata_columns, self._rename_columns)
```
talIguaz, line 84: In case you are inferring you can just get `list(data.values())` rather than do all the `_get_column_data_from_dict` logic

storey/writers.py
```diff
@@ -15,58 +15,81 @@
...
+            self._get_column_data_from_dict(data, event, self._index_cols, self._metadata_index_columns, self._rename_index_columns)
+        elif isinstance(data, list):
+            sub_metadata = bool(self._columns) and bool(self._metadata_columns)
```
talIguaz, line 89: Probably better raise an error/log warning in case user specified `_infer_columns_from_data` with a list type body

storey/writers.py
```diff
@@ -148,20 +170,23 @@ class WriteToParquet(_Batching, _Writer):
     :type timeout_secs: int
     """
 
-    def __init__(self, path: str, index_cols: Union[list, str, None] = None, columns: Optional[List[str]] = None,
-                 partition_cols: Optional[List[str]] = None, infer_columns_from_data: bool = False, **kwargs):
+    def __init__(self, path: str, index_cols: Union[list, str, None] = None, columns: Union[str, List[str], None] = None,
```
talIguaz: List[str]

storey/writers.py
```diff
@@ -186,15 +211,19 @@ class WriteToTSDB(_Batching, _Writer):
     :param aggr_granularity: Granularity of server-side aggregations for this TSDB table (e.g. '1h').
     """
 
-    def __init__(self, path: str, time_col: str, columns: List[str], infer_columns_from_data: bool = False,
-                 index_cols: Union[str, List[str], None] = None, v3io_frames: Optional[str] = None, access_key: Optional[str] = None,
-                 container: str = "", rate: str = "", aggr: str = "", aggr_granularity: str = "", frames_client=None, **kwargs):
+    def __init__(self, path: str, time_col: str = '$time', columns: Union[str, List[str], None] = None,
+                 infer_columns_from_data: bool = False, index_cols: Union[str, List[str], None] = None, v3io_frames: Optional[str] = None,
```
talIguaz, line 219: isn't `index_cols` better be `label_cols` in this specific step?

gtopper: We wanted consistency across writers, no? 🤷‍♂️

storey/writers.py
```diff
@@ -84,12 +84,16 @@ def _event_to_writer_entry(self, event):
...
+            if self._infer_columns_from_data:
+                raise TypeError('Cannot infer_columns_from_data when event type is list. Inference is only possible from dict.')
             sub_metadata = bool(self._columns) and bool(self._metadata_columns)
             sub_index_metadata = bool(self._index_cols) and bool(self._metadata_index_columns)
             if sub_metadata or sub_index_metadata:
                 data = []
                 cursor = self._get_column_data_from_list(data, event, event.body, self._index_cols, self._metadata_index_columns)
                 self._get_column_data_from_list(data, event, event.body[cursor:], self._columns, self._metadata_columns)
+        else:
+            raise TypeError('Writer supports only events of type dict or list.')
```
talIguaz, line 96: better use the step's name rather than "Writer".
a flow can have multiple branches so the error should be clear enough

talIguaz: Actually, we better add the step name to every error we raise.. but that's for later

talIguaz (approved):

PR #67, opened 2020-11-22
Make test faster and more reliable.

gtopper:


tests/test_flow.py
```diff
@@ -518,15 +519,22 @@ def append_body_and_return(lst, x):
 
 
 def test_batch_with_timeout():
+    q = queue.Queue(1)
+
+    def reduce_fn(acc, x):
+        if x[0] == 0:
+            q.put(None)
+        return append_and_return(acc, x)
```
talIguaz: You can just append to acc here rather than call another function

talIguaz (approved):

PR #69, opened 2020-11-23
Query by key

dinal:


integration/test_aggregation_integration.py
```diff
@@ -80,9 +80,10 @@ def test_query_aggregate_by_key(setup_teardown_test, partitioned_by_key):
     other_table = Table(setup_teardown_test, V3ioDriver())
     controller = build_flow([
         Source(),
-        QueryAggregationByKey([FieldAggregator("number_of_stuff", "col1", ["sum", "avg", "min", "max"],
-                                               SlidingWindows(['1h', '2h', '24h'], '10m'))],
-                              other_table),
+        QueryByKey(["number_of_stuff_sum_1h", "number_of_stuff_sum_2h", "number_of_stuff_sum_24h",
```
talIguaz, line 171: why did you delete the avg aggregate?

storey/dtypes.py
```diff
@@ -288,8 +288,6 @@ def __init__(self, name: str, field: Union[str, Callable[[Event], object]], aggr
             self.value_extractor = field
         elif isinstance(field, str):
             self.value_extractor = lambda element: element[field]
-        else:
```
talIguaz, line 291: Why did you removed this check altogether?

storey/aggregations.py
```diff
@@ -172,27 +173,44 @@ async def _emit_worker(self):
...
     """
 
-    def __init__(self, aggregates: Union[List[FieldAggregator], List[Dict[str, object]]], table: Table,
-                 key: Union[str, Callable[[Event], object], None] = None,
-                 emit_policy: Union[EmitEveryEvent, FixedWindows, SlidingWindows, EmitAfterPeriod, EmitAfterWindow,
-                                    EmitAfterMaxEvent, Dict[str, object]] = _default_emit_policy,
-                 augmentation_fn: Optional[Callable[[Event, Dict[str, object]], Event]] = None, enrich_with: Optional[List[str]] = None,
-                 aliases: Optional[Dict[str, str]] = None, **kwargs):
-        AggregateByKey.__init__(self, aggregates, table, key, emit_policy, augmentation_fn, enrich_with, aliases, **kwargs)
+    def __init__(self, features, table, key=None, augmentation_fn=None, aliases=None):
```
talIguaz: Change the parameters to typed params + remove the `type` part in the docstring

storey/aggregations.py
```diff
@@ -172,27 +173,44 @@ async def _emit_worker(self):
...
+        self._schema = table._storage._load_schema(table._container, table._table_path)
```
talIguaz: aren't  you suppose to `await` here? how is this working

storey/aggregations.py
```diff
@@ -172,27 +173,44 @@ async def _emit_worker(self):
...
+        self._aggrs = []
+        self._enrich_cols = []
+        resolved_aggrs = {}
+        for feature in features:
+            if re.match(r".*_[a-z]+_[0-9]+[a-z]", feature):
```
talIguaz: isn't it supposed to be `[a-z]+_[a-z]+_[0-9]+[smhd]`

storey/aggregations.py
```diff
@@ -172,27 +173,44 @@ async def _emit_worker(self):
...
+                parts = feature.rsplit('_', 1)
```
talIguaz: might be more readable to do `aggregate, window = feature.rsplit('_', 1)`

storey/aggregations.py
```diff
@@ -172,27 +173,44 @@ async def _emit_worker(self):
...
+                if parts[0] in resolved_aggrs:
+                    resolved_aggrs[parts[0]].append(parts[1])
+                else:
+                    resolved_aggrs[parts[0]] = [parts[1]]
+            else:
+                self._enrich_cols.append(feature)
+        for name, windows in resolved_aggrs.items():
+            parts = name.rsplit('_', 1)
```
talIguaz: same here, better initialize with `aggregate_name, aggregate = name.rsplit('_', 1)` or something of the sort

storey/aggregations.py
talIguaz: looks like nothing uses it

storey/aggregations.py
dinal: i agree with the end [smhd]
but the feature name is not limited to a-z, it can have _ or any other char in it

storey/dtypes.py
dinal: when i'm creating FieldAggregators (line 200) i create them with None column field cause i don't have one and it doesn't matter in query step
i can change it to a random string or whatever

talIguaz: I would allow None, but still check that no other weird type is passed

storey/aggregations.py
talIguaz: Sure, but now you also accept `_sum_4h`. which does not sound right

storey/aggregations.py
```diff
@@ -187,30 +187,30 @@ class QueryByKey(AggregateByKey):
     :param augmentation_fn: Function that augments the features into the event's body. Defaults to updating a dict. (Optional)
     :type augmentation_fn: Function ((Event, dict) => Event)
```
talIguaz: Remove all `type` comment lines

storey/aggregations.py
```diff
@@ -172,27 +173,40 @@ async def _emit_worker(self):
...
+                    resolved_aggrs[name] = [window]
+            else:
+                self._enrich_cols.append(feature)
+        for name, windows in resolved_aggrs.items():
+            feature, aggr = name.rsplit('_', 1)
+            # setting as SlidingWindow temporarily until actual window type will be read from schema
+            self._aggrs.append(FieldAggregator(name=feature, field=None, aggr=[aggr], windows=SlidingWindows(windows, '10m')))
+
+        AggregateByKey.__init__(self, self._aggrs, table, key, augmentation_fn=augmentation_fn,
+                                enrich_with=self._enrich_cols, aliases=aliases, use_windows_from_schema=True, **kwargs)
```
talIguaz, line 209: looks like `_use_windows_from_schema` isn't used in `AggregateByKey`

storey/aggregations.py
```diff
@@ -317,6 +323,19 @@ async def get_or_save_schema(self):
 
         should_update = True
         if self._schema:
+            if self._use_windows_from_schema:
+                for aggr in self._aggregates:
+                    schema_aggr = self._schema[aggr.name]
+                    window_type = schema_aggr['window_type']
+                    window_secs = str(int(aggr.windows.max_window_millis / 1000)) + 's'
```
talIguaz, line 330: I would actually enable `SlidingWindows` and `FixedWindows` to accept millis as period and windows in addition to strings. that way you won't need to do all these conversion just to get it back to millis in the end

talIguaz (approved):

PR #78, opened 2020-11-26
Fix CSV performance.

gtopper:


storey/writers.py
```diff
@@ -116,51 +116,51 @@ class WriteToCSV(Flow, _Writer):
 
     def __init__(self, path: str, columns: Optional[List[str]] = None, header: bool = False, infer_columns_from_data: bool = False,
                  force_flush_after: int = 1, **kwargs):
-        Flow.__init__(self, **kwargs)
+        _Batching.__init__(self, max_events=128, timeout_secs=3, **kwargs)
```
talIguaz: might consider exposing this to the user. When you write CSV you do want to "control" the latency in which the file is written, sometime 3s might be too much

talIguaz (approved):

PR #79, opened 2020-11-30
Makedirs before write.

gtopper:


talIguaz:
Add test for csv as well

talIguaz (approved):

PR #80, opened 2020-11-30
Reify metadata in WriteToV3IOStream.

gtopper:


talIguaz (commented):
Better change the implementation so that WriteToV3ioStream inherits `_Writer`

integration/test_flow_integration.py
```diff
@@ -123,6 +123,38 @@ def test_write_to_v3io_stream():
     assert shard1_data == [b'1', b'3', b'5', b'7', b'9']
 
 
+def test_write_dict_to_v3io_stream():
+    stream_path = f'bigdata/test_write_dict_to_v3io_stream/{int(time.time_ns() / 1000)}/'
+    asyncio.run(SetupStream().setup(stream_path))
```
talIguaz: How about cleaning after the test?
you can add stream_setup_teardown_fixture for that

integration/test_flow_integration.py
```diff
@@ -123,6 +123,38 @@ def test_write_to_v3io_stream():
...
+    controller = build_flow([
+        Source(),
+        WriteToV3IOStream(V3ioDriver(), stream_path, sharding_func=lambda event: int(event.key), columns=['$key'],
+                          infer_columns_from_data=True)
+    ]).run()
+    expected_shard0 = []
+    expected_shard1 = []
+    for i in range(10):
+        controller.emit({'mydata': 'abcdefg'}, key=f'{i}')
+        expected = {'mydata': 'abcdefg', 'key': f'{i}'}
```
talIguaz, line 144: if you specified only the `$key` in `columns`, why do you expect `mydata` as well?

gtopper: Because `infer_columns_from_data=True`

talIguaz:
* If you introduced `retain_dict` in this PR please make `WriteToTable` use that as well 😗 
* please add a couple more tests to stream with different variations of `columns` and `infer_columns_from_data `

talIguaz (approved):
✅

PR #83, opened 2020-12-01
split windows to hidden and explicit

dinal:


storey/aggregations.py
```diff
@@ -415,67 +418,115 @@ def __getitem__(self, key):
...
         self.max_value = max_value
         self.buckets = []
-        self.is_hidden = is_hidden
         self.should_persist = True
         self.pending_aggr = {}
         self.storage_specific_cache = {}
+        self.base_time = base_time
+        self.explicit_windows = None
+        self.hidden_windows = None
+        self._initial_data = initial_data
```
talIguaz, line 433: no need to store `initial_data` here, you can pass it to the `init` directly

storey/aggregations.py
```diff
@@ -240,17 +240,20 @@ def __init__(self, key, aggregates, base_time, initial_data=None):
...
-                if column_name in self.aggregation_buckets:
-                    self.aggregation_buckets[column_name].is_hidden &= is_hidden
-                else:
-                    initial_column_data = None
-                    if initial_data and column_name in initial_data:
-                        initial_column_data = initial_data[column_name]
-                    self.aggregation_buckets[column_name] = \
-                        AggregationBuckets(aggregation_metadata.name, aggr, aggregation_metadata.windows, base_time,
-                                           aggregation_metadata.max_value, is_hidden, initial_column_data)
+            for meta in aggregation_metadata.aggregations:
```
talIguaz, line 243: why do you need this `for`?

storey/aggregations.py
```diff
@@ -415,67 +418,115 @@ def __getitem__(self, key):
...
+        self._initial_data = initial_data
+        self.last_bucket_start_time = None
+        self.first_bucket_start_time = None
+
+        if is_hidden:
+            self.hidden_windows = initial_windows
+        else:
+            self.explicit_windows = initial_windows
+
+        self.period_millis = initial_windows.period_millis
```
talIguaz, line 442: Why do you need a copy of - `self.period_millis, self.max_window_millis, self.total_number_of_buckets`
you can just have like a Getter method to `get_window_periods()`, and `get_max_window` that will take the max of both hidden and explicit, etc.. and not worry about updating these fields since they will be taken on the fly from the `self.hidden_windows & self.explicit_windows`

storey/aggregations.py
```diff
@@ -240,17 +240,20 @@ def __init__(self, key, aggregates, base_time, initial_data=None):
...
-                    if initial_data and column_name in initial_data:
-                        initial_column_data = initial_data[column_name]
-                    self.aggregation_buckets[column_name] = \
-                        AggregationBuckets(aggregation_metadata.name, aggr, aggregation_metadata.windows, base_time,
-                                           aggregation_metadata.max_value, is_hidden, initial_column_data)
+            for meta in aggregation_metadata.aggregations:
+                for aggr, is_hidden in get_all_raw_aggregates_with_hidden([meta]).items():
+                    column_name = f'{aggregation_metadata.name}_{aggr}'
+                    if column_name in self.aggregation_buckets:
+                        self.aggregation_buckets[column_name].add_to_window(is_hidden, copy.deepcopy(aggregation_metadata.windows))
```
talIguaz, line 247: i think it's better to go over the aggregations once and "aggregate" all the windows and only then create the aggregation buckets with the already finite windows. the `add_to_window` logic shouldn't really be a part of the AggregationBuckets, it's not updating during the flow just once when it's initialized

storey/aggregations.py
```diff
@@ -415,67 +418,115 @@ def __getitem__(self, key):
...
         self.name = name
         self.aggregation = aggregation
-        self.window = window
         self.max_value = max_value
         self.buckets = []
-        self.is_hidden = is_hidden
         self.should_persist = True
         self.pending_aggr = {}
         self.storage_specific_cache = {}
+        self.base_time = base_time
```
talIguaz, line 430: make all these private, `_XXX`

storey/aggregations.py
```diff
@@ -611,14 +671,15 @@ def aggregate(self, timestamp, value):
...
+        args_results = [bucket.get_features(timestamp, get_hidden=True) for bucket in self.args]
+        for j in range(len(self.window.windows)):
+            window_string = self.window.windows[j][1]
             current_args = []
-            for window_result in args_results:
-                current_args.append(window_result[i])
-
+            for i in range(len(args_results)):
+                for key, value in args_results[i].items():
+                    if key.endswith('_' + window_string):
```
talIguaz: by now you probably know exactly the "feature's" name so you can access it by key rather than for and if, no?

PR #99, opened 2020-12-13
Fix nontermination on error when awaiting result.

gtopper:


storey/flow.py
```diff
@@ -391,6 +391,7 @@ async def _worker(self):
         try:
             while True:
                 job = await self._q.get()
+                print(f'got job {job}')
```
talIguaz: 😮

storey/flow.py
```diff
@@ -430,7 +443,7 @@ async def _do(self, event):
             await self._worker_awaitable
             return await self._do_downstream(_termination_obj)
         else:
-            task = self._process_event(event)
+            task = self._safe_process_event(event)
```
talIguaz, line 445: should also add this to `_ConcurrentByKeyJobExecution`

gtopper:
Merging as requested by @dinal.

PR #103, opened 2020-12-15
Deep copy event body when branching.

gtopper:


storey/flow.py
```diff
@@ -47,15 +47,16 @@ async def _do_downstream(self, event):
         # If there is more than one outlet, allow concurrent execution.
         tasks = []
         for i in range(1, len(self._outlets)):
-            tasks.append(asyncio.get_running_loop().create_task(self._outlets[i]._do(event)))
+            event_copy = copy.copy(event)
+            event_copy.body = copy.deepcopy(event.body)
```
talIguaz: why don't you just do deep copy to the whole event? the other attributes are just like 8 builtin types nothing that's worth the separation no?

gtopper: Because you can't deep copy an AwaitableResult (because of the lock). I could maybe remove it, deep copy, then put it back in.

talIguaz (approved):

PR #104, opened 2020-12-16
split windows to explicit and hidden

dinal:


storey/dtypes.py
```diff
@@ -106,6 +106,18 @@ def __init__(self, period, windows):
         self.windows = windows  # list of tuples of the form (3600000, '1h')
         self.total_number_of_buckets = int(self.max_window_millis / self.period_millis)
 
+    def add_to_window(self, new):
```
talIguaz: rename to `merge`

storey/aggregations.py
```diff
@@ -685,6 +743,30 @@ def get_and_flush_pending(self):
...
+        number_of_buckets = 0
+        if self.explicit_windows:
+            number_of_buckets = max(number_of_buckets, self.explicit_windows.total_number_of_buckets)
+        if self.hidden_windows:
+            number_of_buckets = max(number_of_buckets, self.hidden_windows.total_number_of_buckets)
+        return number_of_buckets
+
+    def get_period_millis(self):
+        period_millis = 0
+        if self.explicit_windows:
```
talIguaz: Why do you need the max? there period should be the same just get the period from the first window you have (either explicit or hidden)

integration/test_aggregation_integration.py
```diff
@@ -195,6 +195,75 @@ def test_query_aggregate_by_key(setup_teardown_test, partitioned_by_key):
         f'actual did not match expected. \n actual: {actual} \n expected: {expected_results}'
 
 
+@pytest.mark.parametrize('query_aggregations', [["number_of_stuff_sum_1h", "number_of_stuff_avg_2h"],
```
talIguaz, line 257: Why does the order matter?
Also looks like you need a lot of unit tests to cover different scenarios.

storey/aggregations.py
```diff
@@ -239,18 +239,33 @@ def __init__(self, key, aggregates, base_time, initial_data=None):
...
-                if column_name in self.aggregation_buckets:
-                    self.aggregation_buckets[column_name].is_hidden &= is_hidden
-                else:
-                    initial_column_data = None
-                    if initial_data and column_name in initial_data:
-                        initial_column_data = initial_data[column_name]
-                    self.aggregation_buckets[column_name] = \
-                        AggregationBuckets(aggregation_metadata.name, aggr, aggregation_metadata.windows, base_time,
-                                           aggregation_metadata.max_value, is_hidden, initial_column_data)
+            for meta in aggregation_metadata.aggregations:
```
talIguaz, line 244: Why can't iterate over the `get_all_raw_aggregates_with_hidden` directly

storey/aggregations.py
```diff
@@ -239,18 +239,33 @@ def __init__(self, key, aggregates, base_time, initial_data=None):
...
+                for aggr, is_hidden in get_all_raw_aggregates_with_hidden([meta]).items():
+                    if (aggregation_metadata.name, aggr, aggregation_metadata.max_value) in windows:
+                        aggr_windows = windows[(aggregation_metadata.name, aggr, aggregation_metadata.max_value)]
+                        if is_hidden in aggr_windows:
+                            aggr_windows[is_hidden].add_to_window(is_hidden, copy.deepcopy(aggregation_metadata.windows))
```
talIguaz: isn't `add_to_window` only accept the new windows?

talIguaz: Also, aren't `aggregation_metadata.windows` internal + we only iterate every window once? why do we need to deep copy it

storey/aggregations.py
```diff
@@ -239,18 +239,33 @@ def __init__(self, key, aggregates, base_time, initial_data=None):
...
+                        else:
+                            aggr_windows[is_hidden] = copy.deepcopy(aggregation_metadata.windows)
+                    else:
+                        windows[(aggregation_metadata.name, aggr, aggregation_metadata.max_value)] =\
```
talIguaz, line 253: What if we have `avg_1h` and `sum_1h` won't we have `sum_1h` in both explicit and hidden? is that what we want?

storey/aggregations.py
```diff
@@ -571,9 +609,15 @@ def get_features(self, timestamp):
...
 
         # In case our pre aggregates already have the answer
-        result = {}
-        for i, win in enumerate(self.window.windows):
-            result[f'{self.name}_{self.aggregation}_{win[1]}'] = self._current_aggregate_values[i].get_value()[1]
+        if get_hidden:
+            loop_over = self.hidden_windows
+        else:
+            loop_over = self.explicit_windows
+        if not loop_over:
```
talIguaz: Is this case even possible?

storey/aggregations.py
```diff
@@ -561,7 +598,8 @@ def get_aggregation_for_aggregation(self):
             return 'sum'
         return self.aggregation
 
-    def get_features(self, timestamp):
+    def get_features(self, timestamp, get_hidden=False):
```
talIguaz: we probably need to also request specific windows, because if a user queries for `sum_3h`, `avg_1h`, `stdvar_2h`:
when avg requests sum for virtual aggregates it will get 1h&2h which probably blow things up

storey/aggregations.py
```diff
@@ -584,21 +628,30 @@ def calculate_features(self, timestamp):
         if current_time_bucket_index < 0:
             return result
 
+        all_windows = []
+        if self.explicit_windows:
+            some_window = self.explicit_windows
```
talIguaz: Maybe rename to something more meaningful other than "some_windows"

storey/aggregations.py
```diff
@@ -584,21 +628,30 @@ def calculate_features(self, timestamp):
         if current_time_bucket_index < 0:
             return result
 
+        all_windows = []
```
talIguaz: Better pass what windows you want to calculate the features for, because sometimes you only need the explicit and sometimes only the hidden and sometimes as mentioned in another comment you want specific hidden windows.

storey/aggregations.py
```diff
@@ -638,17 +693,20 @@ def initialize_from_data(self, data, base_time):
             timestamp1, timestamp2 = aggregation_bucket_initial_data.keys()
             first_time, last_time = min(timestamp1, timestamp2), max(timestamp1, timestamp2)
 
-        bucket_index = self.window.total_number_of_buckets - 1
-        self.last_bucket_start_time = self.window.get_window_start_time_by_time(base_time)
+        bucket_index = total_buckets - 1
+        if self.explicit_windows:
```
talIguaz: You are doing a lot of these conditions maybe just check the condition once on init and save the desired `get_window_start_time_by_time` function in `self` and just call it every time

storey/aggregations.py
```diff
@@ -685,6 +743,30 @@ def get_and_flush_pending(self):
         self.pending_aggr = {}
         return pending
 
+    def get_max_window_millis(self):
```
talIguaz, line 738: Now im thinking maybe you should save the `max_windows`, `total_buckets` and `period` as a member of the class and calculate it once since it is used in a bunch of places. can impact performance.

integration/test_aggregation_integration.py
dinal: at some point it worked with 1 order but failed with the other, i wanted to cover it

storey/aggregations.py
dinal: i had to normalize the 2 cases when coming from AggragteByKey and QueryByKey
lets say the aggregations are (sum, avg) - 
AggragteByKey([sum, avg]) will create 1 FieldAggregator with the both of them so when calling  get_all_raw_aggregates_with_hidden([sum, avg]) you will get {'sum': False, 'count': True}, the case of "hidden" sum is lost (hidden for the avg aggrgation)
now if you go from QueryByKey([sum, avg]) you will get 2 FieldAggregator that will call get_all_raw_aggregates_with_hidden twice and yield  avg -> {'sum': True, 'count': True}, sum -> {'sum': False}

storey/aggregations.py
dinal: but it is just a random window i need for the one call of round_up_time_to_window

storey/aggregations.py
dinal: that's what i did in the previous pr :|

storey/aggregations.py
```diff
@@ -604,9 +604,8 @@ def get_features(self, timestamp, windows=None):
         # Either a) we were signaled b) the requested timestamp is prior to our pre aggregates
         if self._need_to_recalculate_pre_aggregates or \
                 self.get_bucket_index_by_timestamp(timestamp) < self.get_bucket_index_by_timestamp(self._last_data_point_timestamp) or \
-                self.max_value:
-            self._need_to_recalculate_pre_aggregates = False
```
talIguaz, line 570: we still need to set `_need_to_recalculate_pre_aggregates` to False

storey/aggregations.py
```diff
@@ -620,22 +619,26 @@ def get_features(self, timestamp, windows=None):
...
             some_window = self.hidden_windows
-            for win in self.hidden_windows.windows:
-                if win not in all_windows:
-                    all_windows.append(win)
+        else:
+            all_windows = []
+            if self.explicit_windows:
+                some_window = self.explicit_windows
+                all_windows.extend(self.explicit_windows.windows)
+            if self.hidden_windows:
```
talIguaz: Is it still possible to get here?
aren't you either don't have `windows` which means you use explicit windows or you have window (which are some of the hidden) and you use them?

dinal: when coming from AggregationBuckets:__init__

talIguaz (approved):
LGTM ✅

PR #106, opened 2020-12-16
Basic Recover step.

gtopper:


storey/flow.py
```diff
@@ -123,6 +123,26 @@ async def _do(self, event):
...
+        Flow.__init__(self, **kwargs)
+
+        self._exception_to_downstream = exception_to_downstream
+
+    async def _do(self, event):
+        if not self._outlets or event is _termination_obj:
+            return await super()._do_downstream(event)
+        else:
+            try:
+                await super()._do_downstream(event)
```
talIguaz, line 162: aren't you supposed to return here as well?

gtopper: Hm, not really? `_do` only returns a termination result.

talIguaz (approved):

PR #108, opened 2020-12-16
Group by key

Michaelliv:


storey/steps/groupbykey.py
```diff
@@ -0,0 +1,32 @@
...
+    Groups events by given key (or key extractor function) into lists of up to max_events events. Each emitted list
+    contained max_events events, unless timeout_secs seconds have passed since the first event in the group was
+    received, at which the group is emitted with potentially fewer than max_events
+    event.
+    :param key: The key on which events are grouped by.
+    :type key: str or callable
+    :param max_events: Maximum number of events per emitted batch. Set to None to emit all events in one batch on flow termination.
+    :type max_events: int or None
+    :param timeout_secs: Maximum number of seconds to wait before a batch is emitted.
+    :type timeout_secs: int
```
gtopper: No need for :type when using type annotations.

storey/flow.py
```diff
@@ -127,7 +147,7 @@ class _UnaryFunctionFlow(Flow):
     def __init__(self, fn, **kwargs):
         super().__init__(**kwargs)
         if not callable(fn):
-            raise TypeError(f'Expected a callable, got {type(fn)}')
+            raise TypeError(f"Expected a callable, got {type(fn)}")
```
gtopper: Please don't...

storey/steps/groupbykey.py
Michaelliv: Thats new to me, I'll remove it

storey/flow.py
```diff
@@ -608,63 +611,98 @@ async def _handle_completed(self, event, response):
...
+    def _init_timeout_task(self):
+        for key in self._batch.keys():
+            self._timeout_task = asyncio.get_running_loop().create_task(
+                self._sleep_and_emit(key)
+            )
+            break
+
+    async def _emit_batch(self, batch_key: Optional[str] = None):
+        batch_to_emit = self._batch.pop(batch_key)
+        if len(batch_to_emit) > 0:
```
talIguaz: i think you will only have an entry in `self._batch` if there are items in `batch_to_emit` so it's a redundant check

storey/flow.py
```diff
@@ -608,63 +611,98 @@ async def _handle_completed(self, event, response):
...
+        self._event_count[key] = self._event_count[key] + 1
+        self._batch[key].append(self._event_to_batch_entry(event))
+
+        if self._event_count[key] == self._max_events:
+            if self._timeout_task is not None and not self._timeout_task.cancelled():
+                self._timeout_task.cancel()
+            await self._emit_batch(key)
+
+    def _init_timeout_task(self):
+        for key in self._batch.keys():
```
talIguaz: Aren't you creating a task per key here? we decided to only have one task for the "next" key to emit

storey/flow.py
```diff
@@ -608,63 +611,98 @@ async def _handle_completed(self, event, response):
...
-        await asyncio.sleep(self._timeout_secs)
-        await self._emit_batch()
+    async def _sleep_and_emit(self, key: Optional[str] = None):
+        time_delta = (
+            datetime.now(timezone.utc) - self._batch_time[key]
+        ).total_seconds()
+        if time_delta < self._timeout_secs:
+            await asyncio.sleep(self._timeout_secs - time_delta)
+        await self._emit_batch(key)
+        self._timeout_task = None
```
talIguaz: rather than setting "yourself" to None you can just sleep till the next key to emit as long as there are keys in the dict

storey/steps/groupbykey.py
```diff
@@ -0,0 +1,29 @@
+from typing import Optional, Union
+
+from storey.dtypes import Event
+from storey.flow import _Batching
+
+
+class GroupByKey(_Batching):
```
talIguaz: In my opinion it's better to just extend this functionality to `Batch`, rather than a new Step. also it's not really `GroupByKey` more of `BatchByKey`

storey/flow.py
Michaelliv: There's a break right after it. could have wrapped `self._batch.keys()` with a list, then take the first element, but that wouldn't be less messy.

storey/flow.py
Michaelliv: Yeah just following previous logic. Will change that.

storey/steps/groupbykey.py
Michaelliv: Then maybe `Batch` should change to `Group` with an optional key, maybe it should follow a more standard name for this type of operation. I've never called a batch function, but i did call GroupBy.

What do you think?

tests/test_group_by_key.py
```diff
@@ -0,0 +1,106 @@
...
+            GroupByKey("value", 4, 100),
+            Reduce([], lambda acc, x: append_and_return(acc, x)),
+        ]
+    ).run()
+
+    values_1 = [i for i in range(3)]
+    values_2 = [i for i in range(3)]
+    values_3 = [i for i in range(3)]
+    for i in range(3):
+        controller.emit({"value": values_1[i], "data": values_1[i]})
```
talIguaz: You are sending the same key sequentially.
better have tests that emit the keys not in order (like [key1, key1, key2, key3, key2, key1]) and test scenarios where you reach max events for a certain key and scenarios where you reach timeout for a specific key, etc..

Michaelliv: No problem

storey/flow.py
talIguaz: oh right, why don't just do `key = next(iter(self._batch))`

storey/steps/groupbykey.py
talIguaz: GroupBy is sort of a finite/reduce operation it's grouping the entire dataset, while Batch/BatchByKey just batches a bunch of events based on some logic(max events/timeout) and emits them forward possible by key.

storey/flow.py
Michaelliv: Nice, will do

storey/flow.py
Michaelliv: Please elaborate

storey/flow.py
```diff
@@ -666,24 +655,52 @@ async def _do(self, event):
...
+        while self._batch:
+            key = next(iter(self._batch.keys()))
+            time_delta = datetime.now(timezone.utc) - self._batch_time[key]
+            delta_seconds = time_delta.total_seconds()
+            if delta_seconds < self._timeout_secs:
+                self._timeout_task_key = key
+                await asyncio.sleep(self._timeout_secs - delta_seconds)
             await self._emit_batch(key)
 
+        self._timeout_task.cancel()
```
talIguaz: This **is** the `_timeout_task` so you don't need to cancel it i guess just reset the timeout_key and finish the function.

tests/test_flow.py
```diff
@@ -2,12 +2,15 @@
...
 from datetime import datetime
+from random import choice
 
 import pandas as pd
 from aiohttp import InvalidURL
 
-from storey import build_flow, Source, Map, Filter, FlatMap, Reduce, FlowError, MapWithState, ReadCSV, Complete, AsyncSource, Choice, \
-    Event, Batch, Table, WriteToCSV, DataframeSource, MapClass, JoinWithTable, ReduceToDataFrame, ToDataFrame, WriteToParquet, \
+from storey import build_flow, Source, Map, Filter, FlatMap, Reduce, FlowError, MapWithState, ReadCSV, Complete, \
+    AsyncSource, Choice, \
```
talIguaz: Weird line breaking

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
...
+    async def _emit_all(self):
+        batch_keys = list(self._batch.keys())
+        for batch_key in batch_keys:
+            await self._emit_batch(batch_key)
 
 
 class Batch(_Batching):
     """
-    Batches events into lists of up to max_events events. Each emitted list contained max_events events, unless timeout_secs seconds
+    Batches events into lists of up to max_events events. Each emitted list  contained max_events events, unless timeout_secs seconds
```
talIguaz: extra space

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
...
     have passed since the first event in the batch was received, at which the batch is emitted with potentially fewer than max_events
     event.
     :param max_events: Maximum number of events per emitted batch. Set to None to emit all events in one batch on flow termination.
     :type max_events: int or None
     :param timeout_secs: Maximum number of seconds to wait before a batch is emitted.
     :type timeout_secs: int
+    :param group_by_key: If true, Batch will will group events by given key
```
talIguaz: No need for `type` in docstring

talIguaz: double `will`

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
...
         self._timeout_secs = timeout_secs
+        self._group_by_key = group_by_key
+
         if self._timeout_secs is not None and self._timeout_secs <= 0:
-            raise ValueError('Batch timeout cannot be 0 or negative')
+            raise ValueError("Batch timeout cannot be 0 or negative")
+
+        self._event_count: Dict[Optional[str], int] = defaultdict(int)
+        self._batch: Dict[Optional[str], List[Any]] = defaultdict(list)
+        self._batch_time: Dict[Optional[str], datetime] = {}
```
gtopper, line 633: Does it make sense for the key to be `None`?

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
...
+    async def _emit_batch(self, batch_key: Optional[str] = None):
+        batch_to_emit = self._batch.pop(batch_key)
+        batch_time = self._batch_time.pop(batch_key)
+        del self._event_count[batch_key]
+        await self._emit(batch_to_emit, batch_time)
 
-            await self._emit(batch_to_emit, batch_time)
+    async def _emit_all(self):
+        batch_keys = list(self._batch.keys())
+        for batch_key in batch_keys:
```
gtopper: Just
```
for batch_key in self._batch:
```
does the trick without creating an intermediate list.

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
...
-        self._batch_time = None
-        self._timeout_task = None
-
+        self._max_events: int = max_events
         self._timeout_secs = timeout_secs
+        self._group_by_key = group_by_key
+
         if self._timeout_secs is not None and self._timeout_secs <= 0:
-            raise ValueError('Batch timeout cannot be 0 or negative')
+            raise ValueError("Batch timeout cannot be 0 or negative")
```
gtopper: Please leave as single quotes.

storey/flow.py
```diff
@@ -608,73 +611,118 @@ async def _handle_completed(self, event, response):
 
 
 class _Batching(Flow):
-    def __init__(self, max_events: Optional[int] = None, timeout_secs=None, **kwargs):
+    def __init__(
+            self,
+            max_events: Optional[int] = None,
+            timeout_secs: Optional[int] = None,
+            group_by_key: bool = False,
+            key: Optional[Union[str, callable]] = None,
```
gtopper: `key_field` or `key_extractor`, really. Do we need `group_by_key` at all? Can't it just be implied when key field is provided?

storey/flow.py
talIguaz: In case it's not grouped by key

storey/flow.py
talIguaz: No, because if you want to group by `event.key` you will pass `group_by_key=True` and no `key`

gtopper: Ah. We dealt with this elsewhere using '$key'. What about `group_by_key=False` and `key_field` being defined?

storey/flow.py
Michaelliv: > No need for `type` in docstring

Habits

storey/flow.py
Michaelliv: > Ah. We dealt with this elsewhere using '$key'. What about **`group_by_key=False` and `key_field`** being defined?

`_Batching` will run as if key_field is not defined.

But i do like the `$key` approach. The only downfall for this is that you have to read the documentation to understand how to use it, whereas switching to `group_by_key=True` will automatically group by `Event.key`. So i guess it's a taste thing. Let me know what is closer to the semantics of the rest of the code base and i'll change it accordingly.

storey/flow.py
Michaelliv: Can't do that since `self._emit_batch(batch_key)` changes `self._batch`'s size

storey/flow.py
```diff
@@ -615,31 +615,38 @@ def __init__(
...
         super().__init__(**kwargs)
 
         self._max_events: int = max_events
         self._timeout_secs = timeout_secs
-        self._group_by_key = group_by_key
 
         if self._timeout_secs is not None and self._timeout_secs <= 0:
             raise ValueError('Batch timeout cannot be 0 or negative')
 
+        self._extract_key: Optional[Callable[[Event], str]] = None
```
talIguaz: Semantically better to set `self._extract_key = create_extact_key_func()` or something and have that method return lambda rather than initializing it here to None and then init it right after in another place

storey/flow.py
```diff
@@ -716,8 +714,7 @@ class Batch(_Batching):
     event.
     :param max_events: Maximum number of events per emitted batch. Set to None to emit all events in one batch on flow termination.
     :param timeout_secs: Maximum number of seconds to wait before a batch is emitted.
-    :param group_by_key: If true, Batch will group events by given key
-    :param key: The key by which events are grouped, when None will use Event.key for grouping.
+    :param key: The key by which events are grouped, use '$key' to group events by the Event.key property.
```
talIguaz: Specify default behavior

tests/test_flow.py
```diff
@@ -644,6 +645,96 @@ def append_body_and_return(lst, x):
...
+        ]
+    ).run()
+
+    values_1 = [i for i in range(4)]
+    values_2 = [i for i in range(4)]
+    values_3 = [i for i in range(4)]
+    values_4 = [i for i in range(4)]
+
+    for _ in range(4):
+        rand_val_1 = choice(values_1)
```
talIguaz, line 667: If everything is random you can't really validate specific flows, maybe better to have test specific flows with a specific predefined input

talIguaz: Also better add tests for other scenarios like:
* timeout of a some keys.
* reaching max events of key2 while key1 timeout is "sleeping"
* and a bunch of other variations

storey/flow.py
```diff
@@ -608,73 +611,115 @@ async def _handle_completed(self, event, response):
...
+    potentially fewer than max_events event.
+    :param max_events: Maximum number of events per emitted batch. Set to None to emit all events in one batch on flow
+    termination.
     :param timeout_secs: Maximum number of seconds to wait before a batch is emitted.
-    :type timeout_secs: int
+    :param key: The key by which events are grouped. By default (None), events are not grouped.
+    Other options may be:
+    Set a '$key' to group events by the Event.key property.
+    set a 'str' key to group events by Event.body[str].
+    set a Callable[Union[Event, Any], Any] to group events by a a custom key extractor.
```
gtopper: Isn't `Union[Event, Any]` just `Any`? :)

storey/flow.py
```diff
@@ -608,73 +611,115 @@ async def _handle_completed(self, event, response):
...
+        self._timeout_task_key: Optional[str] = None
+
+    @staticmethod
+    def _create_key_extractor(key) -> Callable:
+        if key is None:
+            return lambda event: None
+        elif callable(key):
+            return key
+        elif isinstance(key, str):
+            if key == "$key":
```
gtopper: '$key'

talIguaz (approved):

storey/flow.py
Michaelliv: Indeed something or any == Any, but in case of `full_event=True` its specifically `Event`, you think Any is better?

PR #110, opened 2020-12-20
Common steps

Michaelliv:


tests/test_steps.py
```diff
@@ -0,0 +1,123 @@
...
+from storey import build_flow, Source, Map
+from storey.dtypes import FlowError, Event
+from storey.steps import Flatten, Sample, Assert, ForEach
+from time import sleep
+
+
+def test_assert():
+    try:
+        controller = build_flow(
+            [Source(), Assert(lambda event: event > 10, raise_on_exception=True)]
```
gtopper: Please stick to the convention regarding line breaks.

storey/steps/flatten.py
```diff
@@ -0,0 +1,18 @@
...
+    Flattens sequences of sequences (i.e lists of lists, sets of sets, etc...) into a single sequence, by default
+    casts sequence to list. When setting to_set=True, casts sequence to set.
+    """
+
+    def __init__(self, to_set: bool = False, **kwargs):
+        super().__init__(**kwargs)
+        self.to_set = to_set
+
+    def do(self, event):
+        type_cast = set if self.to_set else list
```
gtopper: This should go in `__init__`.

storey/steps/foreach.py
```diff
@@ -0,0 +1,9 @@
+from storey.flow import _UnaryFunctionFlow
+
+
+class ForEach(_UnaryFunctionFlow):
+    """Applies given function on each event in the stream, passes original event downstream"""
```
gtopper: Cool, just need to end sentences with a period :)

tests/test_steps.py
```diff
@@ -0,0 +1,123 @@
...
+        def __init__(self, max: int):
+            self.max = max
+            self.count = 0
+
+        def __call__(self, event):
+            self.count += 1
+            return self.count <= self.max
+
+    controller = build_flow(
+        [Source(), Assert(CountUntil(5)), Sample(5), Assert(CountUntil(1))]
```
gtopper: Using an Assert step is not good enough. If, for example, Sample had a bug where it emitted nothing, this test would still pass. Better use `Reduce` and `assert` as in other tests.

tests/test_steps.py
```diff
@@ -0,0 +1,123 @@
...
+
+    controller = build_flow(
+        [Source(), Assert(CountUntil(6)), Sample(rate_count=2), Assert(CountUntil(2))]
+    ).run()
+
+    controller.emit(1)
+    controller.emit(1)
+    controller.emit(1)
+
+    sleep(2)
```
gtopper: Using `sleep` is not ideal in terms of run time and reliability. See [test_batch_grouping_with_timeout](https://github.com/mlrun/storey/blob/de40cf6/tests/test_flow.py#L793) for an alternative.

storey/steps/sample.py
```diff
@@ -0,0 +1,67 @@
...
+        self._rate_seconds = rate_seconds
+
+        self._count = None
+        self._seconds = None
+
+    def _sample(self, event: Event):
+        return self._sample_by_count(event) or self._sample_by_seconds(event)
+
+    def _sample_by_count(self, event: Event):
+        if self._rate_count is None:
```
talIguaz: You can check these conditions in `_sample` and point top the relevant function

storey/steps/flatten.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def __init__(self, to_set: bool = False, **kwargs):
+        super().__init__(**kwargs)
+        self.to_set: bool = to_set
+
+    async def _do(self, event):
+        if event is _termination_obj:
+            return await self._do_downstream(_termination_obj)
+        else:
+            if self.to_set:
```
talIguaz: Why do we need to convert to a set at all? if a user want's a unique list let him have a set in his event, or use a Map step to do that

storey/steps/flatten.py
```diff
@@ -0,0 +1,34 @@
...
+                flat_elements = set(chain.from_iterable(event.body))
+            else:
+                flat_elements = chain.from_iterable(event.body)
```
talIguaz: Why not iterate over `event.body`? looks like from_iterable will do `['ABC', 'DEF'] --> A B C D E F` doesn't look like something we want

Michaelliv: hmm flattening `['ABC', 'DEF']` will definitely result in that, but why would you do that? You should use Flatten for `[['ABC'], ['ABC']] --> 'ABC' 'DEF'` or `[[1,2,3], [4,5,6], [7,8,9]] --> 1 2 3 4 5 6 7 8 9`.
Flatten is usually used for flattening a sequence of sequences...

storey/steps/flatten.py
Michaelliv: Yeah you are probably right about that, that's just a use case that came to mind. If anyone will need that they can probably just implement `FlattenUnique` or something.

storey/steps/flatten.py
Michaelliv: I mean that's by design... this is how flatten usually works

storey/steps/assertion.py
```diff
@@ -0,0 +1,149 @@
...
+        self.operator: _Operator = operator
+        self.actual: int = 0
+
+    def __call__(self, event):
+        self.actual += 1
+
+    def check(self):
+        assert self.operator(
+            self.actual, self.expected
+        ), f"Expected event count {self.operator} {self.expected}"
```
talIguaz: Better add the actual to the error message

storey/steps/flatten.py
```diff
@@ -10,25 +10,14 @@ class Flatten(Flow):
...
-            for flattened in flat_elements:
+            for flattened in chain.from_iterable(event.body):
                 await self._do_downstream(
-                    Event(
-                        body=flattened,
-                        key=event.key,
-                        time=event.time,
-                        id=event.id
-                    )
+                    Event(body=flattened, key=event.key, time=event.time, id=event.id)
```
talIguaz: You are discarding all other possible fields that can be in the Event

storey/steps/sample.py
```diff
@@ -28,40 +33,55 @@ def __init__(
...
         self._rate_count = rate_count
         self._rate_seconds = rate_seconds
+        self.emit_policy = emit_policy
 
         self._count = None
         self._seconds = None
 
     def _sample(self, event: Event):
-        return self._sample_by_count(event) or self._sample_by_seconds(event)
+        do_sample = False
```
talIguaz: Is it possible for the user to specify both `count` and `seconds`

storey/steps/sample.py
```diff
@@ -0,0 +1,67 @@
...
+            raise ValueError(f"Expected rate_count > 0, got {rate_count}")
+
+        if rate_seconds is not None and rate_seconds <= 0:
+            raise ValueError(f"Expected rate_seconds > 0, got {rate_seconds}")
+
+        self._rate_count = rate_count
+        self._rate_seconds = rate_seconds
+
+        self._count = None
+        self._seconds = None
```
talIguaz: rename to more meaningful names, something like `event_count` and `last_timestamp`

storey/steps/sample.py
```diff
@@ -0,0 +1,67 @@
...
+from storey import Filter, Event
+
+
+class Sample(Filter):
+    """Samples stream events, passes a single event downstream after either reaching 'rate_count' or 'rate_seconds'"""
+
+    def __init__(
+        self,
+        rate_count: Optional[int] = None,
+        rate_seconds: Optional[Union[int, float]] = None,
```
talIguaz: It's not really *rate* isn't it? more like `max_events` and `timeout` or `emit_after_seconds`
Also add docs for theses parameters

storey/steps/sample.py
```diff
@@ -0,0 +1,67 @@
...
+            self._count = 0
+            return True
+
+        self._count += 1
+        if self._count == self._rate_count:
+            self._count = None
+
+        return False
+
+    def _sample_by_seconds(self, event: Event):
```
talIguaz: don't really understand what this method does.
it wait's until the time of an event is `_rate_seconds` behind now, and then will "emit" the current event, not the one that is really behind/intended to emit

storey/dtypes.py
```diff
@@ -52,6 +53,23 @@ def __eq__(self, other):
     def __str__(self):
         return f'Event(id={self.id}, key={self.key}, time={self.time}, body={self.body})'
 
+    def copy(self, body=None, key=None, time=None, id=None, headers=None, method=None, path='/', content_type=None, awaitable_result=None,
+             deep_copy=False) -> 'Event':
```
gtopper, line 57: Shouldn't it be `-> Event`? Shouldn't `path=None` here?

storey/steps/flatten.py
```diff
@@ -0,0 +1,21 @@
...
+class Flatten(Flow):
+    """
+    Flattens sequences of sequences (i.e lists of lists, sets of sets, etc...) into a single sequence, by default
+    casts sequence to list. When setting to_set=True, casts sequence to set.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    async def _do(self, event: Event):
```
gtopper: This is an overridden internal method. No need for type annotation here.

storey/steps/partition.py
```diff
@@ -0,0 +1,29 @@
...
+from storey import Flow, Event
+from storey.dtypes import _termination_obj
+
+Partitioned = namedtuple("Partitioned", ["left", "right"], defaults=[None, None])
+
+
+class Partition(Flow):
+    """
+    Partitions events by calling a predicate function on each event. Each processed event results in a `Partitioned`
+    namedtuple of either (left=Event, right=None) or (left=None, right=Event).
```
gtopper: It's unclear whether True means left or right... I mean it's clear from the code, but not from the signature/docstring.

storey/steps/sample.py
```diff
@@ -0,0 +1,50 @@
+from enum import Enum
+
+from storey import Filter
+
+
+class Sample(Filter):
+    """
+    Samples stream events, passes a single event downstream accordingly to emit_policy. By default emit_policy is set to
```
gtopper: Samples events in accordance with `emit_policy`.

storey/steps/sample.py
```diff
@@ -0,0 +1,50 @@
...
+    EMIT_FIRST, meaning that for Sample(rate_count=n) - first event will be passed downstream, while the n-1 next events
+    are filtered out. Other wise if emit_policy is set to EMIT_LAST - the first n-1 events are filtered out, but event
```
gtopper: Otherwise,

gtopper: It's not clear what happens if the stream is terminated mid-period. "Last" would indicate that the last event will be emitted, but the n-1 description would indicate that it won't be. I'm not really sure why someone might want to emit "last" anyway. It's basically the same as emitted first but skipping the first event.

storey/steps/sample.py
```diff
@@ -0,0 +1,50 @@
...
+        self._emit_policy = emit_policy
+        self._count = None
+
+    def _sample(self):
+        if self._count is None:
+            self._count = 1
+            if self._emit_policy == self.EmitPolicy.EMIT_FIRST:
+                return True
+            elif self._emit_policy == self.EmitPolicy.EMIT_LAST:
+                return False
```
gtopper: I know it's an edge case, but this would break if `rate_count=1`.

gtopper: Also, if someone passes something other than `EmitPolicy` accidentally, some strange behaviour will ensue.

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+            [
+                Source(),
+                Assert().each_event(lambda event: event > 10)
+            ]
+        ).run()
+        controller.emit(1)
+        controller.terminate()
+        controller.await_termination()
+    except FlowError:
+        return True
```
gtopper: Need to check `__cause__` (see existing tests that do this).

gtopper: Also no point of returning any value in a test afaik.

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+        ).run()
+        controller.emit([1, 2, 3])
+        controller.terminate()
+        controller.await_termination()
+        fail("Assert not failing", False)
+    except FlowError:
+        pass
+
+
+def test_sample_by_count():
```
gtopper: Missing tests for alternative emit policy (if you want to keep it - I questioned the value in `EMIT_LAST` in an earlier comment).

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+            self.events.add(event.id)
+            return event
+
+    er = EventRegistry()
+    controller = build_flow(
+        [
+            Source(),
+            Map(er, full_event=True),
+            ForEach(lambda x: None),
+            Assert(full_event=True).each_event(lambda event: event.id in er.events),
```
gtopper: It would really be much easier to Reduce and check that the end result is as expected.

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+
+        if first is not None:
+            return first in dividable_by_two and first not in not_dividable_by_two and second is None
+        else:
+            return second in not_dividable_by_two and second not in dividable_by_two
+
+    controller = build_flow(
+        [
+            Source(),
+            Assert().exactly(6),
```
gtopper, line 394: Is this testing `Source`?

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+            Partition(lambda event: event.body % 2 == 0),
+            Assert().exactly(6),
+            Assert(full_event=True).each_event(lambda event: check_partition(event))
```
gtopper, line 397: Here too, much easier just reducing and asserting that the result is correct.

tests/test_steps.py
```diff
@@ -0,0 +1,376 @@
...
+    controller.emit(1)
+    controller.emit(1)
+
+    controller.terminate()
+    controller.await_termination()
+
+
+def test_partition():
+    dividable_by_two = {2, 4, 6}
+    not_dividable_by_two = {1, 3, 5}
```
gtopper: dividable->divisible

tests/test_steps.py
gtopper: Doing so would also verify the event order, and step termination.

storey/steps/flatten.py
```diff
@@ -0,0 +1,21 @@
+from itertools import chain
+
+from storey import Flow, Event
+from storey.dtypes import _termination_obj
+
+
+class Flatten(Flow):
+    """
+    Flattens sequences of sequences (i.e lists of lists, sets of sets, etc...) into a single sequence, by default
+    casts sequence to list. When setting to_set=True, casts sequence to set.
```
gtopper: Afaict this does not flatten lists of lists or any other data structure, but rather splits an event with an iterable body into multiple events. I'm not sure which behaviour is intended - the one documented or the one implemented.

storey/steps/assertion.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __call__(self, x, y):
+        return self.fn(x, y)
+
+    def __str__(self):
+        return self.str
+
+
+_EQUALS = _Operator("==", lambda x, y: x == y)
+_NOT_EQUALS = _Operator("!=", lambda x, y: x != y)
```
gtopper: _NOT_EQUALS -> _NOT_EQUAL

storey/steps/assertion.py
```diff
@@ -0,0 +1,149 @@
...
+_ABOVE = _Operator(">", lambda x, y: x > y)
+_BELOW = _Operator("<", lambda x, y: x < y)
+_ABOVE_OR_EQUAL = _Operator(">=", lambda x, y: x >= y)
+_BELOW_OR_EQUAL = _Operator("<=", lambda x, y: x <= y)
+
```
gtopper: _GREATER_THAN
_LESS_THAN
?
Never heard of these operators being refered to as above and below.

storey/steps/assertion.py
```diff
@@ -0,0 +1,149 @@
...
+_ANY = _Operator("any of", lambda x, y: any([i for i in x if i in y]))
```
gtopper: If checking for membership, should it be `any([i in y for i in x])`? Otherwise a member of x could be a value that evaluates to `False` (e.g. `False`, `0`, `''`).

storey/steps/assertion.py
```diff
@@ -0,0 +1,149 @@
...
+_ALL = _Operator("all of", lambda x, y: len([i for i in x if i in x]) == len(y))
```
gtopper: Same here, but also it looks like one of the `x`s what suppose to be a `y`.

gtopper: Oh, and `i` implies that this is a counter. Better choose another letter to imply that it's a member.

storey/steps/assertion.py
gtopper: It's also a bit confusing that ANY and ALL mean "any of the members in x are members of y" and "all members of x are members of y". It seems like these are set operations, so normally these would be refered to in terms of intersection and union and subsets.

storey/dtypes.py
Michaelliv: Should be `path=None` indeed, took defaults from constructor.
Can't do `-> Event`, Python 3.7 doesn't support referencing class type inside class.

storey/steps/flatten.py
Michaelliv: Is it bad? it's nice for auto-complete

storey/steps/sample.py
Michaelliv: No, If i want to sample in a rate of 1 per 1000 events, then first/last are quite different events. I'll rewrite the doc.

storey/steps/assertion.py
Michaelliv: hmm how about python's `any` and `all` functions?

storey/dtypes.py
gtopper: Ah ok, I see now that using a string works as well.

storey/steps/flatten.py
gtopper: I guess it's fine, only inconsistent with the rest of the code.

gtopper: I think adding the annotation to the base class will be better though, no?

storey/steps/assertion.py
gtopper: Those are boolean operations rather than set comparison operations.

storey/steps/flatten.py
Michaelliv: Just the small issue with `Union[Event, Any]` which is weird as we talked about earlier...

tests/test_steps.py
Michaelliv: Not really, it's more like "I'm assuring the input of this step"

storey/steps/assertion.py
Michaelliv: Unfortunately intersect, union and disjoint are not very concise in this case, hopefully my changes made it better now

storey/steps/assertion.py
```diff
@@ -0,0 +1,151 @@
...
+    def exactly(self, expected: int):
+        self.termination_assertions.append(_AssertEventCount(expected, _EQUALS))
+        return self
+
+    def match_exactly(self, expected: Collection[Any]):
+        self.termination_assertions.append(_AssertCollection(expected, _EXACTLY))
+        return self
+
+    def match_all_of(self, expected: Collection[Any]):
+        self.termination_assertions.append(_AssertCollection(expected, _CONTAINS_ALL_OF))
```
gtopper: is it "all of", "contains all of", or "match all of"?

storey/steps/partition.py
```diff
@@ -0,0 +1,34 @@
...
+
+Partitioned = namedtuple("Partitioned", ["left", "right"], defaults=[None, None])
+
+
+class Partition(Flow):
+    """
+    Partitions events by calling a predicate function on each event. Each processed event results in a `Partitioned`
+    namedtuple of (left=Optional[Event], right=Optional[Event]).
+
+    For a given event, if the predicate function results is `True`, the event is assigned to `left`. Otherwise, the
```
gtopper: results -> result

gtopper: or is -> in

storey/steps/partition.py
```diff
@@ -0,0 +1,34 @@
...
+    async def _do(self, event):
+        if event is _termination_obj:
+            return await self._do_downstream(_termination_obj)
+        else:
+            if self.predicate(event):
+                event.body = Partitioned(left=event.body, right=None)
+                await self._do_downstream(event)
+            else:
+                event.body = Partitioned(left=None, right=event.body)
+                await self._do_downstream(event)
```
gtopper: Can factor this line out.

storey/steps/partition.py
Michaelliv: Yeah typo

gtopper (approved):

PR #114, opened 2020-12-24
IG-17591 don't update cache on out of order events

dinal:


storey/aggregations.py
```diff
@@ -487,6 +487,7 @@ def __init__(self, name, aggregation, explicit_windows, hidden_windows, base_tim
                 for win in self.hidden_windows.windows:
                     if win not in all_windows:
                         all_windows.append(win)
+            self._need_to_recalculate_pre_aggregates = True
```
talIguaz, line 490: There is no real need to set it to True if you are calculating it on the same line, is just confusing a bit

talIguaz (approved):

PR #119, opened 2020-12-27
Fix wrong error in ConcurrentByKeyJobExecution.

gtopper:


tests/test_flow.py
```diff
@@ -1724,3 +1724,20 @@ def test_csv_reader_parquet_write_ns(tmpdir):
     read_back_df = pd.read_parquet(out_file, columns=columns)
 
     assert read_back_df.equals(expected), f"{read_back_df}\n!=\n{expected}"
+
+
+def test_error_in_concurrent_by_key_task():
```
talIguaz, line 1729: Better add a test with err and `Complete` as well

storey/flow.py
```diff
@@ -603,9 +603,10 @@ async def _safe_process_events(self, events):
             return await self._process_events(events)
         except BaseException as ex:
             for event in events:
-                none_or_coroutine = event._awaitable_result._set_result(ex)
-                if none_or_coroutine:
-                    await none_or_coroutine
+                if event._awaitable_result:
+                    none_or_coroutine = event._awaitable_result._set_result(ex)
```
talIguaz: We have `_set_error` for that

talIguaz (approved):

PR #120, opened 2020-12-27
surrounding saved engine words with backtick in update expressions

katyakats:


storey/drivers.py
```diff
@@ -179,6 +179,11 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
         condition_expression = None
         expressions = []
         pending_updates = {}
+        saved_engine_words = ['min', 'max', 'sqrt', 'avg', 'stddev', 'sum', 'length', 'init_array', 'set', 'remove',
+                              'regexp_replace', 'find', 'find_all_indices', 'extract_by_indices', 'sort_array', 'blob',
+                              'in', 'true', 'false', 'and', 'or', 'not', 'contains', 'ends', 'starts', 'exists', 'select',
+                              'str', 'to_int', 'if_not_exists', 'regexp_instr', 'delete', 'as', 'maxdbl', 'inf', 'nan',
+                              'isnan', 'nanaszero', 'IF_ELSE']
```
dinal: 1. better save the words in a dictionary instead of list so you get 0(1) check instead of iterating the list each time
2. if you always compare lower case to lower case then you can save 'IF_ELSE' as 'if_else'

storey/drivers.py
```diff
@@ -193,6 +198,8 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
         # Generating additional cache expressions
         if additional_data:
             for name, value in additional_data.items():
+                if name.casefold() in (word.casefold() for word in saved_engine_words):
```
dinal: if all words will be kept in lower case then no need for word.casefold()

integration/test_flow_integration.py
```diff
@@ -269,7 +269,7 @@ def test_join_by_key_specific_attributes(setup_kv_teardown_test):
 def test_write_table_specific_columns(setup_teardown_test):
     table = Table(setup_teardown_test, V3ioDriver())
 
-    table._cache['tal'] = {'color': 'blue', 'age': 41, 'iss': True, 'sometime': test_base_time}
+    table._cache['tal'] = {'color': 'blue', 'age': 41, 'iss': True, 'sometime': test_base_time, 'min': 1}
```
dinal: add upper case attribute as well

storey/drivers.py
```diff
@@ -179,6 +179,14 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
         condition_expression = None
         expressions = []
         pending_updates = {}
+        saved_engine_words = {'min': True, 'max': True, 'sqrt': True, 'avg': True, 'stddev': True, 'sum': True,
```
dinal: move declaration to class level

dinal (approved):

PR #126, opened 2020-12-31
make s3 dependency optional

talIguaz:


storey/utils.py
```diff
@@ -105,4 +105,19 @@ def _split_path(path):
...
     return fsspec.filesystem(schema), parsed_url.path
+
+
+def load_dependencies(schema):
+    if schema == "s3":
+        try:
+            import s3fs  # noqa: F401
+        except ImportError:
+            raise StoreyMissingDependencyError(
+                "s3 packages are missing, use pip install mlrun[s3]"
```
gtopper: mlrun?

storey/utils.py
```diff
@@ -105,4 +105,19 @@ def _split_path(path):
 def url_to_file_system(url):
     parsed_url = urlparse(url)
     schema = parsed_url.scheme.lower()
+    load_dependencies(schema)
     return fsspec.filesystem(schema), parsed_url.path
+
+
+def load_dependencies(schema):
```
gtopper: Generic name, but actually this is only for s3.

talIguaz: will be used for other fs in the future, like azure

storey/utils.py
talIguaz: 🙈

setup.py
```diff
@@ -23,6 +23,10 @@ def load_deps(file_name):
 
 install_requires = load_deps('requirements.txt')
 tests_require = load_deps('dev-requirements.txt')
+extras_require = {
+    "s3": ["s3fs~=0.5.2"],
```
gtopper: `~=0.5`?

storey/utils.py
gtopper: Probably would've been better named as `load_filesystem_dependencies`  or `load_fs_dependencies `

gtopper (approved):

PR #127, opened 2021-01-04
move AggregationStore inside Table class and handle both static and a…

dinal:
…ggregations cache in one place


storey/table.py
```diff
@@ -12,58 +18,53 @@ class Table:
          can optimize writes. Defaults to True.
         """
 
-    def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool = True):
+    def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool = True, use_windows_from_schema: bool = False):
```
talIguaz: Why do we need `use_windows_from_schema` in the init?

storey/table.py
```diff
@@ -12,58 +18,53 @@ class Table:
...
 
-    def __getitem__(self, key):
-        return self._cache[key]
-
-    def __setitem__(self, key, value):
-        self._cache[key] = value
-
-    def update_key(self, key, data):
-        if key in self._cache:
+    def update_static_attrs(self, key, data):
```
talIguaz: I would consider keeping `__getitem__` and `__setitem__` to get/update the "static" fields. and treat the aggregations in a special way by calling `get_features` or something.

This is just more intuitive for users that will use the `Table` object

storey/table.py
```diff
@@ -74,3 +75,625 @@ async def persist_key(self, key, event_data_to_persist):
...
+
+    async def _get_or_load_key(self, key, timestamp=None):
+        if self._read_only or not self.get_aggregations_attrs(key):
+            # Try load from the store, and create a new one only if the key really is new
+            initial_data = await self._storage._load_aggregates_by_key(self._container, self._table_path, key)
+            self.set_aggregations_attrs(key, AggregatedStoreElement(key, self._aggregates, timestamp, initial_data))
+
+        return self.get_aggregations_attrs(key)
+
+    async def add_key(self, key, base_timestamp, initial_data):
```
talIguaz: rename to something more relevant like `add_aggregation_by_key`

storey/table.py
```diff
@@ -12,58 +18,53 @@ class Table:
...
             # Create new aggregation element
-            await self._aggregation_store.add_key(key, timestamp, aggregate_initial_data)
+            await self.add_key(key, timestamp, aggregate_initial_data)
 
             if additional_data:
                 # Add additional data to simple cache
-                self.update_key(key, additional_data)
+                self.update_static_attrs(key, additional_data)
 
     async def get_or_load_key(self, key, attributes='*'):
```
talIguaz: Rename to something like `get_or_load_static_attributes_by_key`

storey/table.py
```diff
@@ -74,3 +75,625 @@ async def persist_key(self, key, event_data_to_persist):
...
+
+        self.get_aggregations_attrs(key).aggregate(data, timestamp)
+
+    async def get_features(self, key, timestamp):
+        if not self._schema:
+            await self.get_or_save_schema()
+
+        return self.get_aggregations_attrs(key).get_features(timestamp)
+
+    async def _get_or_load_key(self, key, timestamp=None):
```
talIguaz: rename to mention that it refers to aggregations

storey/table.py
dinal: i disagree, i think it should be clear what you're setting now that table is handling both types of attrs
it's not intuitive that the default it static imo

storey/flow.py
```diff
@@ -278,10 +278,11 @@ async def _call(self, event):
         element = self._get_event_or_body(event)
         if self._group_by_key:
             if isinstance(self._state, Table):
-                key_data = await self._state.get_or_load_key(event.key)
+                key_data = await self._state.get_or_load_static_attributes_by_key(event.key)
             else:
                 key_data = self._state[event.key]
-            res, self._state[event.key] = self._fn(element, key_data)
+            res, mapped_event = self._fn(element, key_data)
```
talIguaz: `mapped_event` -> `new_state` ?

storey/table.py
```diff
@@ -74,3 +75,625 @@ async def persist_key(self, key, event_data_to_persist):
...
+                raise ValueError(
+                    f'Requested aggregates for feature {aggr.name} do not match with existing aggregates at {self._table_path}. '
+                    f"Requested: {aggr.aggregations}, existing: {schema_aggr['aggregates']}")
+            # Check if more raw aggregates are requested, in which case a schema update is required
+            if not self._read_only and requested_raw_aggregates != existing_raw_aggregates:
+                should_update = True
+
+        return should_update
+
+    async def _save_key(self, key):
```
talIguaz: Is it saving just aggregations? better rename as well

storey/table.py
```diff
@@ -74,3 +75,625 @@ async def persist_key(self, key, event_data_to_persist):
...
+        return self._attrs_cache.keys()
+
+
+class _CacheElement:
+    def __init__(self, static_attrs, aggregations):
+        self.static_attrs = static_attrs
+        self.aggregations = aggregations
+
+
+class AggregatedStoreElement:
```
talIguaz, line 229: Why did you moved all of this here as well (from this line forward)? doesn't seem to belong here

storey/table.py
```diff
@@ -15,55 +21,50 @@ class Table:
     def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool = True):
         self._container, self._table_path = _split_path(table_path)
         self._storage = storage
-        self._cache = {}
-        self._aggregation_store = None
+        self._attrs_cache = {}
         self._partitioned_by_key = partitioned_by_key
+        self._aggregates = None
+        self._schema = None
+        self._read_only = False
```
talIguaz: Isn't `read_only` only affect the aggregations? either rename it or move it somewhere else

talIguaz:
Also, please go over all methods inside `Table` and make everything internal that the user shouldn't call directly as private (`_<method>`)

storey/table.py
dinal: leaving it in aggregations.py created cyclic dependency 
also it's internal to store impl, it makes sense it will be in the same place

tests/test_flow.py
```diff
@@ -460,8 +460,8 @@ def test_map_with_state_flow():
 
 def test_map_with_cache_state_flow():
     table_object = Table("table", NoopDriver())
-    table_object.set_static_attrs('tal', {'color': 'blue'})
```
talIguaz: In the tests you can use `table_object['tal'] = {'color': 'blue'}` as it was before 😄

talIguaz (approved):

PR #130, opened 2021-01-06
Fix hang in WriteToTable.

gtopper:


storey/flow.py
```diff
@@ -532,18 +532,19 @@ def __init__(self, max_in_flight=8, **kwargs):
...
                         if pending_event.pending and not pending_event.in_flight:
-                            resp = await self._process_event(pending_event.pending)
+                            resp = await self._safe_process_events(pending_event.pending)
                             for event in pending_event.pending:
                                 await self._handle_completed(event, resp)
                     if self._q.empty():
                         break
                     else:
-                        await self._q.put(_termination_obj)
+                        job = _termination_obj
```
talIguaz: `job` is already `termination_obj`, no?

storey/flow.py
```diff
@@ -559,11 +560,12 @@ async def _worker(self):
                     self._pending_by_key[event.key].pending = []
 
                     task = self._safe_process_events(self._pending_by_key[event.key].in_flight)
-                    await self._q.put((event, asyncio.get_running_loop().create_task(task)))
+                    job = (event, asyncio.get_running_loop().create_task(task))
```
talIguaz: Well what you are doing here is just prioritizing a specific key over the others.
If all keys keep coming all the time, you end up only processing `key1` first,  then `key2` etc

storey/flow.py
gtopper: Hm yeah, maybe this is not needed? Why would q not be empty?

gtopper: I replaced the if/else with just a break.

storey/flow.py
gtopper: Well, a bit... The producer will still block on the queue if there are more keys than queue slots. So there's still back pressure. Do we care about the exact prioritization of keys when writing to a table?

storey/flow.py
```diff
@@ -531,20 +531,29 @@ def __init__(self, max_in_flight=8, **kwargs):
...
+                jobs = self_sent_jobs.pop(received_job_count, None)
+                if jobs:
+                    job = jobs[0]
+                    if len(jobs) > 1:
+                        self_sent_jobs[received_job_count] = jobs[1:]
+                else:
+                    job = await self._q.get()
+                    received_job_count += 1
+                    if job is _termination_obj:
+                        if received_job_count in self_sent_jobs:
```
talIguaz, line 548: Why would `received_job_count in self_sent_jobs` ? you are checking it at L539 and nothing else but this function isn’t adding to `self_sent_jobs` so it should not exist there, no?

storey/flow.py
```diff
@@ -559,11 +568,14 @@ async def _worker(self):
                     self._pending_by_key[event.key].pending = []
 
                     task = self._safe_process_events(self._pending_by_key[event.key].in_flight)
-                    await self._q.put((event, asyncio.get_running_loop().create_task(task)))
+                    tail_position = received_job_count + self._q.qsize()
+                    jobs_at_tail = self_sent_jobs.get(tail_position, [])
```
talIguaz, line 572: Aren't you always gonna miss the get and init with []?

talIguaz (approved):

PR #138, opened 2021-01-14
Refactor aggregates

talIguaz:
* split AggregateStoreElement to two. for QueryAggregateByKey and AggregateByKey
* The latter implementation is optimized to use less allocations and computations.
* added a new benchmark, based on the previous early sense but with more basic aggregates which we are more optimized in.

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -721,9 +721,9 @@ class AggregatedStoreElement:
         initial_data_by_feature = {}
         if initial_data:
             for key, value in initial_data.items():
-                seperator_index = key.rindex('_')
-                feature_name = key[:seperator_index]
-                aggr = key[seperator_index + 1:]
+                separator_index = key.rindex('_')
+                feature_name = key[:separator_index]
+                aggr = key[separator_index + 1:]
                 if feature_name not in initial_data_by_feature:
                     initial_data_by_feature[feature_name] = {}
                 initial_data_by_feature[feature_name][aggr] = value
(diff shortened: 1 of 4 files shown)


storey/table.py
```diff
@@ -703,3 +708,381 @@ def _get_aggregation_default_value(aggregation):
...
+
+                if bucket_index < 0:
+                    return
+
+        # Initialize every remaining buckets
+        for i in range(bucket_index + 1):
+            for aggregation in self._all_raw_aggregates:
+                self.buckets[i][aggregation] = AggregationValue(aggregation, self.max_value)
+
+    def get_and_flush_pending(self):
```
gtopper, line 1079: To be fair that's more like "get and reset pending" :)

storey/table.py
```diff
@@ -703,3 +708,381 @@ def _get_aggregation_default_value(aggregation):
...
+        self.aggregates = aggregates
+        self.storage_specific_cache = {}
+
+        # Group init data by feature name
+        initial_data_by_feature = {}
+        if initial_data:
+            for key, value in initial_data.items():
+                seperator_index = key.rindex('_')
+                feature_name = key[:seperator_index]
+                aggr = key[seperator_index + 1:]
```
gtopper: seperator_index -> separator_index :)

storey/table.py
```diff
@@ -703,3 +708,381 @@ def _get_aggregation_default_value(aggregation):
...
+
+        if self._need_to_recalculate_pre_aggregates or \
+                current_time_bucket_index < self.get_bucket_index_by_timestamp(self._last_data_point_timestamp) or \
+                not self._precalculated_aggregations:
+            result = self.calculate_features(timestamp)
+        else:
+            # In case our pre aggregates already have the answer
+            for (aggregation_name, _, window_str), aggregation in self._current_aggregate_values.items():
+                if aggregation_name in self._explicit_raw_aggregations:
+                    result[f'{self.name}_{aggregation_name}_{window_str}'] = aggregation.get_value()[1]
```
gtopper: I wonder if this can't be done more efficiently.

gtopper: More like lines 977-978.

gtopper (approved):
LGTM @talIguaz 💔

storey/table.py
```diff
@@ -703,3 +708,382 @@ def _get_aggregation_default_value(aggregation):
...
+                for aggr in self._all_raw_aggregates:
+                    self._current_aggregate_values[(aggr, window_millis, windows_str)] = AggregationValue(aggr)
+
+        if initial_data:
+            self.last_bucket_start_time = None
+
+            # Initializing the buckets from the stored data and calculating the initial pre aggregates
+            self.initialize_from_data(initial_data, base_time)
+            self._need_to_recalculate_pre_aggregates = True
+            self.calculate_features(base_time)
```
gtopper, line 808: Hm this is strange... Shouldn't the result be used by something?

dinal: it's to init the pre aggregated values in self._current_aggregate_values

gtopper: Might need some refactoring then.

PR #144, opened 2021-01-21
Aggregation optimizations.

gtopper:
Benchmark:
```
import time
import uuid

import pandas as pd

from storey import Table, Driver, build_flow, AggregateByKey, FieldAggregator, DataframeSource, WriteToTable
from storey.dtypes import SlidingWindows

table = Table(f'bigdata/talexa{uuid.uuid4().hex}', Driver())
df = pd.read_csv('~/Downloads/early_sense.csv')  # .head(1000)
df['timestamp'] = pd.to_datetime(df['timestamp'])
df.sort_values(by=['patient_id', 'timestamp'], inplace=True)
start = time.time()
controller = build_flow([
    DataframeSource(df, key_column='patient_id', time_column='timestamp'),
    AggregateByKey([
        FieldAggregator("hr", "hr", ["avg", "min", "max"],
                        SlidingWindows(['1h', '2h'], '10m')),
        FieldAggregator("rr", "rr", ["avg", "min", "max"],
                        SlidingWindows(['1h', '2h'], '10m')),
        FieldAggregator("spo2", "spo2", ["avg", "min", "max"],
                        SlidingWindows(['1h', '2h'], '10m'))
    ],
        table),
    WriteToTable(table, columns=['room', 'bad', 'department']),
]).run()

end = time.time()
print(end - start)
```

Runtime on dev (avg of 3 runs): 15.097
Runtime on this branch (avg of 3 runs): 9.633
That's a speed up of 56.7% (or a reduction in runtime of %36.2.

Update: down to 8.879 - 70% speed up.
Update 2: down to 8.433 - 79% speed up.


storey/table.py
```diff
@@ -666,48 +666,156 @@ def __init__(self, aggregation, max_value=None, set_data=None):
...
+    name = 'min'
+    default_value = float('inf')
+
+    def __init__(self, max_value=None, set_data=None):
+        self._value = self.default_value
+        super().__init__(max_value, set_data)
+
+    def aggregate(self, time, value):
+        if value > self._value:
+            value = self._value
```
dinal: there's no need to call _set_value in this case no?
the value hasn't changed, it's still self._value

it should be something like 
if value < self._value:
     self._set_value(value)

storey/table.py
```diff
@@ -666,48 +666,156 @@ def __init__(self, aggregation, max_value=None, set_data=None):
...
+
+class MaxValue(AggregationValue):
+    name = 'max'
+    default_value = float('-inf')
+
+    def __init__(self, max_value=None, set_data=None):
+        self._value = self.default_value
+        super().__init__(max_value, set_data)
+
+    def aggregate(self, time, value):
```
dinal, line 743: same comment as in MinValue

dinal:
nice job!!

storey/table.py
gtopper: Oh, that's right! Even better...

storey/table.py
```diff
@@ -685,12 +686,15 @@ def new_from_name(aggregation, max_value=None, set_data=None):
         elif aggregation == 'first':
             return FirstValue(max_value, set_data)
 
-    def _set_value(self, value):
-        if self._max_value and value > self._max_value:
+    def _set_value_with_max(self, value):
+        if value > self._max_value:
```
dinal, line 690: is it really needed? 
you anyway have to check if max_value, why does it matter if it's done in _set_value or in the calling function?
it's just complicates the logic imo

gtopper: Because it happens once in init instead of 5M times.

PR #145, opened 2021-01-27
Allow steps to self-describe and be reused.

gtopper:


tests/test_flow.py
```diff
@@ -1942,3 +1942,22 @@ def test_result_path():
     controller.terminate()
     termination_result = controller.await_termination()
     assert termination_result == {'col1': 1, 'step_result': {"new_field": 5}}
+
+
+def test_to_dict():
```
dinal, line 1947: add to this test more steps, better for all the main steps

dinal (commented):
can you verify that you don't override fields in kwargs?
like if some step as an input_path one might accidentally override flow's input_path

dinal (approved):

PR #147, opened 2021-01-27
Optimization.

gtopper:
Speedup: 15%

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -402,8 +402,8 @@ class ReadOnlyAggregationBuckets:
                 current_window_start = min(len(self.buckets), current_window_start)
 
                 for bucket_id in range(previous_window_start, current_window_start):
-                    current_pre_aggregated_value = aggr.get_value()[1]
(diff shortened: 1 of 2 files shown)


storey/table.py
```diff
@@ -525,12 +525,12 @@ def calculate_features(self, timestamp, windows):
 
             for bucket_index in range(current_time_bucket_index, last_bucket_to_aggregate - 1, -1):
                 if bucket_index < len(self.buckets):
-                    t, v = self.buckets[bucket_index].get_value()
-                    aggregated_value.aggregate(t, v)
+                    bucket = self.buckets[bucket_index]
+                    aggregated_value.aggregate(bucket.last_time, bucket.value)
```
katyakats: what if self.aggregation == first? in this case get_value() will return first_time and not last_time

gtopper: True... Kind of disturbing that no existing test caught that.

katyakats (approved):

PR #154, opened 2021-02-06
add azure support

dinal:


integration/test_azure_filesystem_integration.py
```diff
@@ -0,0 +1,241 @@
...
+        FlatMap(lambda x: x),
+        Map(lambda x: int(x)),
+        Reduce(0, lambda acc, x: acc + x),
+    ]).run()
+
+    termination_result = controller.await_termination()
+    assert termination_result == 495
+
+
+@pytest.mark.skipif(not has_azure_credentials, reason='No zure credentials found')
```
katyakats, line 87: azure, not zure

katyakats (approved):

integration/test_azure_filesystem_integration.py
dinal: i'll fix it in next pr :)

PR #156, opened 2021-02-07
Lock schema and each key in Table.

gtopper:


storey/table.py
```diff
@@ -97,8 +141,10 @@ def _new_aggregated_store_element(self):
 
     async def add_aggregation_by_key(self, key, base_timestamp, initial_data):
         if not self._schema:
-            await self._get_or_save_schema()
-        self._set_aggregations_attrs(key, self._new_aggregated_store_element()(key, self._aggregates, base_timestamp, initial_data))
+            async with self._get_schema_lock():
```
dinal: this function is called only from _lazy_load_key_with_aggregates and it will be already locked from there

storey/table.py
```diff
@@ -44,6 +46,20 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
     def __str__(self):
         return f'{self._container}/{self._table_path}'
 
+    def _get_lock(self, key):
+        key = get_hashed_key(key)
```
dinal: no need to hash it, it's already hashed in all the places the call this function

storey/table.py
```diff
@@ -44,6 +46,20 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
...
+        cache_element = self._attrs_cache.get(key)
+        if cache_element is None:
+            cache_element = _CacheElement({}, None)
+            self._attrs_cache[key] = cache_element
+        cache_element.lock = Lock()
+        return cache_element.lock
+
+    def _get_schema_lock(self):
+        if self._schema_lock is None:
```
dinal, line 59: why not create the lock in ctor instead checking this if each time

storey/table.py
```diff
@@ -109,29 +129,32 @@ async def _aggregate(self, key, data, timestamp):
...
     async def _get_features(self, key, timestamp):
         if self._flush_exception is not None:
             raise self._flush_exception
         if not self._schema:
-            await self._get_or_save_schema()
+            async with self._get_schema_lock():
+                await self._get_or_save_schema()
 
-        attrs = self._get_aggregations_attrs(key)
+        async with self._get_lock(key):
```
dinal: do you need the lock here? it only reads stuff without changing state

storey/table.py
```diff
@@ -109,29 +129,32 @@ async def _aggregate(self, key, data, timestamp):
...
+            cache_item.aggregate(data, timestamp)
+            self._changed_keys.add(get_hashed_key(key))
 
     async def _get_features(self, key, timestamp):
         if self._flush_exception is not None:
             raise self._flush_exception
         if not self._schema:
-            await self._get_or_save_schema()
+            async with self._get_schema_lock():
+                await self._get_or_save_schema()
```
dinal: better move the lock inside the function
_aggregate calls _get_or_save_schema as well

storey/table.py
```diff
@@ -44,6 +46,20 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
     def __str__(self):
         return f'{self._container}/{self._table_path}'
 
+    def _get_lock(self, key):
+        key = get_hashed_key(key)
+        cache_element = self._attrs_cache.get(key)
+        if cache_element is None:
+            cache_element = _CacheElement({}, None)
+            self._attrs_cache[key] = cache_element
+        cache_element.lock = Lock()
```
dinal: this should be in _CacheElement ctor

storey/writers.py
```diff
@@ -670,6 +670,7 @@ async def _do(self, event):
             await self._table._persist(_PersistJob(event.key, data_to_persist, self._handle_completed, event))
         else:
             data_to_persist = self._event_to_writer_entry(event)
-            self._table._update_static_attrs(event.key, data_to_persist)
+            async with self._table._get_lock(event.key):
```
dinal: better move the lock inside the function no? so we won't miss calls to update_static_attrs
like in _FunctionWithStateFlow

dinal (commented):
add a test with diverged flow that changes the table

storey/table.py
gtopper: That wouldn't work, because it needs the event loop.

storey/table.py
gtopper: I guess not, but probably you'll need it once you LRU it.

storey/table.py
gtopper: This gets created outside the storey flow (e.g. via `table['abc'] = ...` and in tests), so an asyncio lock can't be created there.

storey/writers.py
gtopper: No can do, because it gets called in `_lazy_load_key_with_aggregates`, and trying to obtain the lock again will deadlock. But I can add locking in `_FunctionWithStateFlow`.

storey/table.py
dinal: then you should move it inside the if so it won't be created each time

storey/table.py
```diff
@@ -222,14 +244,14 @@ def _aggregates_to_schema(self):
...
-        key = get_hashed_key(key)
+        key = key
         if key in self._attrs_cache:
             return self._attrs_cache[key].aggregations
         else:
             return None
 
     def _set_aggregations_attrs(self, key, element):
-        key = get_hashed_key(key)
+        key = key
```
dinal: ?

storey/table.py
```diff
@@ -238,14 +260,14 @@ def _set_aggregations_attrs(self, key, element):
             self._attrs_cache[key] = _CacheElement({}, element)
 
     def _get_static_attrs(self, key):
-        key = get_hashed_key(key)
+        key = key
```
dinal: ?

storey/table.py
```diff
@@ -238,14 +260,14 @@ def _set_aggregations_attrs(self, key, element):
...
         if key in self._attrs_cache:
             return self._attrs_cache[key].static_attrs
         else:
             return None
 
     def _set_static_attrs(self, key, value):
-        key = get_hashed_key(key)
+        key = key
```
dinal: ?

storey/table.py
```diff
@@ -44,54 +46,72 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
...
-                self._update_static_attrs(key, additional_data)
+                if additional_data:
+                    # Add additional data to simple cache
+                    self._update_static_attrs(key, additional_data)
 
     async def _get_or_load_static_attributes_by_key(self, key, attributes='*'):
         if self._flush_exception is not None:
             raise self._flush_exception
-        key = get_hashed_key(key)
+        key = key
```
dinal: what's this?

storey/table.py
```diff
@@ -44,54 +46,72 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
...
+        self._changed_keys.add(key)
 
     async def _lazy_load_key_with_aggregates(self, key, timestamp=None):
         if self._flush_exception is not None:
             raise self._flush_exception
-        key = get_hashed_key(key)
-        if self._aggregations_read_only or not self._get_aggregations_attrs(key):
-            # Try load from the store, and create a new one only if the key really is new
-            aggregate_initial_data, additional_data = await self._storage._load_aggregates_by_key(self._container, self._table_path, key)
+        key = key
```
dinal: what's this?

dinal:
all the places that you removed "key = key" should be changed to "key = stringify_key(key)"

dinal (approved):

PR #157, opened 2021-02-08
sphinx documentation: first commit

katyakats:


Makefile
```diff
@@ -26,9 +26,11 @@ integration:
 env:
 	pipenv install -r requirements.txt
 
+
```
gtopper: Accidental newline.

Makefile
```diff
@@ -37,3 +39,9 @@ dist: dev-env
 .PHONY: set-version
 set-version:
 	python set-version.py
+
+.PHONY: html-docs
+html-docs: ## Build html docs
```
gtopper: I know I'm nitpicking but what's with the double #?

gtopper: Also I think just docs would be a better name than html-docs.

katyakats: the double # are from mlrun :)))

docs/index.rst
```diff
@@ -0,0 +1,20 @@
+
+
```
gtopper: Accidental newlines?

Makefile
```diff
@@ -29,6 +29,7 @@ env:
 .PHONY: dev-env
 dev-env: env
 	pipenv install -r dev-requirements.txt
+	pipenv install -r docs/requirements.txt
```
gtopper, line 32:
```
pipenv install -r dev-requirements.txt -r docs/requirements.txt
```

gtopper (approved):

PR #169, opened 2021-02-14
flow/check_stage_in_flow need to check recovery step as well

katyakats:


storey/flow.py
```diff
@@ -206,6 +206,18 @@ def _check_stage_in_flow(self, type_to_check):
                 return True
             if step._check_stage_in_flow(type_to_check):
                 return True
+        if self._recovery_step is not None:
+            if isinstance(self._recovery_step, type_to_check):
+                return True
+            if isinstance(self._recovery_step, dict):
+                for step in self._recovery_step.values():
+                    if step is type_to_check:
```
gtopper: Looks like a bug 🔎

dinal:
pls add a test

storey/flow.py
```diff
@@ -206,6 +206,18 @@ def _check_stage_in_flow(self, type_to_check):
...
+                return True
+            if isinstance(self._recovery_step, dict):
+                for step in self._recovery_step.values():
+                    if isinstance(step, type_to_check):
+                        return True
+                    if step._check_stage_in_flow(type_to_check):
+                        return True
+            else:
+                if self._recovery_step._check_stage_in_flow(type_to_check):
+                    return True
```
gtopper: I think this impl would be a bit more straightforward. Note that we don't use the term "stage" in storey.
```
    def _check_step_in_flow(self, type_to_check):
        if isinstance(self, type_to_check):
            return True
        for outlet in self._outlets:
            if outlet._check_step_in_flow(type_to_check):
                return True
        if isinstance(self._recovery_step, Flow):
            if self._recovery_step._check_step_in_flow(type_to_check):
                return True
        elif isinstance(self._recovery_step, dict):
            for step in self._recovery_step.values():
                if step._check_step_in_flow(type_to_check):
                    return True
        return False
```

tests/test_flow.py
```diff
@@ -2108,3 +2108,23 @@ def test_writer_downstream(tmpdir):
...
+def test_complete_in_error_flow():
+    reduce = build_flow([
+        Complete(),
+        Reduce(0, lambda acc, x: acc + x)
+    ])
+    controller = build_flow([
+        Source(),
+        Map(lambda x: x + 1),
+        Map(RaiseEx(5).raise_ex, recovery_step=reduce),
+        reduce
```
gtopper: It's probably a good idea to make the recovery step different from the normal outlet, so that the test verifies that it actually went into recovery. E.g. add a `Map(lambda x: x+100)` to the recovery flow.

gtopper (approved):
LGTM but see suggestion for improving the added test.

PR #173, opened 2021-02-25
align aiohttp and numpy reqs to jupy env

dinal:


gtopper (approved):
Ok, Uri says aiohttp 3.6.3 is installed for a conda package called kfp, which is a part of kubeflow. We'll get a lot of warnings in Python 3.8 and above, but our jupyter uses 3.7.6, so it won't affect everyone all the time.

PR #174, opened 2021-02-25
in case key does not exist in queryByKey do not throw an exception

katyakats:


tests/test_flow.py
```diff
@@ -2153,3 +2153,24 @@ def test_complete_in_error_flow():
...
+
+
+def test_non_existing_key_query_by_key():
+    df = pd.DataFrame([['katya', 'green'], ['dina', 'blue']], columns=['name', 'color'])
+    table = Table('table', NoopDriver())
+    build_flow([
+        DataframeSource(df),
+        WriteToTable(table),
+        Complete()
+    ]).run()
```
gtopper: Need to await termination.

tests/test_flow.py
```diff
@@ -2153,3 +2153,24 @@ def test_complete_in_error_flow():
...
     assert termination_result == 5005
+
+
+def test_non_existing_key_query_by_key():
+    df = pd.DataFrame([['katya', 'green'], ['dina', 'blue']], columns=['name', 'color'])
+    table = Table('table', NoopDriver())
+    build_flow([
+        DataframeSource(df),
+        WriteToTable(table),
+        Complete()
```
gtopper: `Complete` does nothing here.

tests/test_flow.py
```diff
@@ -2153,3 +2153,24 @@ def test_complete_in_error_flow():
...
+    ]).run()
+
+    controller = build_flow([
+        Source(),
+        QueryByKey(["color"],
+                   table, key="name"),
+        Complete(),
```
gtopper: No point to `Complete` if you're not going to check the result.

storey/aggregations.py
```diff
@@ -248,15 +248,16 @@ async def _do(self, event):
...
                 key = self.key_extractor(element)
-            await self._emit_event(key, event)
-
-        except Exception as ex:
-            raise ex
+            except Exception:
+                # in case of an error, emit event without key and features
+                await self._do_downstream(event)
+                return
+        await self._emit_event(key, event)
```
gtopper: Need to use `get` instead of [] to avoid the exception being thrown in the first place.

storey/aggregations.py
```diff
@@ -248,15 +248,14 @@ async def _do(self, event):
...
-            await self._emit_event(key, event)
-
-        except Exception as ex:
-            raise ex
+        element = event.body
+        key = event.key
+        if self.key_extractor:
+            key = self.key_extractor(element)
+            if key is None:
+                await self._do_downstream(event)
```
gtopper, line 257: You push the original event downstream? That doesn't seem correct...?

storey/aggregations.py
```diff
@@ -248,15 +248,14 @@ async def _do(self, event):
...
 
-        try:
-            element = event.body
-            key = event.key
-            if self.key_extractor:
-                key = self.key_extractor(element)
-            await self._emit_event(key, event)
-
-        except Exception as ex:
-            raise ex
```
gtopper, line 259: Weird that this was here to begin with.

tests/test_flow.py
```diff
@@ -2153,3 +2153,23 @@ def test_complete_in_error_flow():
...
+    controller = build_flow([
+        DataframeSource(df),
+        WriteToTable(table),
+    ]).run()
+    controller.await_termination()
+
+    controller = build_flow([
+        Source(),
+        QueryByKey(["color"],
+                   table, key="name"),
```
gtopper: I think this fits on one line :)

gtopper (approved):

PR #175, opened 2021-02-28
Add support for aggregations over sparse-feature events.

gtopper:


dinal (commented):
can you change the test so the 2 features will progress in different time deltas?

dinal (approved):

PR #177, opened 2021-03-07
Demo-related improvements.

gtopper:
* Allow sparse data in writers, and handle it appropriately in each writer.
* Rename EmitBase -> EmitPolicy
* Raise type error when EmitPolicy is of the wrong type (such as when `EmitEveryEvent` is passed in instead of `EmitEveryEvent()`).
* Report the offending update expression on error where relevant.

### DIFF
--- a/storey/drivers.py
+++ b/storey/drivers.py
@@ -13,6 +13,7 @@ from .utils import schema_file_name
 
 class Driver:
     """Abstract class for database connection"""
+
     async def _save_schema(self, container, table_path, schema):
         pass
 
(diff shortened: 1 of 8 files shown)


storey/drivers.py
```diff
@@ -158,7 +160,9 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
 
         if should_raise_error:
             raise V3ioError(
-                f'Failed to save aggregation for {table_path}/{key}. Response status code was {response.status_code}: {response.body}')
+                f'Failed to save aggregation for {table_path}/{key}. Response status code was {response.status_code}: {response.body}\n'
+                f'Update expression was: {update_expression}'
```
dinal, line 164: why did you remove the update expr? it's easier to debug with it

gtopper: That's why I added it, not remove it 🤔

katyakats (approved):

storey/drivers.py
dinal: lol :D

PR #182, opened 2021-03-11
Multiple keys

katyakats:


storey/dataframe.py
```diff
@@ -44,7 +45,12 @@ async def _do(self, event):
         if event is _termination_obj:
             df = pd.DataFrame(self._data, columns=self._columns)
             if self._insert_key_column_as:
-                df[self._insert_key_column_as] = self._key_column
+                if isinstance(self._insert_key_column_as, list):
+                    split_key_columns = [item.split('.') for item in self._key_column]
```
dinal, line 49: item name can have several dots, we want to split only on the last one

storey/sources.py
```diff
@@ -63,14 +63,14 @@ def _build_event(self, element, key, event_time):
             if key:
                 event.key = key
             elif self._key_field:
-                event.key = event.body[self._key_field]
+                event.key = get_key_from_data(event.body, self._key_field)
```
dinal, line 66: i can't add this comment in the right place
need to change the signature of __init__ func that key_field is a union

dinal (commented):
what about QueryByKey? user is going to pass a list with using online service

dinal:
pls add a test that uses the cache as well, like something with aggregatebykey

PR #184, opened 2021-03-11
Cache

dinal:


storey/table.py
```diff
@@ -235,11 +251,122 @@ def __getitem__(self, key):
          """
         return self._get_static_attrs(key)
 
+    async def _flush_worker(self):
+        while not self._terminated:
+            last_flush = time.monotonic() - (self._flush_interval_milli * 1000)
+            for key, item in self._attrs_cache.items():
+                if item.last_change > last_flush:
+                    await self._persist(_PersistJob(key, None, None))
+            await asyncio.sleep(self._flush_interval_milli / 1000)
```
gtopper: This is probably not exactly the semantic we want. If interval is every 2 seconds, and it took 2 seconds or more to persist, then we don't want to sleep at all. And if it took 1 second to persist, we only want to sleep one more second. So we need to use last_flush as a reference point here. I believe this is how it's done in `_Batching`.

storey/table.py
```diff
@@ -235,11 +251,122 @@ def __getitem__(self, key):
...
+
+    async def _persist(self, job):
+        if not self._q:
+            self._terminated = False
+            self._q = asyncio.queues.Queue(self._max_updates_in_flight)
+            self._worker_awaitable = asyncio.get_running_loop().create_task(self._persist_worker())
+
+        if self._worker_awaitable.done():
+            await self._worker_awaitable
+            raise FlowError("ConcurrentByKeyJobExecution worker has already terminated")
```
gtopper: ConcurrentByKeyJobExecution? Who dis?

storey/table.py
```diff
@@ -235,11 +251,122 @@ def __getitem__(self, key):
...
+                self._pending_by_key[job.key].pending = []
+                task = self._safe_process_events(self._pending_by_key[job.key].in_flight)
+                await self._q.put((job, asyncio.get_running_loop().create_task(task)))
+                if self._worker_awaitable.done():
+                    await self._worker_awaitable
+
+    async def _safe_process_events(self, jobs):
+        try:
+            # TODO only last??
+            job = jobs[-1]
```
gtopper: Yes, because they're supposed to have the same key?

dinal: i kept the same behaviour as it was before (taking only the last event) but that's not entirely correct 
it will work fine for aggregations but not for static attrs.
each event may have different attrs to write and you could miss some if writing only the last one

gtopper: Decided to update the TODO with this information.

storey/table.py
```diff
@@ -235,11 +251,122 @@ def __getitem__(self, key):
...
+            if task and task is not _termination_obj and task[0].extra_data and task[0].extra_data._awaitable_result:
+                task[0].extra_data._awaitable_result._set_error(ex)
+            if not self._q.empty():
+                await self._q.get()
+            raise ex
+
+    async def _terminate(self):
+        if self._q:
+            self._terminated = True
+            last_flush = time.monotonic() - (self._flush_interval_milli * 1000)
```
gtopper: Hm, millis * 1000? Should have been division?

storey/table.py
```diff
@@ -214,8 +224,14 @@ def _get_static_attrs(self, key):
     def _set_static_attrs(self, key, value):
         if key in self._attrs_cache:
             self._attrs_cache[key].static_attrs = value
+            self._attrs_cache[key].last_changed = time.monotonic()
         else:
             self._attrs_cache[key] = _CacheElement(value, None)
+            try:
+                # static attrs can be set before creating a running loop
+                self._init_flush_task()
+            except RuntimeError:
```
gtopper: What's RuntimeError and why are we catching it here?

integration/test_aggregation_integration.py
```diff
@@ -13,8 +13,9 @@
 
 
 @pytest.mark.parametrize('partitioned_by_key', [True, False])
-def test_aggregate_and_query_with_different_windows(setup_teardown_test, partitioned_by_key):
-    table = Table(setup_teardown_test, V3ioDriver(), partitioned_by_key=partitioned_by_key)
+@pytest.mark.parametrize('flush_interval', [0, 1000])
```
gtopper: We need tests for `flush_interval` specifically. i.e. ones that would fail if the feature wasn't working. Something like a flow without WriteToTable followed by another flow reading the persisted data back (without reusing the Table object which would cache the data in memory).

storey/table.py
gtopper: Also, since we divide it everywhere, maybe we can ask for seconds to begin with? (in the constructor parameter)
Otherwise, at least we can convert it to seconds one time when we get it.

storey/writers.py
```diff
@@ -554,12 +554,21 @@ def __init__(self, table: Union[Table, str], columns: Optional[List[str]] = None
...
-        return await self._table._persist_key(events[0].key, data_to_persist)
-
     async def _handle_completed(self, event, response):
         await self._do_downstream(event)
+
+    async def _do(self, event):
+        if event is _termination_obj:
+            await self._table._terminate()
+            return await self._do_downstream(_termination_obj)
+        elif self._table._flush_interval_milli == 0:
```
gtopper: Why is flush interval also used by WriteToTable? I thought it's for the table to persist by itself, not a parameter to WriteToTable.

gtopper: It seems to me that WriteToTable should persist all events it gets, which is in turn controlled by the emit policy (so whether it will persist for each event or not is controlled by that already).

storey/table.py
```diff
@@ -235,11 +261,135 @@ def __getitem__(self, key):
          """
         return self._get_static_attrs(key)
 
+    async def _flush_worker(self):
+        try:
+            while not self._terminated:
+                for key, item in self._attrs_cache.items():
```
gtopper: Discussed as a performance issue

integration/test_flow_integration.py
```diff
@@ -496,3 +496,23 @@ def set_moshe_time_to_none(data):
...
+
+    controller.emit({'col1': 0}, 'dina', test_base_time + timedelta(minutes=25))
+    response = asyncio.run(get_kv_item(setup_teardown_test, 'dina')).output.item
+    assert response == {}
+    time.sleep(4)
+
+    response = asyncio.run(get_kv_item(setup_teardown_test, 'dina')).output.item
+    assert response == {'col1': 0}
+
+    controller.terminate()
```
gtopper, line 516: Missing `controller.await_termination()`

integration/test_flow_integration.py
```diff
@@ -496,3 +496,23 @@ def set_moshe_time_to_none(data):
...
+    controller = build_flow([
+        Source(),
+        WriteToTable(table),
+
+    ]).run()
+
+    controller.emit({'col1': 0}, 'dina', test_base_time + timedelta(minutes=25))
+    response = asyncio.run(get_kv_item(setup_teardown_test, 'dina')).output.item
+    assert response == {}
+    time.sleep(4)
```
gtopper, line 511: This could fail sometimes.

integration/test_flow_integration.py
```diff
@@ -496,3 +496,23 @@ def set_moshe_time_to_none(data):
...
     assert response.status_code == 200
     assert expected == response.output.item
+
+
+def test_cache_flushing(setup_teardown_test):
+    table = Table(setup_teardown_test, V3ioDriver(), flush_interval_secs=3)
+    controller = build_flow([
+        Source(),
+        WriteToTable(table),
+
```
gtopper: Accidental blank line.

storey/table.py
```diff
@@ -17,9 +19,12 @@ class Table:
     :param storage: Storage driver
     :param partitioned_by_key: Whether that data is partitioned by the key or not, based on this indication storage drivers
      can optimize writes. Defaults to True.
+    :param flush_interval_secs: interval in seconds to flush the cache to storage
```
gtopper: Currently disabled when 0 is passed. Needs docu and maybe a clearer way of setting this.

storey/table.py
```diff
@@ -244,6 +275,148 @@ def __getitem__(self, key):
...
+                    received_job_count += 1
+                    if task is _termination_obj:
+                        if received_job_count in self_sent_jobs:
+                            await self._q.put(_termination_obj)
+                            continue
+                        for key, pending_event in self._pending_by_key.items():
+                            if pending_event.pending and not pending_event.in_flight:
+                                for job in pending_event.pending:
+                                    resp = await self._internal_persist_key(key, job.data)
+                                    self._changed_keys.discard(key)
```
gtopper, line 305: Can't it be that the key was changed again while `await self._internal_persist_key` suspended execution?

integration/test_flow_integration.py
dinal: wdym?
time.sleep may fail or the get_item?

storey/table.py
dinal: not really, this flow will execute when a termination_obj is received so no updates will come after it

integration/test_flow_integration.py
gtopper: I mean the test may fail, because sleeping a constant time isn't guaranteed to be enough.

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+                self._terminated = False
+
+        if self._worker_awaitable.done():
+            await self._worker_awaitable
+            raise FlowError("Persist worker has already terminated")
+        else:
+            # Initializing the key with 2 lists. One for pending requests and one for requests that an update request has been issued for.
+            if isinstance(job.key, list):
+                # list can't be key in a dictionary
+                job_key = str(job.key)
```
gtopper, line 378: Should probably be the hashing function.

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+                    await self._worker_awaitable
+
+    async def _safe_process_events(self, jobs):
+        try:
+            # TODO using only last event might not work correctly if
+            #  there are different non aggregation attrs in each event
+            job = jobs[-1]
+            if isinstance(job.key, list):
+                # list can't be key in a dictionary
+                job_key = str(job.key)
```
gtopper, line 401: Should probably be the hashing function.

storey/writers.py
```diff
@@ -652,12 +652,23 @@ def __init__(self, table: Union[Table, str], columns: Optional[List[str]] = None
...
     async def _handle_completed(self, event, response):
         await self._do_downstream(event)
+
+    async def _do(self, event):
+        if event is _termination_obj:
+            await self._table._terminate()
+            return await self._do_downstream(_termination_obj)
+
+        if event.key is None:
+            raise ValueError("Can't write to table without a key")
```
gtopper: Could be clearer, e.g. "Event could not be written to table because it has no key"

storey/table.py
```diff
@@ -1,25 +1,43 @@
...
+
+
 class Table:
     """Table object, represents a single table in a specific storage.
 
     :param table_path: Path to the table in the storage.
     :param storage: Storage driver
     :param partitioned_by_key: Whether that data is partitioned by the key or not, based on this indication storage drivers
      can optimize writes. Defaults to True.
+    :param flush_interval: Policy indication when to flush the cache. default is after 300 secs
```
gtopper: I was going to suggest
`How often the cache will be flushed, in seconds. Default is 300 (5 minutes).`
but I see that it's not an int anymore.

What about disabling it? How is the user supposed to know what to pass in now that the type is `FlushInterval`?

integration/test_aggregation_integration.py
```diff
@@ -3,19 +3,23 @@
 import pytest
 import math
 import pandas as pd
+import v3io
+import v3io.aio.dataplane
```
gtopper: These imports are not used.

integration/test_flow_integration.py
dinal: flush interval is 3 secs, do you think 1 sec for doing 1 IO is not enough? i can change the numbers a bit

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+            self._terminated = True
+            for key in self._changed_keys:
+                await self._persist(_PersistJob(key, None, None), from_terminate=True)
+
+            await self._q.put(_termination_obj)
+            await self._worker_awaitable
+            if self._flush_task:
+                self._flush_task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await self._flush_task
```
gtopper: Why await it at all if we just cancelled it?

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
          """
         return self._get_static_attrs(key)
 
+    async def _flush_worker(self):
+        try:
+            while not self._terminated:
+                await asyncio.sleep(self._flush_interval_secs)
```
gtopper, line 276: This is probably not exactly the semantic we want. If interval is every 2 seconds, and it took 2 seconds or more to persist, then we don't want to sleep at all. And if it took 1 second to persist, we only want to sleep one more second. So we need to use last_flush as a reference point here. I believe this is how it's done in _Batching.

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+                await self._q.get()
+            raise ex
+
+    async def _terminate(self):
+        if self._flush_exception is not None:
+            raise self._flush_exception
+        if not self._terminated:
+            self._terminated = True
+            for key in self._changed_keys:
+                await self._persist(_PersistJob(key, None, None), from_terminate=True)
```
gtopper, line 354: Do we really want to persist them serially?

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
          """
         return self._get_static_attrs(key)
 
+    async def _flush_worker(self):
+        try:
+            while not self._terminated:
+                await asyncio.sleep(self._flush_interval_secs)
+                for key in self._changed_keys:
+                    await self._persist(_PersistJob(key, None, None))
```
gtopper, line 278: Do we really want to persist them serially?

storey/table.py
gtopper: It may also persist twice, here and in the flush worker, because the flush worker is only cancelled after this bit.

storey/writers.py
```diff
@@ -634,12 +634,12 @@ class WriteToTable(_ConcurrentByKeyJobExecution, _Writer):
 
     def __init__(self, table: Union[Table, str], columns: Optional[List[str]] = None, infer_columns_from_data: Optional[bool] = None,
                  **kwargs):
+        Flow.__init__(self, **kwargs)
```
gtopper: Why move the __init__ call up? The whole point of the kwargs stuff is to happen before this call.

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+            return await self._internal_persist_key(job_key, job.data)
+        except BaseException as ex:
+            for job in jobs:
+                if job.extra_data and job.extra_data._awaitable_result:
+                    none_or_coroutine = job.extra_data._awaitable_result._set_error(ex)
+                    if none_or_coroutine:
+                        await none_or_coroutine
+            raise ex
+
+    def update_interval(self, flush_interval: FlushInterval):
```
gtopper: This method is never called.

storey/table.py
```diff
@@ -28,6 +46,15 @@ def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool =
         self._schema = None
         self._aggregations_read_only = False
         self._use_windows_from_schema = False
+        self._q = None
+        self._max_updates_in_flight = max_updates_in_flight
+        self._pending_by_key = {}
+        self._flush_interval_secs = flush_interval.interval_secs
+        self._cache_size = cache_size
```
gtopper: Better add this when you're actually going to use it.

storey/table.py
```diff
@@ -1,25 +1,43 @@
...
+
 class Table:
     """Table object, represents a single table in a specific storage.
 
     :param table_path: Path to the table in the storage.
     :param storage: Storage driver
     :param partitioned_by_key: Whether that data is partitioned by the key or not, based on this indication storage drivers
      can optimize writes. Defaults to True.
+    :param flush_interval: Policy indication when to flush the cache. default is after 300 secs
+    :param max_updates_in_flight: max concurrent number of io requests
```
gtopper: Maximum number of concurrent updates.

storey/table.py
dinal: this is the code in development 
https://github.com/mlrun/storey/blob/122fd04da3301c09f0aad058682a9d19041cf67d/storey/flow.py#L735

storey/table.py
```diff
@@ -64,32 +97,42 @@ async def _get_or_load_static_attributes_by_key(self, key, attributes='*'):
...
+                                      self._partitioned_by_key, additional_data_persist)
+
+    def _set_aggregation_metadata(self, aggregates: List[FieldAggregator], use_windows_from_schema: bool = False):
+        self._use_windows_from_schema = use_windows_from_schema
+        self._aggregates = aggregates
+
+    def _init_flush_task(self):
+        if self._flush_task is None and self._flush_interval_secs > 0:
+            loop = asyncio._get_running_loop()
+            if loop:
```
gtopper: Having to check this probably means thta `_init_flush_task` gets called in too many places. It doesn't really make sense for `_init_flush_task` to be called from outside an async def.

storey/table.py
```diff
@@ -244,6 +288,151 @@ def __getitem__(self, key):
...
+
+            await self._q.put(_termination_obj)
+            await self._worker_awaitable
+            if self._flush_task:
+                self._flush_task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await self._flush_task
+            self._q = None
+            self._flush_task = None
+            self._flush_exception = None
```
gtopper, line 359: What if an exception happened after we checked but before we cancelled it? (relates to the cancellation happening too late).

storey/table.py
dinal: orit asked for the interval to be configurable, i added this method to be able to change flush interval in runtime without stopping the process

storey/table.py
dinal: i moved the cancelation to the start of terminate and i'll check the exception after cancelation

storey/table.py
dinal: i call _init_flush_task from any flow that changes the cache (aggregations or static attrs)
the problem is that we allow setting static attrs on the table before we started running the loop (like in test_write_table_specific_columns) so in some cases _set_static_attrs() may be called before there's a running event loop

storey/table.py
gtopper: Never mind, this pushes the job into the persist_worker's job queue.

storey/table.py
dinal: yea changed it, i first cancel flush worker and then continue with terminate flow

storey/table.py
dinal: i don't think this semantics is better.
if the interval is 2 secs and it took 2 or more secs to flush then we start flushing again? so 100% of our time is dedicated to flushing? it's not right.
also, we flush everything when timeout expires, so what should be the time of "last_flush", when we start flushing the first event? when we finished the last? there's the all in_flight logic, do we consider it in the last_flush time? 
imo it will make stuff confusing and error prone without actual benefit.

storey/writers.py
```diff
@@ -652,12 +652,23 @@ def __init__(self, table: Union[Table, str], columns: Optional[List[str]] = None
...
+        if event is _termination_obj:
+            await self._table._terminate()
+            return await self._do_downstream(_termination_obj)
+
+        if event.key is None:
+            raise ValueError("Can't write to table without a key")
+
+        if self._table._flush_interval_secs == 0:
+            data_to_persist = self._event_to_writer_entry(event)
+            await self._table._persist(_PersistJob(event.key, data_to_persist, self._handle_completed, event))
```
gtopper, line 670: No `do_downstream` in this case?

integration/test_flow_integration.py
gtopper: Ideally the test would be deterministic, e.g. we'd wait for an event to be emitted before checking. I suppose we can leave it as is and change it if it fails.

storey/table.py
gtopper: So that's probably not the right place to call this method. One unexpected bad thing that could happen is there could be *another* event loop running. Best to keep this in the async context.

storey/table.py
gtopper: So bug in dev?

dinal: don't know. i'll wait for katya to take a look and get back to us

storey/writers.py
dinal: do_downstream is called from handle_completed callback

storey/table.py
dinal: why would there be another event loop?
_get_running_loop() only returns the running event loop if there's one or None

storey/table.py
gtopper: If we are meant to flush every minute, and it takes one minute to flush, then yes, I'd expect it to be flushing continuously. If we're supposed to flush every 10 seconds, and it takes one minute to flush, it makes no sense to wait another 10 seconds when flushing already took too long relative to what was intended. Otherwise it's basically an arbitrary sleep between flushes.

storey/table.py
gtopper: I'm not sure I understand in what scenarios this API can be used.

storey/table.py
gtopper: I can definitely write code to do this. It's hard to predict when this will arise in the real world. For example, jupyter sometimes runs its own event loop, such that you would get it if you call get_running_loop().

storey/table.py
```diff
@@ -1,25 +1,43 @@
...
+from .dtypes import FieldAggregator, SlidingWindows, FixedWindows, FlowError, _termination_obj
 
 from .aggregation_utils import is_raw_aggregate, get_virtual_aggregation_func, get_implied_aggregates, get_all_raw_aggregates, \
     get_all_raw_aggregates_with_hidden
-from .drivers import Driver
-from .dtypes import FieldAggregator, SlidingWindows, FixedWindows
 from .utils import _split_path, get_hashed_key
 
 
+class FlushInterval:
```
gtopper: We decided to nuke it and use `option[int]`.

storey/table.py
gtopper: We decided that we can let it be as is for now and maybe reconsider in the future.

storey/table.py
```diff
@@ -17,9 +18,13 @@ class Table:
     :param storage: Storage driver
     :param partitioned_by_key: Whether that data is partitioned by the key or not, based on this indication storage drivers
      can optimize writes. Defaults to True.
+    :param flush_interval_secs: How often the cache will be flushed is seconds. None for flush every event. default is 5 minutes
```
gtopper: is -> in

gtopper: `default is 5 minutes`
->
`Default is 300 (5 minutes).`

storey/table.py
```diff
@@ -17,9 +18,13 @@ class Table:
...
+     default is FlushInterval(300) - flush every 5 minutes
```
gtopper: Hm, this is already mentioned above without the outdated reference to `FlushInterval`.

tests/test_flow.py
```diff
@@ -1938,7 +1940,7 @@ def test_error_in_concurrent_by_key_task():
         WriteToTable(table, columns=['twice_total_activities']),
     ]).run()
 
-    controller.emit(None)
+    controller.emit({'col1': 0}, 'tal')
```
gtopper, line 1943: Aw

dinal: you can't emit an event without a key (it throws an exception now)

gtopper: Tal 🥰

gtopper (approved):

PR #185, opened 2021-03-14
Support timedelta column in kv

urihoenig:


storey/drivers.py
```diff
@@ -357,6 +357,8 @@ def _convert_python_obj_to_expression_value(value):
             secs = int(timestamp)
             nanosecs = int((timestamp - secs) * 1e+9)
             return f'{secs}:{nanosecs}'
+        elif isinstance(value, timedelta):
+            return str(value.value)
```
katyakats, line 361: can you add a test for it?

katyakats (approved):

PR #186, opened 2021-03-18
add support for multiple key fields

katyakats:


storey/sources.py
```diff
@@ -558,10 +558,17 @@ def _blocking_io_loop(self):
...
                                     element[header[i]] = parsed_line[i]
                         if self._key_field:
-                            key_field = self._key_field
-                            if self._with_header and isinstance(key_field, str):
-                                key_field = field_name_to_index[key_field]
-                            key = parsed_line[key_field]
+                            if isinstance(self._key_field, list):
+                                key = []
+                                for single_key_field in self._key_field:
+                                    if isinstance(single_key_field, str):
```
dinal: what about the if self._with_header part?

storey/table.py
```diff
@@ -79,12 +83,14 @@ async def close(self):
...
 
     async def _aggregate(self, key, data, timestamp):
+        key = get_hashed_key(key)
         if not self._schema:
             await self._get_or_save_schema()
 
         self._get_aggregations_attrs(key).aggregate(data, timestamp)
 
     async def _get_features(self, key, timestamp):
+        key = get_hashed_key(key)
```
dinal: it's not needed here, it calls _get_aggregations_attrs that already does the hashing

storey/table.py
```diff
@@ -79,12 +83,14 @@ async def close(self):
         await self._storage.close()
 
     async def _aggregate(self, key, data, timestamp):
+        key = get_hashed_key(key)
```
dinal: it's not needed here, it calls _get_aggregations_attrs that already does the hashing

storey/table.py
```diff
@@ -31,6 +31,7 @@ def __str__(self):
         return f'{self._container}/{self._table_path}'
 
     def _update_static_attrs(self, key, data):
+        key = get_hashed_key(key)
```
dinal: same comment

storey/table.py
```diff
@@ -186,18 +192,21 @@ def _aggregates_to_schema(self):
...
 
     def _set_aggregations_attrs(self, key, element):
+        key = get_hashed_key(key)
         if key in self._attrs_cache:
             self._attrs_cache[key].aggregations = element
         else:
             self._attrs_cache[key] = _CacheElement({}, element)
 
     def _get_static_attrs(self, key):
+        key = get_hashed_key(key)
```
dinal, line 206: i can't put the comment in the right place but need to add this line in _set_static_attrs as well

PR #187, opened 2021-03-18
Aggregations emit NaN values where a value cannot be calculated due to lack of data.

gtopper:


storey/table.py
```diff
@@ -723,11 +724,13 @@ class MinValue(AggregationValue):
     default_value = float('inf')
```
dinal, line 726: it's unused (same for MaxValue)

storey/table.py
```diff
@@ -723,11 +724,13 @@ class MinValue(AggregationValue):
...
 
     def __init__(self, max_value=None, set_data=None):
-        self.value = max_value or self.default_value
+        self.value = math.nan
         super().__init__(max_value, set_data)
 
     def aggregate(self, time, value):
-        if value < self.value:
+        if self.value is math.nan:
+            self.value = value
```
dinal: what about max_value? value may be bigger than max_value
same comment to MaxValue

dinal (approved):

PR #191, opened 2021-03-24
ML-303: In case key does not exist do not return features default values

katyakats:


integration/test_aggregation_integration.py
```diff
@@ -1171,9 +1171,51 @@ def test_aggregate_multiple_keys(setup_teardown_test):
...
     ]
 
     assert actual == expected_results, \
-        f'actual did not match expected. \n actual: {actual} \n expected: {expected_results}'
\ No newline at end of file
+        f'actual did not match expected. \n actual: {actual} \n expected: {expected_results}'
+
+
+def test_read_non_existing_key(setup_teardown_test):
+    current_time = pd.Timestamp.now()
```
gtopper: Better use a constant time for deterministic results, as done in other tests.

gtopper (approved):

PR #192, opened 2021-03-25
ML-338: reading from csv None in boolean column

katyakats:


tests/test-with-none-values.csv
```diff
@@ -0,0 +1,3 @@
+string,bool,bool_with_nan
```
gtopper: bool_with_nan?

gtopper (approved):

PR #193, opened 2021-04-04
ML-345: null in TS column

katyakats:


integration/test_flow_integration.py
```diff
@@ -462,3 +462,27 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
     assert response.status_code == 200
     assert expected == response.output.item
 
+
+def test_write_none_time(setup_teardown_test):
+
+    table = Table(setup_teardown_test, V3ioDriver())
+    current_time = pd.Timestamp.now()
```
gtopper: Better use a constant time.

integration/test_flow_integration.py
```diff
@@ -462,3 +462,27 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
...
+
+    data = pd.DataFrame(
+        {
+            "first_name": ["moshe", "yosi"],
+            "color": ['blue', 'yellow'],
+            "time": [current_time - pd.Timedelta(minutes=25), None,]
```
gtopper: The last comma is a bit strange, even though it's legal syntax.

integration/test_flow_integration.py
```diff
@@ -462,3 +462,27 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
...
+    controller = build_flow([
+        DataframeSource(data, key_field='first_name'),
+        WriteToTable(table),
+    ]).run()
+    controller.await_termination()
+
+    response = asyncio.run(get_kv_item(setup_teardown_test, 'yosi'))
+    expected = {'first_name': 'yosi', 'color': 'yellow'}
+    assert response.status_code == 200
+    assert expected == response.output.item
```
gtopper: Missing newline at EOF.

integration/test_flow_integration.py
```diff
@@ -462,3 +462,27 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
     assert response.status_code == 200
     assert expected == response.output.item
 
+
+def test_write_none_time(setup_teardown_test):
```
gtopper, line 466: Shouldn't we try to read it back too?

katyakats: i am reading it in the end of the test (get kv item)

storey/drivers.py
```diff
@@ -223,7 +224,9 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
             for name, value in additional_data.items():
                 if name.casefold() in self.saved_engine_words.keys():
                     name = f'`{name}`'
-                expressions.append(f'{name}={self._convert_python_obj_to_expression_value(value)}')
+                expression_value = self._convert_python_obj_to_expression_value(value)
+                if expression_value:
```
gtopper, line 228: I don't think this can ever be false.

gtopper, line 228: If the timestamp is null and we don't delete the attribute, we're preserving whatever value already exists in the database, which doesn't seem like correct behaviour.

katyakats: yes, we talked about it yesterday :)

storey/drivers.py
```diff
@@ -352,6 +355,8 @@ def _convert_python_obj_to_expression_value(value):
         elif isinstance(value, bytes):
             return f"blob('{base64.b64encode(value).decode('ascii')}')"
         elif isinstance(value, datetime):
```
gtopper, line 359: FYI, this will catch pandas NaT, but not numpy NaT:
`isinstance(np.datetime64("NaT"), datetime.datetime)` is False

storey/drivers.py
gtopper: And now we won't forget :)

storey/drivers.py
```diff
@@ -223,7 +224,11 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
             for name, value in additional_data.items():
                 if name.casefold() in self.saved_engine_words.keys():
                     name = f'`{name}`'
-                expressions.append(f'{name}={self._convert_python_obj_to_expression_value(value)}')
+                expression_value = self._convert_python_obj_to_expression_value(value)
+                if expression_value:
+                    expressions.append(f'{name}={self._convert_python_obj_to_expression_value(value)}')
+                else:
+                    expressions.append(f'REMOVE {name}')
```
gtopper, line 231: Maybe add a test for this case?

katyakats: this is "moshe". it's time is being deleted in the test

gtopper (approved):

storey/drivers.py
gtopper: Will be addressed separately, as @katyakats says there are more issues with numpy time.

PR #194, opened 2021-04-06
ML-313: in case timestamp is pandas timestamp and is local, local tim…

katyakats:
…ezone needed to be set


tests/test_flow.py
```diff
@@ -14,6 +14,7 @@
     WriteToParquet, QueryByKey, \
     WriteToTSDB, Extend, SendToHttp, HttpRequest, WriteToTable, NoopDriver, Driver, Recover, V3ioDriver, ReadParquet
 
+local_time_zone = datetime.now().astimezone().tzinfo
```
gtopper, line 17: This may change.

PR #196, opened 2021-04-06
ML-313: in case this is pandas timestamp,set timezone manually (other…

katyakats:
…wise timestamp() gives wrong result)

### DIFF
--- a/storey/aggregations.py
+++ b/storey/aggregations.py
@@ -115,6 +115,16 @@ class AggregateByKey(Flow):
 
         raise TypeError('aggregates should be a list of FieldAggregator/dictionaries')
 
+    def _get_timestamp(self, event):
+        event_timestamp = event.time
+        if isinstance(event_timestamp, datetime):
+            if isinstance(event_timestamp, pd.Timestamp) and event_timestamp.tzinfo is None
+                # timestamp for pandas timestamp gives the wrong result in case there is no timezone (ML-313)
+                local_time_zone = datetime.now().astimezone().tzinfo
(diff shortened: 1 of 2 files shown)


storey/aggregations.py
```diff
@@ -172,6 +176,9 @@ async def _sleep_and_emit(self):
     async def _emit_event(self, key, event):
         event_timestamp = event.time
         if isinstance(event_timestamp, datetime):
+            if isinstance(event_timestamp, pd.Timestamp) and event_timestamp.tzinfo is None:
+                local_time_zone = datetime.now().astimezone().tzinfo
+                event_timestamp = event_timestamp.replace(tzinfo=local_time_zone)
```
gtopper: Just need a comment with the ticket number and maybe a short explanation.

storey/aggregations.py
```diff
@@ -133,6 +134,9 @@ async def _do(self, event):
 
             event_timestamp = event.time
             if isinstance(event_timestamp, datetime):
+                if isinstance(event_timestamp, pd.Timestamp) and event_timestamp.tzinfo is None:
+                    local_time_zone = datetime.now().astimezone().tzinfo
+                    event_timestamp = event_timestamp.replace(tzinfo=local_time_zone)
```
gtopper: Just need a comment with the ticket number and maybe a short explanation.

gtopper: And it might make sense to extract a function from these 6 lines of code that are duplicated.

gtopper (approved):

PR #198, opened 2021-04-12
iterate over a copy of changed keys

dinal:


storey/table.py
```diff
@@ -350,8 +344,9 @@ async def _terminate(self):
                 raise self._flush_exception
         if not self._terminated:
             self._terminated = True
-            for key in self._changed_keys:
-                await self._persist(_PersistJob(key, None, None), from_terminate=True)
+            for key in self._changed_keys.copy():
```
gtopper, line 347: I guess you don't need the copy here in terminate, although...
I'm pretty sure termination is (and was before) wrong when the table is used in divergent branches of a flow. But I guess that's beyond the scope of this PR.

gtopper (approved):

PR #202, opened 2021-04-21
ML-338: reading from csv None in date column

katyakats:


tests/test-with-none-values.csv
```diff
@@ -1,3 +1,3 @@
-string,bool,bool_with_none,int_with_nan,float_with_nan
```
gtopper, line 1: You can call it date with missing or none or null if you like, but definitely not nan...

gtopper (approved):

PR #203, opened 2021-04-21
Support hashed partitioning in parquet writer.

gtopper:


tests/test_flow.py
```diff
@@ -1619,7 +1619,8 @@ def test_write_to_parquet_with_metadata(tmpdir):
     columns = ['event_key', 'my_int', 'my_string']
     controller = build_flow([
         Source(),
-        WriteToParquet(out_file, columns=['event_key=$key', 'my_int', 'my_string'])
+        WriteToParquet(out_file, columns=['event_key=$key', 'my_int', 'my_string'],
+                       partition_cols=['$year', '$month', '$day', '$hour'])
     ]).run()
```
katyakats, line 1624: why are you passing partition columns in this test and in others?

gtopper: Because the default is hashing the key into 64 buckets, which messes up the order when you read it back.

katyakats: isn't is a problem for the user as well?

katyakats (approved):

PR #213, opened 2021-04-29
handling emitAfterMaxEvent in case of multiple keys

katyakats:


gtopper (approved):
LGTM

storey/aggregations.py
```diff
@@ -152,18 +152,18 @@ async def _do(self, event):
...
+                self._events_in_batch[safe_key]['event'] = event
                 if self._emit_policy.timeout_secs and self._timeout_task is None:
                     self._timeout_task = asyncio.get_running_loop().create_task(self._sleep_and_emit())
-                if self._events_in_batch[key]['counter'] == self._emit_policy.max_events:
-                    event_from_batch = self._events_in_batch.pop(key, None)
+                if self._events_in_batch[safe_key]['counter'] == self._emit_policy.max_events:
+                    event_from_batch = self._events_in_batch.pop(safe_key, None)
                     if event_from_batch is not None:
-                        await self._emit_event(key, event_from_batch['event'])
+                        await self._emit_event(safe_key, event_from_batch['event'])
```
dinal: you emit here the safe_key but in line 153 it emits the regular key
i think it should emit the regular key cause we try to pass the key as the user defined it to all steps

PR #217, opened 2021-05-10
Fix flush_after_seconds and parse time in emit sources.

gtopper:


storey/sources.py
```diff
@@ -42,10 +43,25 @@ def _set_error(self, ex):
...
+        return obj
+    elif isinstance(obj, float) or isinstance(obj, int):
+        return datetime.fromtimestamp(obj, tz=pytz.utc)
+    elif isinstance(obj, str):
+        if time_format is None:
+            return datetime.fromisoformat(obj)
+        else:
+            return datetime.strptime(obj, time_format)
+    else:
+        raise ValueError(f"Could not parse '{obj}' as a time.")
```
urihoenig: I guess that you wanted to print type(obj) and not obj

tests/test_flow.py
```diff
@@ -352,6 +354,41 @@ def test_write_parquet_read_parquet_partitioned(tmpdir):
     assert read_back_result == expected
 
 
+from unittest.mock import MagicMock
+
+
+async def async_test_write_parquet_flush(tmpdir):
+    out_dir = f'{tmpdir}/test_write_parquet_read_parquet_partitioned/{uuid.uuid4().hex}/'
+    columns = ['my_int', 'my_string']
+    target = ParquetTarget(out_dir, partition_cols='my_int', columns=columns, flush_after_seconds=0)
```
urihoenig: Why not test with another value of flush_after_seconds? (verify it's not flushed at first & then flushed)

tests/test_flow.py
```diff
@@ -2426,3 +2463,67 @@ def test_csv_reader_with_none_values():
...
     assert termination_result[1].key == 'b'
     assert termination_result[1].body == ['b', True, None, math.nan, math.nan, None]
+
+
+def test_bad_time_string_input():
+    controller = build_flow([
+        SyncEmitSource(time_field='time'),
+    ]).run()
+
+    try:
```
urihoenig, line 2473: You can use with pytest.raises

tests/test_flow.py
```diff
@@ -2426,3 +2463,67 @@ def test_csv_reader_with_none_values():
...
+    finally:
+        controller.terminate()
+
+
+def test_bad_time_input():
+    controller = build_flow([
+        SyncEmitSource(time_field='time'),
+    ]).run()
+
+    try:
```
urihoenig, line 2485: with pytest.raises here too

storey/sources.py
gtopper: I'll do both.

urihoenig (approved):

PR #226, opened 2021-05-31
ML-625: csv - discovering type if None is in first row

katyakats:


storey/sources.py
```diff
@@ -534,6 +534,9 @@ def _infer_type(self, value):
         except ValueError:
             pass
 
+        if value == "":
```
gtopper: `if value == '':`

storey/sources.py
```diff
@@ -585,6 +590,10 @@ def _blocking_io_loop(self):
                             if not self._types:
                                 for field in parsed_line:
                                     self._types.append(self._infer_type(field))
+                            else:
+                                for index in range(len(parsed_line)):
+                                    if self._types[index] == 'n':
+                                        self._types[index] = self._infer_type(parsed_line[index])
```
gtopper: Suggested that we should probably keep a set of these indices.

gtopper (approved):

PR #228, opened 2021-06-03
Add a Redis driver

abrookins:
This PR adds a Redis driver to Storey.

In this first early version, we’re targeting the smallest set of features: saving a key and loading it.

Because the current async Redis and Redis Cluster clients have diverged, this PR uses the sync Redis client (redis-py) but runs client commands on a thread in the current event loop.


storey/drivers.py
```diff
@@ -426,3 +430,65 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        redis_key = self.make_key(table_path, key)
+        aggregation_prefixes = (self._aggregation_attribute_prefix,
+                                self._aggregation_time_attribute_prefix)
+        non_aggregation_attrs = [
+            name for name in attributes
+            if not name.startswith(aggregation_prefixes)
+        ]
+
+        if attributes == "*":
+            command = partial(asyncify(self.redis.hgetall), redis_key)
```
abrookins: I need to rework this so that when we use HGETALL, we exclude the aggregation attributes.

abrookins:
This is ready for feedback!

As noted in Slack, the latest version of the async driver for Redis (aioredis) wasn't released when I started this project, so I “asyncified” the sync driver, redis-py. I’ve since released 2.0 of aioredis, so I’ll go through and convert this PR to use it soon.

Aside from that, there are two areas specifically that I could use help/feedback on:

1. The `test_read_non_existing_key()` test is failing. Another test is failing for what appears to be the same reason. More details follow.
2. I don't understand why in the integration tests I copied (located in the file `test_aggregation_integration.py`) some expected values are floats and others are integers. More details follow.

## The `test_read_non_existing_key()` test is failing

I copied the aggregation tests into a new file that tested them all with `RedisDriver`. The last remaining test failures relate to the fact that the returned aggregation from a controller includes a key with a 0 value.

Here's what that looks like in `test_read_non_existing_key()`:

```python
        controller.emit({'last_name': 'levi', 'some_data': 5}, 'non_existing_key')
    
        controller.terminate()
        actual = controller.await_termination()
    
        print(actual[0])
    
>       assert 'number_of_stuff_sum_1h' not in actual[0]
E       AssertionError: assert 'number_of_stuff_sum_1h' not in {'last_name': 'levi', 'number_of_stuff_sum_1h': 0.0, 'some_data': 5}
```

As you can see, the test expects not to get "number_of_stuff_sum_1h", but it's there (`'number_of_stuff_sum_1h': 0.0`).

I've stepped through the aggregation and driver code and can't figure out where this is coming from or what is supposed to stop it from ending up in the result. Can someone help me work through that?

## Floats vs. ints

Something that confused me as I worked on this PR is why some expected values in the integration tests are floats and others are integers. For example, in this test, the aggregation results are all integers: https://github.com/mlrun/storey/blob/development/integration/test_aggregation_integration.py#L1352

Elsewhere, the expected values are floats, e.g.: https://github.com/mlrun/storey/blob/development/integration/test_aggregation_integration.py#L1352

It's not that whole numbers are treated as integers... (as the first example shows). In various places in my Redis version of the same tests, I've had to change the expected values from integers to floats because that's what I'm getting back from the controller. 🤔 

Not knowing how important this was on the Storey side, I went to some lengths to ensure that the driver stores whole numbers as integers instead of floats (seems reasonable anyway, to save space?) and to return them as integers. However, while this does use less space in Redis overall, I still get floats back from the controller in some cases.

How important is this on the Storey side?

PR #230, opened 2021-06-07
ML-613: not writing value for empty key

katyakats:
https://jira.iguazeng.com/browse/ML-589

### DIFF
--- a/storey/sources.py
+++ b/storey/sources.py
@@ -85,6 +85,9 @@ class FlowControllerBase:
             else:
                 key = []
                 for field in self._key_field:
+                    if body[field] is None:
+                        key = [None]
+                        break
                     key.append(body[field])
         if not event_time and self._time_field:
             event_time = _convert_to_datetime(body[self._time_field], self._time_format)
@@ -603,12 +606,15 @@ class CSVSource(_IterableSource):
                                 for single_key_field in self._key_field:
                                     if self._with_header and isinstance(single_key_field, str):
                                         single_key_field = field_name_to_index[single_key_field]
+                                    if parsed_line[single_key_field] is None:
+                                        key = [None]
+                                        break
                                     key.append(parsed_line[single_key_field])
                             else:
                                 key_field = self._key_field
                                 if self._with_header and isinstance(key_field, str):
                                     key_field = field_name_to_index[key_field]
-                                key = parsed_line[key_field]
+                                key = [parsed_line[key_field]]
                         if self._time_field:
                             time_field = self._time_field
                             if self._with_header and isinstance(time_field, str):
@@ -687,9 +693,12 @@ class DataframeSource(_IterableSource):
                     if isinstance(self._key_field, list):
                         key = []
                         for key_field in self._key_field:
+                            if body[key_field] is None:
+                                key = [None]
+                                break
                             key.append(body[key_field])
                     else:
-                        key = body[self._key_field]
+                        key = [body[self._key_field]]
                 time = None
                 if self._time_field:
                     time = body[self._time_field]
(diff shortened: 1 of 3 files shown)


tests/test_flow.py
```diff
@@ -2556,6 +2556,19 @@ def test_custom_string_time_input():
...
+    data = pd.DataFrame({'first_name': ['moshe', None, 'katya'], 'some_data': [1, 2, 3]})
+    data.set_index(keys=['first_name'], inplace=True)
+
+    controller = build_flow([
+        DataframeSource(data, key_field=['first_name']),
+        Reduce([], append_and_return),
+    ]).run()
+    result = controller.await_termination()
+
+    assert len(result) == 2
```
gtopper: I know the cardinality is the point, but still better to make sure we got the right two elements.

storey/sources.py
```diff
@@ -619,20 +620,33 @@ def _blocking_io_loop(self):
...
                         if self._time_field:
                             time_field = self._time_field
                             if self._with_header and isinstance(time_field, str):
                                 time_field = field_name_to_index[time_field]
                             time_as_datetime = parsed_line[time_field]
                         else:
                             time_as_datetime = datetime.now()
-                        self._event_buffer.put(Event(element, key=key, time=time_as_datetime))
+                        event = Event(element, key=key, time=time_as_datetime)
+                        if create_event:
```
gtopper: Better move the work for creating the event into the if, to avoid unnecessary work.

storey/sources.py
```diff
@@ -703,17 +718,28 @@ async def _run_loop(self):
...
                 if self._time_field:
                     time = body[self._time_field]
                 id = None
                 if self._id_field:
                     id = body[self._id_field]
-                event = Event(body, key=key, time=time, id=id)
-                await self._do_downstream(event)
+                if create_event:
+                    event = Event(body, key=key, time=time, id=id)
+                    await self._do_downstream(event)
```
gtopper: Better move the work for creating the event into the if, to avoid unnecessary work.

tests/test_flow.py
```diff
@@ -2556,6 +2556,21 @@ def test_custom_string_time_input():
...
+
+    controller = build_flow([
+        DataframeSource(data, key_field=['first_name']),
+        Reduce([], append_and_return),
+    ]).run()
+    result = controller.await_termination()
+    expected = [{'first_name': 'moshe', 'some_data': 1}, {'first_name': 'katya', 'some_data': 3}]
+
+    assert result == expected, \
+        f'actual did not match expected. \n actual: {result} \n expected: {expected}'
```
gtopper: No need for this - pytest already prints this stuff. I know some tests (aggr...) do this, but none of the tests in this file do.

storey/sources.py
katyakats: it;s in the if :)

gtopper: I meant the
```
                time = None
                if self._time_field:
                    time = body[self._time_field]
                id = None
                if self._id_field:
                    id = body[self._id_field]
```
stuff.

storey/sources.py
```diff
@@ -619,20 +620,33 @@ def _blocking_io_loop(self):
...
+                                    create_event = False
                         if self._time_field:
                             time_field = self._time_field
                             if self._with_header and isinstance(time_field, str):
                                 time_field = field_name_to_index[time_field]
                             time_as_datetime = parsed_line[time_field]
                         else:
                             time_as_datetime = datetime.now()
-                        self._event_buffer.put(Event(element, key=key, time=time_as_datetime))
+                        if create_event:
```
gtopper: Here too, I meant the `time_field` stuff, that's not needed in case we don't create an event.

gtopper (approved):

PR #231, opened 2021-06-07
ML-602: Typed columns for explicitly setting parquet schema.

gtopper:


storey/targets.py
```diff
@@ -68,6 +102,13 @@ def parse_notation(columns, metadata_columns, rename_columns):
                 raise ValueError(f'The following columns are used both for partitioning and indexing, which is not allowed: '
                                  f'{list(cols_both_partition_and_index)}')
 
+    _type_string_to_pyarrow_type = {
+        'str': pyarrow.string(),
+        'int': pyarrow.int64(),
+        'float': pyarrow.float64(),
+        'datetime': pyarrow.timestamp('ns'),
+    }
+
```
katyakats, line 108: i suppose some more types need to be added (like bool).

storey/targets.py
```diff
@@ -330,7 +371,8 @@ class ParquetTarget(_Batching, _Writer):
     :type storage_options: dict
     """
 
-    def __init__(self, path: str, index_cols: Union[str, List[str], None] = None, columns: Union[str, List[str], None] = None,
+    def __init__(self, path: str, index_cols: Union[str, List[str], None] = None,
```
katyakats, line 372: please add documentation (maybe needed in feature store as well)

katyakats (changes requested):

katyakats (approved):

PR #232, opened 2021-06-07
First/last aggregations

urihoenig:
ML-428

### DIFF
--- a/storey/drivers.py
+++ b/storey/drivers.py
@@ -132,6 +132,7 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
         if not update_expression:
             return
 
+        print(f'update_expression={update_expression}')
         key = str(key)
         response = await self._v3io_client.kv.update(container, table_path, key, expression=update_expression,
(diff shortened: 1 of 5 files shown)


storey/drivers.py
```diff
@@ -132,6 +132,7 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
         if not update_expression:
             return
 
+        print(f'update_expression={update_expression}')
```
dinal: remove

storey/table.py
```diff
@@ -977,11 +980,13 @@ def aggregate(self, time, value):
 class LastValue(AggregationValue):
     name = 'last'
 
-    def __init__(self, max_value=None, set_data=None):
-        super().__init__(max_value, set_data)
+    def __init__(self, max_value=None, set_data=None, set_time=None):
```
dinal, line 983: not related to __init__
LastValue doesn't have reset() impl of it's own so it will use the one in AggregationValue and will set set_time to None instead to -math.inf

PR #239, opened 2021-06-16
Flatten/copy bug issue fix

Michaelliv:
An issue was spotted when a customer was using `ModelServerV2.to_mock_server()`, caused by using a custom event class (`MockEvent`) instead of Storey's `Event` class. The custom event class was not properly aligned with Storey's Event class interface and therefore failed on steps that use the Event.copy method.
  
This PR introduces a number of minor changes to address this issue. 
1) Simplified `Flatten` implementation to use `FlatMap` internally (to keep a consistent behavior between the two)
2) Removed the copy method from Event class, since it introduced too much assumptions about the what type of event will be used in the flow.
3) Updated tests to reflect changes.

### DIFF
--- a/storey/steps/flatten.py
+++ b/storey/steps/flatten.py
@@ -1,18 +1,5 @@
-from storey import Flow
-from storey.dtypes import _termination_obj
+from storey import FlatMap
 
 
-class Flatten(Flow):
-    """
-    Splits an event with an iterable body into multiple events.
-    """
-
-    def __init__(self, **kwargs):
-        super().__init__(full_event=True, **kwargs)
-
-    async def _do(self, event):
-        if event is _termination_obj:
-            return await self._do_downstream(_termination_obj)
-        else:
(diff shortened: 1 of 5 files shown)


storey/steps/flatten.py
```diff
@@ -1,18 +1,5 @@
...
-        super().__init__(full_event=True, **kwargs)
-
-    async def _do(self, event):
-        if event is _termination_obj:
-            return await self._do_downstream(_termination_obj)
-        else:
-            for element in event.body:
-                await self._do_downstream(event.copy(body=element))
+def Flatten(**kwargs):
+    return FlatMap(lambda x: x, **kwargs)
```
gtopper, line 8: The only thing is, I think passing `full_event=True` will always break. Although this is probably not so important.

gtopper (approved):
LGTM other than one minor issue that might be worth addressing (or not) – see comment.

PR #241, opened 2021-06-16
storey: fixed window rework

eyalsol:

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -1149,7 +1149,7 @@ class AggregationBuckets:
             return bucket_index
         else:
             self.advance_window_period(timestamp)
-            return self.total_number_of_buckets - 1  # return last index
+            return self.get_bucket_index_by_timestamp(timestamp)
 
     #  Get the index of the bucket corresponding to the requested timestamp
     #  Note: This method can return indexes outside the 'buckets' array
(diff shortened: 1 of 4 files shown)


storey/utils.py
```diff
@@ -42,6 +43,13 @@ def get_one_unit_of_duration(string_time):
     return multiplier
 
 
+def get_window_start_time_by_time(timestamp, window_millis):
+    midnight = datetime.fromtimestamp(int(timestamp/1000)).replace(hour=0, minute=0, second=0)
```
gtopper: Looks to me like this is where the time zone fails CI. The timestamp would be epoch (UTC), but midnight will depend on the local time zone.

Also, what's special about midnight? Why should we use it to determine the window start time, rather than a generic window aligned to epoch?

eyalsol: This make sense. the idea to calculate common base time to start the window but not to move to another day

storey/utils.py
```diff
@@ -42,6 +43,14 @@ def get_one_unit_of_duration(string_time):
     return multiplier
 
 
+def get_window_start_time_by_time(timestamp, window_millis):
+    midnight = datetime.utcfromtimestamp(int(timestamp/1000)).replace(hour=0, minute=0, second=0)
+    window_seconds = int(window_millis / 1000)
+    timestamp_seconds = int(timestamp / 1000)
+    return max(int(timestamp_seconds / window_seconds) * window_seconds,
+               midnight.replace(tzinfo=timezone.utc).timestamp()) * 1000
```
gtopper: Might as well replace the time zone together with the hour/minute/second, no?

eyalsol: Done

storey/utils.py
```diff
@@ -42,6 +43,14 @@ def get_one_unit_of_duration(string_time):
     return multiplier
 
 
+def get_window_start_time_by_time(timestamp, window_millis):
+    midnight = datetime.utcfromtimestamp(int(timestamp/1000)).replace(hour=0, minute=0, second=0, tzinfo=timezone.utc)
```
gtopper: I know I've been harping on this one, but I'd feel better with just simple arithmetic, as is the case in the rest of the aggregation logic.
```
_seconds_per_day = 24 * 60 * 60


def get_window_start_time_by_time(timestamp, window_millis):
    timestamp_seconds = int(timestamp / 1000)
    midnight = int(timestamp_seconds / _seconds_per_day) * _seconds_per_day
    window_seconds = int(window_millis / 1000)
    return max(int(timestamp_seconds / window_seconds) * window_seconds, midnight) * 1000
```

gtopper: Also trying to understand why we hardcode midnight as opposed to just going by window (new requirement?).

storey/utils.py
```diff
@@ -42,6 +42,12 @@ def get_one_unit_of_duration(string_time):
     return multiplier
 
 
+def get_window_start_time_by_time(timestamp, window_millis):
```
gtopper: I think you can just inline this one – doesn't seem like a utility for anything other than aggregations.

eyalsol: @gtopper 
function removed and code is moved to its caller

storey/table.py
```diff
@@ -1162,13 +1162,24 @@ def get_nearest_window_index_by_timestamp(self, timestamp, window_millis):
...
+                        self.get_window_range(self.get_end_bucket(self._last_data_point_timestamp), current_window_millis)
+                    current_window_start = \
+                        self.get_window_range(self.get_end_bucket(timestamp), current_window_millis)
 
-                previous_window_start = max(0, previous_window_start)
-                current_window_start = max(0, current_window_start)
-                previous_window_start = min(len(self.buckets) - 1, previous_window_start)
-                current_window_start = min(len(self.buckets), current_window_start)
+                    previous_window_start = max(0, previous_window_start)
+                    current_window_start = max(0, current_window_start)
```
gtopper, line 1180: Doesn't seem like `max` is needed here.

storey/table.py
```diff
@@ -1162,13 +1162,24 @@ def get_nearest_window_index_by_timestamp(self, timestamp, window_millis):
...
+                    previous_window_start = min(len(self.buckets) - 1, previous_window_start)
+                    current_window_start = min(len(self.buckets), current_window_start)
```
gtopper, line 1182: The min here doesn't seem right... Can this actually happen? Would it be correct?

storey/utils.py
```diff
@@ -42,6 +42,12 @@ def get_one_unit_of_duration(string_time):
     return multiplier
 
 
+# def get_window_start_time_by_time(timestamp, window_millis):
+#    window_seconds = int(window_millis / 1000)
+#    timestamp_seconds = int(timestamp / 1000)
+#    return int(timestamp_seconds / window_seconds) * window_seconds * 1000
```
gtopper: Dead code 😵

storey/table.py
eyalsol: this "old" code on sliding windows too.... the coordinate system can allows negative values.   my next step is not to allow this.
for now it there (for sliding windows)

storey/table.py
eyalsol: same as above

gtopper (approved):

PR #242, opened 2021-06-20
ML-707: bug fix in multiple integer key columns

katyakats:


dinal (approved):

storey/utils.py
```diff
@@ -164,6 +164,7 @@ def update_in(obj, key, value):
 
 
 def hash_list(list_to_hash):
+    list_to_hash = [str(element) for element in list_to_hash]
```
gtopper, line 167: Looks like this is actually the [fastest way](https://waymoot.org/home/python_string/) to do it. ✅

integration/test_aggregation_integration.py
```diff
@@ -1363,3 +1363,46 @@ def test_non_existing_key_query_by_key_from_v3io_key_is_list(setup_teardown_test
     controller.emit({'nameeeee': 'katya'}, 'katya')
     controller.terminate()
     controller.await_termination()
+
+
+def test_multiple_keys_int(setup_teardown_test):
```
gtopper, line 1368: Maybe mention in a comment that this is regression for ML-707.

gtopper: I don't think we need an integration test to test this, a unit test would suffice.

gtopper (approved):
Great (see nonblocking comments).

PR #249, opened 2021-06-22
ML-717: Use logger.debug instead of print.

gtopper:


storey/flow.py
```diff
@@ -183,14 +183,19 @@ async def _do_downstream(self, event):
                 event_copy._awaitable_result = awaitable_result
                 tasks.append(asyncio.get_running_loop().create_task(self._outlets[i]._do_and_recover(event_copy)))
             event._awaitable_result = awaitable_result
-        if self.verbose:
-            step_name = type(self).__name__
-            event_string = self._event_string(event)
-            print(f'{step_name} -> {type(self._outlets[0]).__name__} | {event_string}')
+        if self.verbose and self.context:
```
dinal: why do you need to check "if self.verbose and self.context" twice?
save it once and check "if logger"
also in line 197 non need to check again " if verbose"

dinal (approved):

PR #252, opened 2021-06-27
ML 719: infer date type only for time_field column and "parse_dates" columns

katyakats:


storey/sources.py
```diff
@@ -502,6 +504,12 @@ def __init__(self, paths: Union[List[str], str], header: bool = False, build_dic
         self._timestamp_format = timestamp_format
         self._type_inference = type_inference
         self._storage_options = kwargs.get('storage_options')
+        self._parse_dates = parse_dates
+        self._dates_indexes = []
```
urihoenig: date_indices

storey/sources.py
```diff
@@ -502,6 +504,12 @@ def __init__(self, paths: Union[List[str], str], header: bool = False, build_dic
...
+        if isinstance(self._parse_dates, List) and isinstance(self._parse_dates[0], int):
```
urihoenig: Why not verify all?
isinstance(self._parse_dates, List) and all([isinstance(f, int) for f in self._parse_dates])

storey/sources.py
```diff
@@ -588,16 +590,21 @@ def _blocking_io_loop(self):
                         field_name_to_index = {}
                         for i in range(len(header)):
                             field_name_to_index[header[i]] = i
+                            if header[i] == self._time_field or (self._parse_dates and header[i] in self._parse_dates):
```
urihoenig, line 598: + Isn't this relevant only if _dates_indexes is empty?
+ Shouldn't it be cleared after every path handling? The index of columns can be different between files, no?

tests/test_flow.py
```diff
@@ -2622,3 +2626,17 @@ def test_csv_none_value_string(tmpdir):
     r2 = pd.read_parquet(out_file_par)
 
     assert r2['str'].compare(read_back_df['str']).empty
+
+
+def test_csv_multiple_time_columns(tmpdir):
+    controller = build_flow([
+        CSVSource('tests/test-multiple-time-columns.csv', header=True, time_field='t1', parse_dates=["t2"]),
```
urihoenig, line 2643: I'd test multiple files with different column orders

storey/sources.py
katyakats: i assume that it;s either names or integer. not a mixture of both

storey/sources.py
```diff
@@ -655,6 +655,7 @@ def _blocking_io_loop(self):
                                 self.context.logger.error(
                                     f"For {parsed_line} value of key {key_field} is None"
                                 )
+                self._dates_indexes = []
```
urihoenig: This will override the dates_indexes from the __init__.
You don't want to do it if it wasn't initially empty.

storey/sources.py
```diff
@@ -502,6 +504,16 @@ def __init__(self, paths: Union[List[str], str], header: bool = False, build_dic
         self._timestamp_format = timestamp_format
         self._type_inference = type_inference
         self._storage_options = kwargs.get('storage_options')
+        self._parse_dates = parse_dates
+        self._dates_indices = []
+        if isinstance(parse_dates, List):
+            if self._with_header and not all([isinstance(f, str) for f in self._parse_dates]):
+                raise ValueError('parse_dates can be list of int only when there is header')
+                self._dates_indices = parse_dates
```
urihoenig, line 513: indentation (or not)

storey/sources.py
```diff
@@ -502,6 +504,16 @@ def __init__(self, paths: Union[List[str], str], header: bool = False, build_dic
...
+        if isinstance(self._time_field, int):
+            if self._with_header:
+                raise ValueError('time field can be list of int only when there is header')
```
urihoenig: it's an int, not a list of int

storey/sources.py
urihoenig: I guess that there should've been a test that would've failed here

urihoenig (approved):

PR #253, opened 2021-06-28
ML-638: Convert str/int to datetime based on writer schema or source time_field.

gtopper:


storey/dtypes.py
```diff
@@ -30,7 +30,12 @@ def __init__(self, body: object, key: Optional[Union[str, List[str]]] = None, ti
         self.body = body
         self.key = key
         if time is not None and not isinstance(time, datetime):
-            raise TypeError(f'Event time parameter must be a datetime. Got {type(time)} instead.')
+            if isinstance(time, str):
+                time = datetime.fromisoformat(time)
+            elif isinstance(time, int):
+                time = datetime.utcfromtimestamp(time)
+            else:
+                raise TypeError(f'Event time parameter must be a datetime. Got {type(time)} instead.')
```
dinal: Event time parameter must be a datetime, string or int

storey/targets.py
```diff
@@ -124,14 +127,18 @@ def _init_partition_col_indices(self):
...
             self._non_partition_columns = []
+            self._non_partition_column_types = []
             for index, col in enumerate(self._columns):
                 if col in self._partition_cols:
                     self._partition_col_to_index[col] = index
                     self._partition_col_indices.append(index)
                 else:
                     self._non_partition_columns.append(col)
+                    if self._column_types:
+                        self._non_partition_column_types.append(self._column_types[index])
```
dinal, line 141: do you want to append only if the type doesn't exists in the list already?

gtopper: Hm, no? That wouldn't make sense :)

PR #265, opened 2021-07-13
[Storey] Adding aggregation test for `first` and `last` aggregates on fixed wi…

eyalsol:
…ndow

Fixes https://jira.iguazeng.com/browse/ML-428 

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -600,7 +600,8 @@ class ReadOnlyAggregationBuckets:
                 for bucket_id in range(previous_window_start, current_window_start):
                     current_pre_aggregated_value = aggr.value
                     bucket_aggregated_value = self.buckets[bucket_id].value
-                    if self.aggregation == "min" or self.aggregation == "max":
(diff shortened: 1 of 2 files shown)


storey/table.py
```diff
@@ -1022,7 +1023,8 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
 
     def aggregate(self, time, value):
         if time is not None and not math.isnan(value) and (self.time is None or time < self.time):
-            self._set_value(value)
+            if math.isnan(self.value) and math.isnan(self.default_value):
```
dinal, line 1026: why does it matter what's the default value?

dinal: actually, wouldn't it better to check 
if self.value == self.default_value: ...
instead of assuming nan means it's not set

eyalsol: NaN, by IEEE definition: comparing NaN == NaN will always return false:
https://en.wikipedia.org/wiki/NaN  look for "Comparison with NaN"  

The above is right way to compare

PR #266, opened 2021-07-18
ML-686: v3io_driver support for parallel array operations

eyalsol:

### DIFF
--- a/storey/drivers.py
+++ b/storey/drivers.py
@@ -3,10 +3,12 @@ import json
 import os
 from datetime import datetime, timedelta
 from typing import Optional
+from collections import OrderedDict
 import pandas as pd
 
 import v3io
 import v3io.aio.dataplane
+from v3io.dataplane import kv_array
 
 from .dtypes import V3ioError
 from .utils import schema_file_name
@@ -70,7 +72,7 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
     :param access_key: V3IO access key. If not set, the V3IO_ACCESS_KEY environment variable will be used.
     """
 
-    def __init__(self, webapi: Optional[str] = None, access_key: Optional[str] = None):
+    def __init__(self, webapi: Optional[str] = None, access_key: Optional[str] = None, use_parallel_operations=True):
         NeedsV3ioAccess.__init__(self, webapi, access_key)
         self._v3io_client = None
         self._closed = True
@@ -80,6 +82,7 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
         self._error_code_string = "ErrorCode"
         self._false_condition_error_code = "16777244"
         self._mtime_header_name = 'X-v3io-transaction-verifier'
+        self._parallel_ops = use_parallel_operations
 
     def _lazy_init(self):
         self._closed = False
@@ -97,6 +100,7 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
                           'find_all_indices', 'extract_by_indices', 'sort_array', 'blob', 'if_else', 'in', 'true', 'false', 'and', 'or',
                           'not', 'contains', 'ends', 'starts', 'exists', 'select', 'str', 'to_int', 'if_not_exists', 'regexp_instr',
                           'delete', 'as', 'maxdbl', 'inf', 'nan', 'isnan', 'nanaszero'}
+    parallel_aggregates = {'min': 'pmin', 'max': 'pmax', 'sum': 'psum', 'count': 'psum', 'sqr': 'psum'}
 
     async def _save_schema(self, container, table_path, schema):
         self._lazy_init()
@@ -126,6 +130,19 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
 
     async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data=None):
         self._lazy_init()
+
+        # test whether server support parallel operations
+        if self._parallel_ops:
+            try:
+                test_expression = "a=init_array(2,'double',0.0);a[0..1]=pmax(a[0..1], a[0..1]);delete(a);"
+                response = await self._v3io_client.kv.update(container, table_path, str(key),
+                                                             expression=test_expression, condition="",
+                                                             raise_for_status=v3io.aio.dataplane.RaiseForStatus.never)
+                if response.status_code != 200:
+                    self._parallel_ops = False
+            except Exception:
+                pass
+
         should_raise_error = False
         update_expression, condition_expression, pending_updates = self._build_feature_store_update_expression(aggr_item, additional_data,
                                                                                                                partitioned_by_key)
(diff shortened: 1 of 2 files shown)


storey/drivers.py
```diff
@@ -126,6 +130,19 @@ async def _load_schema(self, container, table_path):
...
+
+        # test whether server support parallel operations
+        if self._parallel_ops:
+            try:
+                test_expression = "a=init_array(2,'double',0.0);a[0..1]=pmax(a[0..1], a[0..1]);delete(a);"
+                response = await self._v3io_client.kv.update(container, table_path, str(key),
+                                                             expression=test_expression, condition="",
+                                                             raise_for_status=v3io.aio.dataplane.RaiseForStatus.never)
+                if response.status_code != 200:
+                    self._parallel_ops = False
```
gtopper, line 142: Might wanna log a warning there.

storey/drivers.py
```diff
@@ -229,13 +246,15 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
         return update_expression, condition_expression, pending_updates
 
     def _discard_old_pending_items(self, pending, max_window_millis):
-        res = {}
+        res = OrderedDict()
         if pending:
-            last_time = int(list(pending.keys())[-1] / max_window_millis) * max_window_millis
+            pending_keys = list(pending.keys())
```
gtopper: `.keys()` is redundant.

storey/drivers.py
```diff
@@ -293,20 +312,30 @@ def _build_conditioned_feature_store_request(self, aggregation_element, pending=
...
+        feature_attrs = ["a", "b"]
 
         times_update_expressions = {}
         new_cached_times = {}
         pending_updates = {}
         initialized_attributes = {}
+
+        # build a dictionary to save operations that can be paralleled
+        if self._parallel_ops:
+            pexpressions = {key: {val: None for val in feature_attrs} for key in self.parallel_aggregates.keys()}
```
gtopper: `.keys()` is redundant.

storey/drivers.py
```diff
@@ -322,17 +351,53 @@ def _build_simplified_feature_store_request(self, aggregation_element):
...
                             if array_time_attribute_name not in times_update_expressions:
                                 times_update_expressions[array_time_attribute_name] = \
                                     f'{array_time_attribute_name}={expected_time_expr}'
                             new_cached_times[name] = (array_time_attribute_name, expected_time)
 
                         # Updating the specific cells
                         if cached_time <= expected_time:
-                            arr_at_index = f'{array_attribute_name}[{index_to_update}]'
-                            expressions.append(f'{arr_at_index}={aggregation_value.get_update_expression(arr_at_index)}')
+                            if use_parallel and aggregation in self.parallel_aggregates.keys():
```
gtopper: `.keys()` is redundant.

storey/drivers.py
```diff
@@ -322,17 +351,53 @@ def _build_simplified_feature_store_request(self, aggregation_element):
...
+                                        aggregation_value.value)
+                                    pexpressions[aggregation][feature_attr]["last_index"] = index_to_update
+                            else:
+                                arr_at_index = f'{array_attribute_name}[{index_to_update}]'
+                                expressions.append(
+                                    f'{arr_at_index}={aggregation_value.get_update_expression(arr_at_index)}'
+                                )
+
+        if self._parallel_ops:
+            for aggregation in self.parallel_aggregates.keys():
```
gtopper: `.keys()` is redundant.

storey/drivers.py
```diff
@@ -293,20 +312,30 @@ def _build_conditioned_feature_store_request(self, aggregation_element, pending=
...
     def _build_simplified_feature_store_request(self, aggregation_element):
         expressions = []
+        feature_attrs = ["a", "b"]
 
         times_update_expressions = {}
         new_cached_times = {}
         pending_updates = {}
         initialized_attributes = {}
+
+        # build a dictionary to save operations that can be paralleled
```
gtopper: paralleled -> parallelized

storey/drivers.py
```diff
@@ -70,7 +72,7 @@ class V3ioDriver(NeedsV3ioAccess, Driver):
     :param access_key: V3IO access key. If not set, the V3IO_ACCESS_KEY environment variable will be used.
     """
 
-    def __init__(self, webapi: Optional[str] = None, access_key: Optional[str] = None):
+    def __init__(self, webapi: Optional[str] = None, access_key: Optional[str] = None, use_parallel_operations=True):
```
gtopper, line 75: I guess we need an intergration test that actually uses this?

storey/drivers.py
gtopper: `key` is a bit misleading here. Better call it `parallel_aggregate` or `aggregate` or `aggr`.

storey/drivers.py
```diff
@@ -322,17 +351,53 @@ def _build_simplified_feature_store_request(self, aggregation_element):
...
+        if self._parallel_ops:
+            for aggregation in self.parallel_aggregates.keys():
+                for feature_attr in feature_attrs:
+                    if pexpressions[aggregation][feature_attr] is not None:
+                        d = pexpressions[aggregation][feature_attr]
+                        encoded_array = kv_array.encode_list(d["values"]).decode()
+                        paggregate = self.parallel_aggregates[aggregation]
+                        sliced_array = f'{d["attr_name"]}[{d["first_index"]}..{d["last_index"]}]'
+                        expressions.append(
+                            f"{sliced_array}={paggregate}({sliced_array}, blob('{encoded_array}'))"
```
gtopper: Is `blob` correct here? I thought this would be an array operation.

storey/drivers.py
```diff
@@ -322,17 +351,53 @@ def _build_simplified_feature_store_request(self, aggregation_element):
...
+                                    pexpressions[aggregation][feature_attr] = \
+                                        {"attr_name": array_attribute_name,
+                                         "first_index": index_to_update,
+                                         "last_index": index_to_update,
+                                         "values": [aggregation_value.value],
+                                         "default_value": aggregation_value.default_value
+                                         }
+                                else:
+                                    last_index = pexpressions[aggregation][feature_attr]["last_index"]
+                                    if last_index + 1 < index_to_update:
```
gtopper: I think this `if` is redundant.

gtopper:
If the point of the change is to improve performance, we should probably benchmark it to make sure that it does so at least in some common cases. There is a lot of overhead that could potentially slow things down, so we should probably check this.

storey/drivers.py
eyalsol: `test_query_virtual_aggregations_flow` tests the `V3ioDriver` `use_parallel_operations` parameter

storey/drivers.py
eyalsol: this is from requirement and so does the implementation accordingly

storey/drivers.py
gtopper: We use integration tests for that.

storey/drivers.py
dinal: +1
log what type of operation is going to be used + error code in debug level

PR #269, opened 2021-07-21
ML-846: Fix dataframe multi-column index assignment for pandas 1.3.

katyakats:


storey/dataframe.py
```diff
@@ -44,7 +44,7 @@ async def _do(self, event):
         if event is _termination_obj:
             df = pd.DataFrame(self._data, columns=self._columns)
             if self._insert_key_column_as:
-                df[self._insert_key_column_as] = self._key_column
+                df[self._insert_key_column_as] = pd.DataFrame(self._key_column)
```
gtopper, line 47: Wouldn't it be a `pd.Series`?

katyakats: nope... pd.Series results in error (Columns must be same length as key).

gtopper (approved):

PR #274, opened 2021-07-26
ML-885:   in case file is directly in bucket, s3f3 tries to create the dir and can fail on permissions

katyakats:


dinal (commented):
change to a meaningful title

dinal (approved):

PR #281, opened 2021-08-18
ML-971: Add inner_join parameter to JoinWithTable.

gtopper:


integration/test_flow_integration.py
```diff
@@ -256,19 +254,56 @@ def test_join_by_key_specific_attributes(setup_kv_teardown_test):
 
     controller = build_flow([
         SyncEmitSource(),
-        Filter(lambda x: x['col1'] > 8),
-        JoinWithTable(table, lambda x: x['col1'], attributes=['age']),
```
katyakats, line 260: why did you delete the filter?

dinal (approved):

integration/test_flow_integration.py
gtopper: It seemed irrelevant to the test.

PR #284, opened 2021-08-26
Support for sample per event's key on `SampleWindow`

eyalsol:

### DIFF
--- a/storey/steps/sample.py
+++ b/storey/steps/sample.py
@@ -1,7 +1,8 @@
 from enum import Enum
 
 from storey import Flow
-from storey.dtypes import _termination_obj
+from storey.dtypes import Event, _termination_obj
+from typing import Optional, Union, Callable
 
 
 class EmitPeriod(Enum):
@@ -22,6 +23,11 @@ class SampleWindow(Flow):
     Available options:
         2.1) True - The last event seen will be emitted downstream.
         2.2) False - The last event seen will NOT be emitted downstream.
+    :param key: The key by which events are sampled. By default (None), events are not sampled by key.
+        Other options may be:
+        Set to '$key' to sample events by the Event.key property.
+        set to 'str' key to sample events by Event.body[str].
+        set a Callable[[Event], str] to sample events by a custom key extractor.
     """
 
     def __init__(
@@ -29,6 +35,7 @@ class SampleWindow(Flow):
         window_size: int,
         emit_period: EmitPeriod = EmitPeriod.FIRST,
         emit_before_termination: bool = False,
+        key: Optional[Union[str, Callable[[Event], str]]] = None,
         **kwargs,
     ):
         super().__init__(full_event=True, **kwargs)
@@ -42,9 +49,24 @@ class SampleWindow(Flow):
         self._window_size = window_size
         self._emit_period = emit_period
         self._emit_before_termination = emit_before_termination
-
+        self._per_key_count = dict()
         self._count = 0
         self._last_event = None
+        self._extract_key: Callable[[Event], str] = self._create_key_extractor(key)
+
+    @staticmethod
+    def _create_key_extractor(key) -> Callable:
+        if key is None:
+            return lambda event: None
+        elif callable(key):
+            return key
+        elif isinstance(key, str):
+            if key == '$key':
+                return lambda event: event.key
+            else:
+                return lambda event: event.body[key]
+        else:
+            raise ValueError(f'Unsupported key type {type(key)}')
 
     async def _do(self, event):
         if event is _termination_obj:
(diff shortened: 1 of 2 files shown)


storey/steps/sample.py
```diff
@@ -22,13 +23,22 @@ class SampleWindow(Flow):
     Available options:
         2.1) True - The last event seen will be emitted downstream.
         2.2) False - The last event seen will NOT be emitted downstream.
+    :param count_per_key: Whether to sample up to window size on per key basis. False samples on all events.
```
gtopper: `count_per_key` is redundant. It's always used as `key and count_per_key`. Can just use `key`.

storey/steps/sample.py
```diff
@@ -42,31 +52,61 @@ def __init__(
         self._window_size = window_size
         self._emit_period = emit_period
         self._emit_before_termination = emit_before_termination
-
+        self._count_per_key = count_per_key
+        self._key_selector = key_selector
+        self._per_key_count = dict()
         self._count = 0
         self._last_event = None
+        self._extract_key: Optional[Callable[[Event], str]] = self._create_key_extractor(key)
```
gtopper: Either have `self._create_key_extractor(None)` return `None` or remove `Optional`.

storey/steps/sample.py
```diff
@@ -22,13 +23,22 @@ class SampleWindow(Flow):
     Available options:
         2.1) True - The last event seen will be emitted downstream.
         2.2) False - The last event seen will NOT be emitted downstream.
+    :param count_per_key: Whether to sample up to window size on per key basis. False samples on all events.
+    :param key: The key by which events are sampled. By default (None), events are not sampled by key.
+        Other options may be:
+        Set a '$key' to sample events by the Event.key property.
+        set a 'str' key to sample events by Event.body[str].
+        set a Callable[Any, Any] to sample events by a custom key extractor.
```
gtopper: This line says `Callable[Any, Any]`, the type annotations say `Callable[[Event], str]`.

storey/steps/sample.py
```diff
@@ -22,13 +23,22 @@ class SampleWindow(Flow):
     Available options:
         2.1) True - The last event seen will be emitted downstream.
         2.2) False - The last event seen will NOT be emitted downstream.
+    :param count_per_key: Whether to sample up to window size on per key basis. False samples on all events.
+    :param key: The key by which events are sampled. By default (None), events are not sampled by key.
+        Other options may be:
+        Set a '$key' to sample events by the Event.key property.
```
gtopper: `Set a '$key'` -> `Set to '$key'`

storey/steps/sample.py
```diff
@@ -22,13 +23,22 @@ class SampleWindow(Flow):
...
+        set a 'str' key to sample events by Event.body[str].
```
gtopper: `set a 'str' key` ->
`set an str key` or `set a string key`

storey/steps/sample.py
```diff
@@ -42,31 +49,59 @@ def __init__(
...
     async def _do(self, event):
         if event is _termination_obj:
             if self._last_event is not None:
                 await self._do_downstream(self._last_event)
             return await self._do_downstream(_termination_obj)
         else:
-            self._count += 1
+            key = self._extract_key(event)
+
+            if key is not None:
```
gtopper, line 79: I'm pretty sure you could remove these checks and `self._count` and everything would still work. Something to consider.

storey/steps/sample.py
```diff
@@ -42,31 +49,59 @@ def __init__(
...
+                self._count += 1
+                count = self._count
 
             if self._emit_before_termination:
                 self._last_event = event
 
-            if self._should_emit():
+            if self._should_emit(count):
                 self._last_event = None
                 await self._do_downstream(event)
```
gtopper: Not new in this PR, but I'm worried that this could throw and error and we won't reset the count, causing it to increment infinitely going forward, and never emitting again.

gtopper: This wasn't as much as a problem in the past, but now the default behaviour in EmitSource is to supress errors, and there is the concept of recovery steps (which may be upstream of the sampler).

tests/test_steps.py
```diff
@@ -344,6 +344,23 @@ def test_sample_emit_last_with_emit_before_termination():
     controller.await_termination()
 
 
+def test_sample_emit_event_per_key():
+    controller = build_flow(
+        [
+            SyncEmitSource(key_field=str),
+            Assert().exactly(25),
+            SampleWindow(5, key='$key'),
+            Assert().exactly(5).match_exactly([0, 1, 2, 3, 4]),
```
gtopper, line 353: Don't use `Assert`. It won't fail if nothing is ever emitted by `SampleWindow`. Use `Reduce` and `assert` instead.

gtopper: Never mind, I see this is used all over in this test...

storey/steps/sample.py
```diff
@@ -42,31 +49,64 @@ def __init__(
...
+                self._count += 1
+                count = self._count
 
             if self._emit_before_termination:
                 self._last_event = event
 
-            if self._should_emit():
+            if self._should_emit(count):
                 self._last_event = None
+                # reset counters before downstreaming
```
gtopper: That's not right. It could be that you're emitting on first event in window.

storey/steps/sample.py
```diff
@@ -42,31 +49,59 @@ def __init__(
...
                 self._last_event = None
                 await self._do_downstream(event)
 
-            if self._count == self._window_size:
-                self._count = 0
+            if count == self._window_size:
+                if key is not None:
+                    self._per_key_count[key] = 0
+                else:
+                    self._count = 0
```
gtopper: Not really a regression in this PR, but I think it needs to be as follows to avoid really bad behaviour in case of an error downstream.
```
should_emit = self._should_emit(count)

if count == self._window_size:
  if key is not None:
    self._per_key_count[key] = 0
  else:
    self._count = 0

if should_emit:
  self._last_event = None
```

gtopper (approved):

PR #285, opened 2021-08-29
[ML-1019] - Persist by key before advancing the window

eyalsol:

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -107,10 +107,8 @@ class Table:
                     self._set_static_attrs(key, {})
             return self._get_static_attrs(key)
 
-    async def _internal_persist_key(self, key, event_data_to_persist):
+    async def _internal_persist_key(self, key, event_data_to_persist, aggr_by_key=None, additional_data_persist=None):
         async with self._get_lock(key):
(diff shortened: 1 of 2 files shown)


gtopper (commented):
The code looks correct for what it's meant to do.

Semantically though... This change forces early persistence when we advance the window, regardless of caching or `NoSqlTarget`. I'm pretty sure this is going to have unexpected consequences. One I can think of: a table with `flush_interval_secs=None` used in a flow with no `NoSqlTarget`. Normally, this would persist nothing. Now it will persist whenever the window is advanced, and only then.

storey/table.py
```diff
@@ -167,9 +170,13 @@ async def _add_aggregation_by_key(self, key, base_timestamp, initial_data):
         if self._aggregations_read_only and initial_data is None:
             self._set_aggregations_attrs(key, None)
         else:
+            persist_func = None
+            if self._flush_interval_secs is not None and self._flush_interval_secs > 0:
```
gtopper: If `self._flush_interval_secs is None` we don't need this early persistence? How come?

PR #289, opened 2021-09-05
ML-1062: Sliding window, incorrect results from "first" agg func

eyalsol:


storey/table.py
```diff
@@ -1176,6 +1176,7 @@ def __init__(self, name, explicit_raw_aggregations, hidden_raw_aggregations, vir
         self.explicit_windows = explicit_windows
         self.max_value = max_value
         self.buckets = []
+        self.cache = None
```
dinal: call it something specific, like first_aggr_sliding_window_cache

PR #293, opened 2021-09-19
Refactor table's `persist` to persist asynchronously before advancing…

eyalsol:
… fixed windows

* ML-1019: Persist by key before advancing the window
* ML-1055: Refactor Query (service.get). fixed windows should return last closed window

### DIFF
--- a/storey/table.py
+++ b/storey/table.py
@@ -107,10 +107,8 @@ class Table:
                     self._set_static_attrs(key, {})
             return self._get_static_attrs(key)
 
-    async def _internal_persist_key(self, key, event_data_to_persist):
+    async def _internal_persist_key(self, key, event_data_to_persist, aggr_by_key=None, additional_data_persist=None):
         async with self._get_lock(key):
-            aggr_by_key = self._get_aggregations_attrs(key)
-            additional_data_persist = self._get_static_attrs(key)
             if event_data_to_persist:
                 if not additional_data_persist:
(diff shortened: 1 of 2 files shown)


storey/table.py
```diff
@@ -1540,6 +1585,12 @@ def __init__(self):
 class _PersistJob:
     def __init__(self, key, data, callback, extra_data=None):
         self.key = key
-        self.data = data
+        self.data = copy.deepcopy(data)
```
gtopper, line 1588: Looks like this is used only from NoSqlTarget, when periodic flushing is disabled. So why do we need to deepcopy it?

storey/table.py
```diff
@@ -254,6 +253,15 @@ def _aggregates_to_schema(self):
 
         return schema
 
+    def _flush_pending(self, key):
+        aggregations_attrs = self._get_aggregations_attrs(key)
+        if not aggregations_attrs:
+            return
+        for _, bucket in aggregations_attrs.aggregation_buckets.items():
```
gtopper: `items()` -> `values()`

gtopper (commented):
Any performance measurements? This could have a significant impact, especially when the window is advanced often relative to the sample time interval.

storey/table.py
eyalsol: I don't want a mixture of copied/not copied members. so I deep copy all

dinal (approved):

PR #297, opened 2021-10-13
ML-1223: Create stream automatically in StreamTarget if doesn't exist

dinal:


storey/targets.py
```diff
@@ -589,13 +589,16 @@ class StreamTarget(Flow, _Writer):
         inferred from data and used in place of explicit columns list if none was provided, or appended to the provided list. If header is
         True and columns is not provided, infer_columns_from_data=True is implied. Optional. Default to False if columns is provided,
         True otherwise.
+    :param shard_count: if stream doesn't exist and it will be created set number of shards in the stream. Default 1
+    :param retention_period_hours: if stream doesn't exist and it will be created set retention time. Default 24h
```
```
:param shard_count: If stream doesn't exist, it will be created with this number of shards. Defaults to 1.
:param retention_period_hours: If stream doesn't exist, it will be created with this retention time in hours. Defaults to 24.
```

storey/targets.py
```diff
@@ -686,10 +691,10 @@ async def _worker(self):
             await self._storage.close()
 
     async def _lazy_init(self):
-        if not self._shard_count:
-            response = await self._storage._describe(self._container, self._stream_path)
 
-            self._shard_count = response.shard_count
+        await self._storage._create_stream(self._container, self._stream_path, self._shard_count, self._retention_period_hours)
```
gtopper: We don't want to call this every time we get an event.

storey/targets.py
```diff
@@ -686,10 +691,10 @@ async def _worker(self):
             await self._storage.close()
 
     async def _lazy_init(self):
-        if not self._shard_count:
-            response = await self._storage._describe(self._container, self._stream_path)
```
gtopper, line 690: We still need to know the number of shards in a pre-existing stream.

storey/targets.py
gtopper: Also, won't this fail if the stream already exists?

integration/integration_test_utils.py
```diff
@@ -117,16 +116,6 @@ def setup_stream_teardown_test():
     asyncio.run(recursive_delete(stream_path, V3ioHeaders()))
 
 
-async def setup_stream(stream_path):
```
gtopper, line 120: We still want to test the case where the stream already exists.

storey/targets.py
dinal: it ignores the error if stream already exists 
 
`res.raise_for_status([409, 204])`

gtopper: Right, `raise_for_status` method name is misleading 🙈

storey/targets.py
```diff
@@ -589,13 +589,16 @@ class StreamTarget(Flow, _Writer):
...
+    :param retention_period_hours: If stream doesn't exist, it will be created with this retention time in hours. Defaults to 24.
     :param storage_options: Extra options that make sense for a particular storage connection, e.g. host, port, username, password, etc.,
         if using a URL that will be parsed by fsspec, e.g., starting “s3://”, “gcs://”. Optional
     :type storage_options: dict
     """
 
     def __init__(self, storage: Driver, stream_path: str, sharding_func: Optional[Callable[[Event], int]] = None, batch_size: int = 8,
-                 columns: Optional[List[str]] = None, infer_columns_from_data: Optional[bool] = None, **kwargs):
+                 columns: Optional[List[str]] = None, infer_columns_from_data: Optional[bool] = None,
+                 shard_count=1, retention_period_hours=24, **kwargs):
```
gtopper: Missing type annotations for new parameters.

storey/targets.py
```diff
@@ -686,9 +691,11 @@ async def _worker(self):
             await self._storage.close()
 
     async def _lazy_init(self):
-        if not self._shard_count:
+        if not self._initialized:
+            await self._storage._create_stream(self._container, self._stream_path, self._shard_count,
+                                               self._retention_period_hours)
+            # get actual number of shards (for pre existing stream)
             response = await self._storage._describe(self._container, self._stream_path)
```
gtopper: Not a big deal, but this is only necessary when `create_stream` failed with 409.

integration/test_flow_integration.py
```diff
@@ -139,12 +140,26 @@ def test_write_to_v3io_stream_with_column_inference(setup_stream_teardown_test):
     ]
 
 
+def test_write_to_pre_existing_stream(setup_stream_teardown_test):
+    stream_path = setup_stream_teardown_test
+    setup_stream(stream_path)
```
gtopper: `setup_stream` is async, so I don't think this actually runs anything.

integration/integration_test_utils.py
```diff
@@ -108,15 +108,13 @@ def setup_kv_teardown_test():
 def setup_stream_teardown_test():
     # Setup
     stream_path = _generate_table_name('bigdata/stream_test')
-    asyncio.run(setup_stream(stream_path))
```
gtopper, line 111: Should probably change the name if it doesn't actually set up the stream anymore. But given the previous comment, I think we'd better leave this method as is and create another one next to it, say `assign_stream_teardown_test`.

integration/test_flow_integration.py
```diff
@@ -153,12 +154,26 @@ def test_write_to_v3io_stream_with_column_inference(setup_stream_teardown_test):
     ]
 
 
-def test_write_dict_to_v3io_stream(setup_stream_teardown_test):
-    stream_path = setup_stream_teardown_test
+def test_write_to_pre_existing_stream(assign_stream_teardown_test):
+    stream_path = assign_stream_teardown_test
+    asyncio.run(create_stream(stream_path))
+    df = pd.DataFrame([['hello', "goodbye"]], columns=['first', 'second'])
+    stream_path = assign_stream_teardown_test
```
```
stream_path = assign_stream_teardown_test
```
twice?

dinal: oops :)

gtopper (approved):

PR #301, opened 2021-10-18
New Flow.to_dict implementation.

gtopper:
Via @yaronha.

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1,6 +1,7 @@
 import asyncio
 import copy
 import datetime
+import inspect
 import time
 import traceback
 from asyncio import Task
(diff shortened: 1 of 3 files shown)


storey/flow.py
```diff
@@ -26,39 +27,56 @@ def __init__(self, recovery_step=None, termination_result_fn=lambda x, y: x if x
         self.logger = getattr(self.context, 'logger', None) if self.context else None
 
         self._kwargs = kwargs
-        self._full_event = kwargs.pop('full_event', False)
```
yaronha, line 29: why not pop (no need to keep it in the kwargs as well)

gtopper: It's only in kwargs. The new `to_dict` doesn't explicitly refer to `_full_event`.

gtopper: Updated

yaronha:
@gtopper LGTM

PR #310, opened 2021-11-04
Support retries in _ConcurrentJobExecution.

gtopper:
Remove redundant call to set_error following #308.

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -603,10 +603,13 @@ class HttpResponse:
 
 
 class _ConcurrentJobExecution(Flow):
-    def __init__(self, max_in_flight=8, retries=0, **kwargs):
+    _BACKOFF_MAX = 120
+
+    def __init__(self, max_in_flight=8, retries=0, backoff_factor=1, **kwargs):
         Flow.__init__(self, **kwargs)
         self._max_in_flight = max_in_flight
         self._retries = retries
+        self._backoff_factor = backoff_factor
 
     def _init(self):
         self._q = None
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -643,17 +647,23 @@ async def _cleanup(self):
...
             try:
                 return await self._process_event(event)
-            except BaseException as ex:
-                none_or_coroutine = event._awaitable_result._set_error(ex)
-                if none_or_coroutine:
-                    await none_or_coroutine
-                raise ex
-        else:
-            return await self._process_event(event)
+            except Exception as ex:
```
urihoenig, line 656: Not sure why it was BaseException in first place, but why has it changed?

storey/flow.py
```diff
@@ -643,17 +647,23 @@ async def _cleanup(self):
...
-    async def _safe_process_event(self, event):
-        if event._awaitable_result:
+    async def _process_event_with_retries(self, event):
+        times_attempted = 0
+        max_attempts = self._retries + 1
+        while True:
             try:
                 return await self._process_event(event)
-            except BaseException as ex:
-                none_or_coroutine = event._awaitable_result._set_error(ex)
```
urihoenig, line 651: Why have you removed it?

tests/test_flow.py
```diff
@@ -2799,3 +2799,37 @@ def test_completion_on_error_in_concurrent_execution_step():
             awaitable_result.await_result()
     finally:
         controller.terminate()
+
+
+@pytest.mark.parametrize('backoff_factor', [(0, 0), (1, 3)])
```
urihoenig: Any need to test failure after retries?

yaronha:
@gtopper where do we differentiate between error types that need to retry or not (e.g. 5xx vs 3xx)

storey/flow.py
gtopper: So the answer is a bit nonobvious. If this throws an exception, it will be caught inside of the `_worker` method, where `_set_error` is already called for any `BaseException`. So this line was redundant and potentially could even cause termination issues (depending on the evolving semantics of completions).

storey/flow.py
gtopper: See [answer above](https://github.com/mlrun/storey/pull/310/files#r744456755).

tests/test_flow.py
gtopper: Added.

gtopper:
@yaronha: In `RemoteStep`, in mlrun. We need to make changes there for that level of control, because I copied the sync behaviour (fail on >=400), but this need to go in a PR to mlrun. This PR to storey is for the generic mechanism.

tests/test_flow.py
```diff
@@ -2799,3 +2799,45 @@ def test_completion_on_error_in_concurrent_execution_step():
...
+        if expected_sleep is None:
+            with pytest.raises(ATestException):
+                awaitable_result.await_result()
+        else:
+            awaitable_result.await_result()
+        end = time.time()
+    finally:
+        controller.terminate()
+    if expected_sleep is None:
+        with pytest.raises(ATestException):
```
urihoenig, line 2839: Why twice?

gtopper: `await_termination` raises the last error that happened. At least that’s the behaviour right now.

urihoenig (approved):

PR #331, opened 2022-01-04
ML-1108: fix aliases (for aggregated features)

katyakats:


katyakats:
fix docstring and also, aliases are now fixed for the original feature name and not the enriched. this may cause a name conflict.

gtopper:
The existing aliases parameter is meant to deconflict augmented field names from existing field names. Using the aliases for the original fields will therefore break this functionality and cause overwriting of fields.

Also, the name "alias" makes sense for the augmented fields (as in, alias of a foreign table's column in SQL), but for an existing field, the correct nomenclature would be "rename" (to alias an existing field would mean we now have it under both names).

The way to move forward without breaking functionality or API would be to add a "rename" parameter for renaming existing fields.

gtopper:
Ticket fixed in mlrun, independent of AggregateByKey.

PR #346, opened 2022-02-24
ML-1108: Allow for aliasing of aggregations.

gtopper:


katyakats (changes requested):
can you add a test?

gtopper:
I adjusted an existing test to cover this case.

katyakats (approved):

PR #352, opened 2022-05-30
Redis driver

alxtkr77:
moved review to https://github.com/mlrun/storey/pull/367 - closing this review.


storey/drivers.py
```diff
@@ -501,3 +508,358 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                expression_value = self._convert_python_obj_to_redis_value(value)
+                # NOTE: This logic assumes that static attributes we're supposed
+                # to delete will appear in the `additional_data` dict with a
+                # "falsey" value. This is the same logic the V3ioDriver uses.
+                if expression_value:
+                    fields_to_set[name] = expression_value
+                else:
+                    fields_to_delete.append(name)
+            if fields_to_set:
+                pipeline.hset(self._static_data_key(redis_key_prefix), mapping=fields_to_set)
```
assaf758, line 741: should hmset

storey/drivers.py
```diff
@@ -501,3 +509,429 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+
+class RedisDriver(Driver):
+    REDIS_TIMEOUT = 5  # Seconds
+    REDIS_WATCH_INTERVAL = 1  # Seconds
+    DATETIME_FIELD_PREFIX = "_dt:"
+    TIMEDELTA_FIELD_PREFIX = "_td:"
+    DEFAULT_KEY_PREFIX = "storey:"
+
+    def __init__(self, redis_client: Optional[Union[redis.Redis, rediscluster.RedisCluster]] = None,
```
assaf758, line 546: please add webapi parameter

.gitignore
```diff
@@ -0,0 +1,154 @@
+### Python template
```
gtopper: Please don't check this in. `cp .gitignore.template .gitignore` and modify as you like.

Pipfile
```diff
@@ -0,0 +1,23 @@
+[[source]]
```
gtopper: Please don't check a Pipfile in. Requirements go in `requirements.txt` or `dev-requirements.txt` or occasionally `docs/requirements.txt`

integration/conftest.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+import fakeredis
+import redis as r
+import pytest
+
+from storey import RedisDriver
+
+REDIS_ENDPOINT = os.environ.get('REDIS_URL')
```
gtopper, line 9: If it's a URL why not call it `REDIS_URL`? Or the other way around if it's an endpoint.

integration/integration_test_utils.py
```diff
@@ -91,6 +94,19 @@ def setup_teardown_test():
...
+def setup_redis_teardown_test(redis, redis_driver):
+    # Setup
+    table_name = _generate_table_name()
+
+    # Test runs
+    yield table_name
+
+    # Teardown
+    for key in redis_driver.redis.scan_iter(f"{redis_driver._key_prefix}*"):
+        redis_driver.redis.delete(key)
```
gtopper: Can we use FLUSHDB?

integration/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+def redis():
+    if REDIS_ENDPOINT:
+        yield r.Redis.from_url(REDIS_ENDPOINT)
+    else:
+        yield fakeredis.FakeRedis(decode_responses=True)
+
+
+@pytest.fixture()
+def redis_key_prefix():
+    yield "storey-test:"
```
gtopper, line 22: Could we create a database for the tests rather than rely on a prefix?

requirements.txt
```diff
@@ -9,3 +9,6 @@ grpcio-tools>1.34.0,<1.42
 grpcio>1.34.0,<1.42
 v3io-frames>=0.10.2,<1
 v3iofs~=0.1.9
+redis~=3.5
+redis-py-cluster~=2.1.3
+lupa~=1.13
```
gtopper, line 14: These should probably be in `extras_require` (see `setup.py`).

storey/drivers.py
```diff
@@ -1,17 +1,22 @@
...
-from typing import Optional
+import time
+from datetime import datetime, timedelta, timezone
+from enum import Enum
 from collections import OrderedDict
+from typing import List, Union, Optional
 import pandas as pd
 
+import redis
+import rediscluster
```
gtopper, line 13: We probably want to move these imports either to an inner scope or to another file, so that users (current and future) can use this library without installing redis (same idea as moving it to `extra-requires` as mentioned in another comment).

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+        return lua_script, condition_expression, pending_updates, redis_keys_involved
+
+    async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data):
+        redis_key_prefix = self.make_key(container, table_path, key)
+        update_expression, mtime_condition, pending_updates, redis_keys_involved = self._build_feature_store_lua_update_script(
+            redis_key_prefix, aggr_item, partitioned_by_key, additional_data)
+        if not update_expression:
+            return
+        current_time = int(time.time_ns()/1000)
```
gtopper, line 779: We don't currently check formatting in CI, and so far this has never been a problem because we all used PyCharm... I guess we may have to introduce autoformatting.

integration/conftest.py
alxtkr77: Actually there is no 'database' in redis:
https://stackoverflow.com/questions/35621324/how-to-create-own-database-in-redis#:~:text=You%20don't%20create%20a,between%20the%20databases%2C%20call%20SELECT%20.&text=Show%20activity%20on%20this%20post.,-Use%20select%20%2C%20for

requirements.txt
alxtkr77: What is the guide to separate requirements.txt and extras_require()?

storey/drivers.py
alxtkr77: Means the v3io imports should be moved as well? Where would you move it(what additional modules do you suggest)?

storey/drivers.py
alxtkr77: Should I change anything or its a general comment?

integration/conftest.py
gtopper: We decided to maybe consider this later.

storey/drivers.py
gtopper: We discussed that it would be good to align the formatter with pycharm (or it might already be aligned but was not used).

storey/drivers.py
gtopper: We discussed moving this into a new file (redis_driver.py or something similar).

requirements.txt
gtopper: We discussed the difference between ordinary requirements and extra requirements (the latter are optional).

integration/test_aggregation_with_redis.py
```diff
@@ -0,0 +1,1854 @@
+import asyncio
```
gtopper, line 1: Would be good to avoid duplicating this giant test file.

tests/test_flow.py
```diff
@@ -3324,3 +3328,67 @@ def test_rename():
     controller.terminate()
     termination_result = controller.await_termination()
     assert termination_result == [{'b': 1, 'c': 3, 'e': 5}]
+
```
gtopper, line 3331: Should be 2 blank lines.

storey/targets.py
```diff
@@ -880,3 +880,65 @@ async def _do(self, event):
                 self._table._update_static_attrs(key, data_to_persist)
             self._table._init_flush_task()
             await self._do_downstream(event)
+
+
+class RedisNoSqlTarget(_Writer, Flow):
+    """
+    Persists the data in `table` to its associated storage by key.
```
gtopper, line 887: Why do we need to duplicate `NoSqlTarget`? Afaict only the driver needs to change, not the target. Same as we don't have a `JoinWithRedisTable`.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+class NeedsRedisAccess:
+    """Checks that params for access to Redis exist and are legal
+
+    :param webapi: URL to the web API (https or http). If not set, the REDIS_URL environment variable will be used.
+
+    """
+
+    def __init__(self, webapi=None):
+        webapi = webapi or os.getenv('REDIS_URL')
```
gtopper, line 527: I think better call it endpoint or url or something. webapi is what we call it in v3io, and what we have here is not even http(s).

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                return None
+            return f'{cls.DATETIME_FIELD_PREFIX}{value.timestamp()}'
+        elif isinstance(value, timedelta):
+            return f'{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}'
+        else:
+            if isinstance(value, float):
+                if value == math.inf or value == -math.inf:
+                    # Use the shorter infinity form (inf, -inf) to save space.
+                    value = f'{str(value)}'
+                elif not value % 1:
```
gtopper, line 618: A bit nitpicky, but I think
```
value % 1 == 0
```
makes more sense in this context.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            return f'"{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}"'
+        elif isinstance(value, bool):
+            return f'"{json.dumps(value)}"'
+        else:
+            if isinstance(value, float):
+                if value == math.inf:
+                    value = "math.huge"
+                elif value == -math.inf:
+                    value = "-math.huge"
+                elif not value % 1:
```
gtopper, line 638: A bit nitpicky, but I think
```
value % 1 == 0
```
makes more sense in this context.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                    value = int(value)
+            return json.dumps(value)
+
+    def _convert_python_obj_to_lua_value(cls, value):
+        if isinstance(value, datetime):
+            if pd.isnull(value):
+                return None
+            return f'"{cls.DATETIME_FIELD_PREFIX}{value.timestamp()}"'
+        elif isinstance(value, timedelta):
+            return f'"{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}"'
```
gtopper, line 629: This is a bit awkward. It looks like the v3io implementation just turns a timedelta into a string. Here, it's actually implemented as a time delta. Which is cool, but I don't think this type should have different sematics based on the DB.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                    # Store whole numbers as integers to save space.
+                    value = int(value)
+
+            return json.dumps(value)
+
+    @classmethod
+    def _convert_redis_value_to_python_obj(cls, value):
+        value = cls._convert_to_str(value)
+        if value.startswith(cls.TIMEDELTA_FIELD_PREFIX):
+            return timedelta(seconds=float(value.split(":")[1]))
```
gtopper, line 648: This is a bit awkward. It looks like the v3io implementation just turns a timedelta into a string. Here, it's actually implemented as a time delta. Which is cool, but I don't think this type should have different sematics based on the DB.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        if pending:
+            last_time = int(list(pending.keys())[-1] / max_window_millis) * max_window_millis
+            min_time = last_time - max_window_millis
+            for _time, value in pending.items():
+                if _time > min_time:
+                    res[_time] = value
+        return res
+
+    def _build_feature_store_lua_update_script(self, redis_key_prefix, aggregation_element, partitioned_by_key, additional_data):
+        lua_script = None
```
gtopper, line 682: Redundant assignment.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        if additional_data:
+            redis_keys_involved.append(self._static_data_key(redis_key_prefix))
+            additional_data_lua_script = f'local additional_data_key="{self._static_data_key(redis_key_prefix)}";\n'
+            for name, value in additional_data.items():
+                expression_value = self._convert_python_obj_to_lua_value(value)
+                # NOTE: This logic assumes that static attributes we're supposed
+                # to delete will appear in the `additional_data` dict with a
+                # "falsey" value. This is the same logic the V3ioDriver uses.
+                if expression_value:
+                    # additional_data_lua_script = f"""{additional_data_lua_script}
```
gtopper, line 698: Dead code ☠️

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            if update_ok and aggr_item:
+                await self._fetch_state_by_key(aggr_item, container, table_path, key)
+            # else:
+                # should_raise_error = True
+
+        # if should_raise_error:
+        #     raise RedisError(
+        #         f'Failed to save aggregation for {table_path}/{key}. Response status code was {response.status_code}: {response.body}\n'
+        #         f'Update expression was: {update_expression}'
+        #     )
```
gtopper, line 810: Dead code ☠️

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+    async def _fetch_state_by_key(self, aggr_item, container, table_path, key):
+        key = str(key)
+        # Aggregation Redis keys start with the Redis key prefix for this Storey container, table
+        # path, and "key," followed by ":aggr_"
+        aggregations = {}
+
+        redis_key_prefix = self.make_key(container, table_path, key)
+        aggr_key_prefix = f"{redis_key_prefix}:{self._aggregation_attribute_prefix}"
+        # XXX: We can't use `async for` here...
```
gtopper, line 922: What does XXX stand for? :)

storey/table.py
```diff
@@ -1027,6 +1033,9 @@ def reset(self, value=None):
         else:
             self.value = float(value)
 
+    def aggregate_lua_script(self, vl1, vl2):
+        return 'type({}) == "number" and math.min({},{}) or {}'.format(vl1, vl1, vl2, vl2)
```
gtopper, line 1037: Any reason not to use modern f notation? i.e.:
```
return f'type({vl1}) == "number" and math.min({vl1},{vl2}) or {vl2}'
```
The same applies to other uses of `format()`.

storey/drivers.py
```diff
@@ -449,6 +456,7 @@ async def _load_aggregates_by_key(self, container, table_path, key):
                         aggregations[feature_and_aggr_name] = {}
                     aggregations[feature_and_aggr_name][time_in_millis] = value
                     aggregations[feature_and_aggr_name][associated_time_attr] = time_in_millis
+
```
gtopper, line 459: Please don't...

storey/utils.py
```diff
@@ -304,3 +307,21 @@ def find_filters(partitions_time_attributes, start, end, filters, filter_column)
...
+        if col.startswith('igzpart_'):
+            cols_to_drop.append(col)
+    df.drop(labels=cols_to_drop, axis=1, inplace=True, errors='ignore')
+
+
+def asyncify(fn):
+    """Run a synchronous function asynchronously."""
+    async def inner_fn(*args, **kwargs):
+        loop = asyncio.get_event_loop()
+        with concurrent.futures.ThreadPoolExecutor() as pool:
```
gtopper, line 324: Hm, is this creating a new executor every time? In existing storey code, we pass `None` for the executor, to indicate the default instance should be used.

storey/drivers.py
gtopper: This issue is still outstanding.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            if pd.isnull(value):
+                return None
+            return f'{cls.DATETIME_FIELD_PREFIX}{value.timestamp()}'
+        elif isinstance(value, timedelta):
+            return f'{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}'
+        else:
+            if isinstance(value, float):
+                if value == math.inf or value == -math.inf:
+                    # Use the shorter infinity form (inf, -inf) to save space.
+                    value = f'{str(value)}'
```
gtopper, line 617: `str` and `f'...'` are redundant.
```suggestion
                    value = str(value)
```

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            if isinstance(value, float):
+                if value == math.inf:
+                    value = "math.huge"
+                elif value == -math.inf:
+                    value = "-math.huge"
+                elif not value % 1:
+                    # Store whole numbers as integers to save space.
+                    value = int(value)
+
+            return json.dumps(value)
```
gtopper, line 642: Does it actually make sense to serialize the value to json?

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            return f'{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}'
+        else:
+            if isinstance(value, float):
+                if value == math.inf or value == -math.inf:
+                    # Use the shorter infinity form (inf, -inf) to save space.
+                    value = f'{str(value)}'
+                elif not value % 1:
+                    # Store whole numbers as integers to save space.
+                    value = int(value)
+            return json.dumps(value)
```
gtopper, line 621: Does it actually make sense to serialize the value to json?

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        elif value.startswith(cls.DATETIME_FIELD_PREFIX):
+            return datetime.fromtimestamp(float(value.split(":")[1]), tz=timezone.utc)
+        elif value == "inf":
+            return math.inf
+        elif value == "-inf":
+            return -math.inf
+        try:
+            return json.loads(value)
+        except Exception:
+            return value
```
gtopper, line 658: This could be pretty expensive, if it tries to parse each value as json and raises an exception (potentially for every row x column), only to return the original value. What redis types does this cover?

tests/test_flow.py
gtopper: I guess I'll just reformat it later. Probably a bit late to add black now, but it's an option too.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            return
+        current_time = int(time.time_ns()/1000)
+        if mtime_condition is not None:
+            update_expression = f'if redis.call("HGET", "{redis_key_prefix}","{self._mtime_name}") == "{mtime_condition}" then \
+                {update_expression} redis.call("HSET","{redis_key_prefix}","{self._mtime_name}",{current_time});return 1;else return 0;end;'
+        else:
+            update_expression = f'{update_expression}redis.call("HSET","{redis_key_prefix}","{self._mtime_name}",{current_time});return 1;'
+
+        redis_keys_involved.append(redis_key_prefix)
+        update_expression = f"""{update_expression}"""
```
gtopper, line 787: This is really just
```suggestion
        update_expression = str(update_expression)
```

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+    async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data):
+        redis_key_prefix = self.make_key(container, table_path, key)
+        update_expression, mtime_condition, pending_updates, redis_keys_involved = self._build_feature_store_lua_update_script(
+            redis_key_prefix, aggr_item, partitioned_by_key, additional_data)
+        if not update_expression:
+            return
+        current_time = int(time.time_ns()/1000)
+        if mtime_condition is not None:
+            update_expression = f'if redis.call("HGET", "{redis_key_prefix}","{self._mtime_name}") == "{mtime_condition}" then \
```
gtopper, line 781: Consistency
```suggestion
            update_expression = f'if redis.call("HGET","{redis_key_prefix}","{self._mtime_name}") == "{mtime_condition}" then \
```

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        return values
+
+    async def _get_associated_time_attr(self, redis_key_prefix, aggr_name):
+        aggr_without_relevant_attr = aggr_name[:-2]
+        feature_name_only = aggr_without_relevant_attr[:aggr_without_relevant_attr.rindex('_')]
+        feature_with_relevant_attr = f"{feature_name_only}{aggr_name[-2:]}"
+        associated_time_key = self._aggregation_time_key(redis_key_prefix,
+                                                         feature_with_relevant_attr)
+        time_val = await asyncify(self.redis.get)(associated_time_key)
+        time_val = self._convert_to_str(time_val)
```
gtopper, line 857: From what I see, we assign a string to `time_val`, convert it to a string (?) then convert it to an int just below. Could it be that this line is unnecessary?

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+    @classmethod
+    def _convert_to_str(cls, key):
+        try:
+            if isinstance(key, bytes):
+                return key.decode('utf-8')
+            else:
+                return key
+        except Exception as e:
+            print(e)
+            # Logs the error appropriately.
```
gtopper, line 669: Why add this catch and print? What error are we expecting here? And in case of an error, it will be suppressed and the function will return `None`, which will then fail a bit further down the line with a confusing error (or worse – do something unexpected).

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                                        do redis.call("RPUSH",list_attribute_key,{default_value}) end;\n'
+                                if array_time_attribute_key not in times_updates:
+                                    times_updates[array_time_attribute_key] = expected_time_expr
+                                new_cached_times[name] = (array_time_attribute_key, expected_time)
+
+                            # Updating the specific cells
+                            if cached_time <= expected_time:
+                                # condition = None
+                                # if partitioned_by_key and self._mtime_name in aggregation_element.storage_specific_cache:
+                                #     condition = aggregation_element.storage_specific_cache[self._mtime_name]
```
gtopper, line 760: Dead code ☠️

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+                                lua_script = f'{lua_script}local old_value=redis.call("LINDEX", list_attribute_key, {index_to_update});\
```
gtopper, line 761: I think it's better to settle on no space after commas, or with space after commas, but not a mixture of the two.

storey/drivers.py
```diff
@@ -501,3 +509,450 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            if partitioned_by_key:
+                condition_expression = aggregation_element.storage_specific_cache.get(self._mtime_name, None)
+            lua_tonum_function = 'local function tonum(str) if str == "inf" then return math.huge elseif str == "-inf" \
+                then return -math.huge end return tonumber(str) end'
+            lua_script = f'{lua_script}local len;local redis_key_prefix="{redis_key_prefix}";local list_attribute_key;\n'
+            lua_script = f'{lua_script}{lua_tonum_function}\n'
+            for name, bucket in aggregation_element.aggregation_buckets.items():
+                # Only save raw aggregates, not virtual
+                if bucket.should_persist:
+                    # In case we have pending data that spreads over more then 2 windows, discard the old ones.
```
gtopper, line 720: then -> than

storey/dtypes.py
```diff
@@ -60,6 +60,14 @@ class V3ioError(Exception):
     pass
 
 
+class RedisError(Exception):
+    pass
+
+
+class RedisTimeoutError(Exception):
```
gtopper, line 67: This class is not used anywhere.

storey/table.py
```diff
@@ -970,6 +970,12 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         raise NotImplementedError()
 
+    def aggregate_lua_script(self):
+        raise NotImplementedError()
+
+    def aggregate_argument(self, time, argument):
```
gtopper, line 976: This method is not used anywhere.

storey/table.py
```diff
@@ -1064,6 +1079,12 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         self._set_value(self.value + 1)
 
+    def aggregate_argument(self, time, argument):
```
gtopper, line 1082: This method is not used anywhere.

storey/table.py
```diff
@@ -1075,6 +1096,12 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         self._set_value(self.value + value * value)
 
+    def aggregate_argument(self, time, argument):
```
gtopper, line 1099: This method is not used anywhere.

storey/table.py
```diff
@@ -1064,6 +1079,12 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         self._set_value(self.value + 1)
 
+    def aggregate_argument(self, time, argument):
+        self._set_value(self.value + argument)
+
+    def aggregate_lua_script(self, vl1, vl2):
+        return '{}+{}'.format(vl1, vl2)
```
gtopper, line 1086: Hm, shouldn't this be
```suggestion
        return f'{vl1}+1'
```
(changed the style, but I'm referring to the fact that I expected a +1 here)

If I'm right, why didn't this fail any of the aggregation tests...?

storey/targets.py
alxtkr77: Agree

integration/test_aggregation_with_redis.py
alxtkr77: Right. Would do as a lower priority

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree. Should be removed

storey/table.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/utils.py
alxtkr77: Need to dig in. Not my code

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: This used to convert any type to string

storey/drivers.py
alxtkr77: This used to convert any type to string

storey/drivers.py
alxtkr77: To tell the truth I don't remember :). Should be tested without the exception

storey/drivers.py
alxtkr77: Is it? I need the " to be part of the result string

storey/drivers.py
alxtkr77: redis returns bstr , not str

storey/drivers.py
alxtkr77: Agree . Should be removed

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/drivers.py
alxtkr77: Agree

storey/dtypes.py
alxtkr77: Agree. Should be removed

storey/table.py
alxtkr77: Agree. Should be removed

storey/table.py
alxtkr77: Agree. Should be removed

storey/table.py
alxtkr77: Agree. Should be removed

storey/table.py
alxtkr77: Not always. On aggregation I do count1+count2 and not count1+1

integration/conftest.py
assaf758: I'm not really sure what is the difference between URL, API or ENDPOINT :-)
mlrun use API - do you want to settle on that?

requirements.txt
assaf758: done

integration/conftest.py
assaf758: fixed

storey/drivers.py
assaf758: done

storey/drivers.py
assaf758: ran flake8

tests/test_flow.py
assaf758: fixed

storey/drivers.py
assaf758: done

storey/drivers.py
assaf758: done

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/table.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed - removed the try/except

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/dtypes.py
assaf758: fixed

storey/table.py
assaf758: fixed

storey/table.py
assaf758: fixed

storey/table.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/utils.py
assaf758: yes, it's using multiple threads.
I believe the motivation is to workaround the Redis client not being async.
in order to reuse the thread pool across all RedisDriver requests, i made the the pool a RedisDriver class var.

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/targets.py
assaf758: Removed

storey/drivers.py
assaf758: i'm not sure why you think it's awkward - different drivers use different implementations.. we can discuss further offline.

storey/drivers.py
assaf758: same as above

integration/test_aggregation_with_redis.py
assaf758: Done.

PR #359, opened 2022-06-09
ML-2272: Add MongoDB source, target and driver(doesn't support aggregate)

davesh0812:


requirements.txt
```diff
@@ -9,3 +9,4 @@ grpcio-tools>1.34.0,<1.42
 grpcio>1.34.0,<1.42
 v3io-frames>=0.10.2,<1
 v3iofs~=0.1.9
+pymongo~=4.1.1
```
gtopper: Should probably be in `extras_require` (`setup.py`) rather than here.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
             self._dfs.append(df)
+
+    class MongoDBSource(_IterableSource, WithUUID):
+        """Use mongodb collection as input source for a flow.
+
+        :parameter key_field: column to be used as key for events. can be list of columns
+        :parameter time_field: column to be used as time for events.
+        :parameter id_field: column to be used as ID for events.
+        :parameter db_name: the name of the database as mention on mongodb
+        :parameter connection_string:
```
gtopper: Needs a description.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+        :parameter collection_name: the name of the collection as mention on mongodb
```
gtopper: > as mention on mongodb

❓

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+
+        for additional params, see documentation of  :class:`~storey.flow.Flow`
+        """
+
+        def __init__(
+                self,
+                db_name: str = None,
+                connection_string: str = None,
+                collection_name: str = None,
+                query: dict = None,
```
gtopper: This field is not documented.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+        """Use mongodb collection as input source for a flow.
+
+        :parameter key_field: column to be used as key for events. can be list of columns
+        :parameter time_field: column to be used as time for events.
+        :parameter id_field: column to be used as ID for events.
+        :parameter db_name: the name of the database as mention on mongodb
+        :parameter connection_string:
+        :parameter collection_name: the name of the collection as mention on mongodb
+
+        for additional params, see documentation of  :class:`~storey.flow.Flow`
```
gtopper: There is no documentation there.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+
+        def __init__(
+                self,
+                db_name: str = None,
+                connection_string: str = None,
+                collection_name: str = None,
+                query: dict = None,
+                key_field: Optional[Union[str, List[str]]] = None,
+                start_filter: Optional[datetime] = None,
+                end_filter: Optional[datetime] = None,
```
gtopper: These fields are not documented.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                time_field: Optional[str] = None,
+                id_field: Optional[str] = None,
+                **kwargs,
+        ):
+
+            if end_filter or start_filter:
+                start_filter = datetime.min if start_filter is None else start_filter
+                end_filter = datetime.max if end_filter is None else end_filter
```
gtopper: Can simplify as
```
                start_filter = start_filter or datetime.min
                end_filter = end_filter or datetime.max
```
But do you even have to add these to the dictionary if they're not set?

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                time_query = {time_field: {"$gte": start_filter, "$lt": end_filter}}
```
gtopper: What if `time_field` is not set?

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                raise TypeError(
+                    "cannot specify without connection_string, db_name and collection_name args"
+                )
+
+            from pymongo import MongoClient
+
+            mongodb_client = MongoClient(connection_string)
+            my_db = mongodb_client[db_name]
+            my_collection = my_db[collection_name]
+            self._my_collection = my_collection.find(query)
```
gtopper: Actual IO should go in `_run_loop` rather than in the constructor.

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                    event = Event(body, key=key, time=time, id=id)
+                    await self._do_downstream(event)
+                else:
+                    if self.context:
+                        self.context.logger.error(
+                            f"For {body} value of key {key_field} is None"
+                        )
+            return await self._do_downstream(_termination_obj)
+
+        def get_val_from_multi_dictionry(self, event, field):
```
gtopper: dictionry -> dictionary

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+            if key_field is not None:
+                kwargs["key_field"] = key_field
+            if time_field is not None:
+                kwargs["time_field"] = time_field
+            if id_field is not None:
+                kwargs["id_field"] = id_field
+            _IterableSource.__init__(self, **kwargs)
+            WithUUID.__init__(self)
+
+            if db_name is None or collection_name is None or connection_string is None:
```
gtopper: Can simplify as
```
not all([db_name, collection_name, connection_string])
```

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                raise TypeError(
```
gtopper: TypeError -> ValueError

storey/sources.py
```diff
@@ -875,3 +875,111 @@ def _init(self):
...
+                    body["_id"] = str(body["_id"])
+
+                key = None
+                if self._key_field:
+                    if isinstance(self._key_field, list):
+                        key = []
+                        for key_field in self._key_field:
+                            if key_field not in body or pandas.isna(body[key_field]):
+                                create_event = False
+                                break
```
gtopper, line 958: Pretty sure you don't need to set `create_event` if you're breaking anyway.

gtopper: Also, are you sure it's correct to just break in such a case?

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
             else:
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
             self._dfs.append(df)
+
+    class MongoDBSource(_IterableSource, WithUUID):
+        """Use mongodb collection as input source for a flow.
+
+        :parameter key_field: string. column to be used as key for events. can be list of columns
```
gtopper: The parameter types are annotated, so no need to put it in the class comment. Also the comment doesn't match the type annotation.

setup.py
```diff
@@ -27,6 +27,7 @@ def load_deps(file_name):
     "s3": ["s3fs~=0.5"],
     "az": ["adlfs~=0.5"],
     "kafka": ["kafka-python~=2.0"],
+    "mongodb": ["pymongo~=4.1.1"],
```
gtopper, line 30: Do we have to be so restrictive as to require the latest version only?

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+        :parameter time_field: string. column to be used as time for events.
+        :parameter id_field: string. column to be used as ID for events.
+        :parameter db_name: string. the name of the database you would like to access as it mention on mongodb
+        :parameter connection_string: string. your mongodb connection string
+        :parameter collection_name: string. the collection name you would like to access,
```
gtopper: the name of the collection to access

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+                                    from the current db as it mention on mongodb
```
gtopper: Still not sure what
> as it mention on mongodb

means.

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+                id_field: Optional[str] = None,
+                **kwargs,
+        ):
+
+            if time_field:
+                time_query = {time_field: None}
+                if start_filter:
+                    time_query[time_field] = {"$gte": start_filter}
+                if end_filter:
+                    time_query[time_field]["$lt"] = end_filter
```
gtopper: I'm pretty sure this will raise an error if you pass an `end_filter` but no `start_filter`.

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+                if time_query[time_field] is not None and query is not None:
+                    query.update(time_query)
+                elif time_query[time_field] is not None and query is None:
+                    query = time_query
```
gtopper: Could simplify the ifs here, but can query even be None? Does that work?

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+            if not all([db_name, collection_name, connection_string]):
+                raise ValueError(
+                    "cannot specify without connection_string, db_name and collection_name args"
+                )
+            self.attrs = {
+                "query": query,
+                "collection_name": collection_name,
+                "db_name": db_name,
+                "connection_string": connection_string,
+            }
```
gtopper: Why not set these directly on `self`?

storey/sources.py
```diff
@@ -875,3 +875,124 @@ def _init(self):
...
+                self._id_field = id_field
+
+        async def _run_loop(self):
+            from pymongo import MongoClient
+
+            mongodb_client = MongoClient(self.attrs['connection_string'])
+            my_db = mongodb_client[self.attrs["db_name"]]
+            my_collection = my_db[self.attrs["collection_name"]]
+            my_collection = my_collection.find(self.attrs["query"])
+            for body in my_collection:
```
```suggestion
            collection = my_db[self.attrs["collection_name"]]
            cursor = my_collection.find(self.attrs["query"])
            for body in cursor:
```
Nomenclature.

gtopper (commented):
It would be advisable to add a test.

PR #363, opened 2022-07-04
ML-2286: Allow PyArrow to set parquet schema

gtopper:
when columns and infer_columns_from_data are not passed.

### DIFF
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -27,10 +27,7 @@ class _Writer:
     def __init__(self, columns: Union[str, List[Union[str, Tuple[str, str]]], None], infer_columns_from_data: Optional[bool],
                  index_cols: Union[str, List[Union[str, Tuple[str, str]]], None] = None, partition_cols: Union[str, List[str], None] = None,
                  retain_dict: bool = False, storage_options: Optional[dict] = None):
-        if infer_columns_from_data is None:
-            infer_columns_from_data = not bool(columns)
         self._infer_columns_from_data = infer_columns_from_data
-
         self._metadata_columns = {}
         self._metadata_index_columns = {}
         self._rename_columns = {}
@@ -214,6 +211,12 @@ class _Writer:
                     new_data.append(new_value)
                 else:
                     new_data[column] = new_value
+            return new_data
+        elif isinstance(new_data, dict):
+            for column in metadata_columns:
+                metadata_attr = metadata_columns[column]
+                value = getattr(event, metadata_attr)
+                new_data[column] = value
 
     @staticmethod
     def _get_column_data_from_list(new_data, event, original_data, columns, metadata_columns):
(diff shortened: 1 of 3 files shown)


storey/targets.py
```diff
@@ -299,6 +304,9 @@ class CSVTarget(_Batching, _Writer):
 
     def __init__(self, path: str, columns: Optional[List[str]] = None, header: bool = False, infer_columns_from_data: Optional[bool] = None,
                  max_lines_before_flush: int = 128, max_seconds_before_flush: int = 3, **kwargs):
+        if not columns:
```
assaf758, line 306: what if columns is not provided and user's infer_columns_from_data is false - do we silently overwrite infer_columns_from_data to true?

storey/targets.py
```diff
@@ -378,8 +386,9 @@ class ParquetTarget(_Batching, _Writer):
         Optional. Defaults to None (will be inferred if event is dictionary).
     :param infer_columns_from_data: Whether to infer columns from the first event, when events are dictionaries.
         If True, columns will be inferred from data and used in place of explicit columns list if none was provided, or appended to the
-        provided list. If header is True and columns is not provided, infer_columns_from_data=True is implied.
-        Optional. Default to False if columns is provided, True otherwise.
+        provided list. If columns were not provided and infer_columns_from_data=False, PyArrow will infer the schema per file written,
+        which may cause the schemas to differ between files (e.g. if a column in all null in one file but not in another).
```
assaf758: in all => is all

storey/targets.py
```diff
@@ -214,6 +211,12 @@ def _get_column_data_from_dict(self, new_data, event, columns, columns_types, me
                     new_data.append(new_value)
                 else:
                     new_data[column] = new_value
+            return new_data
```
assaf758: why do you need a return new_data? btw, no caller of this function use the returned object..

storey/targets.py
```diff
@@ -450,9 +460,12 @@ def _event_to_batch_entry(self, event):
 
     async def _emit(self, batch, batch_key, batch_time, last_event_time=None):
         df_columns = []
-        if self._index_cols:
-            df_columns.extend(self._index_cols)
-        df_columns.extend(self._non_partition_columns)
+        if self._non_partition_columns:
```
assaf758, line 462: is this related to the infer_columns_from_data change or additional unrelated change?

storey/targets.py
gtopper: We do, as is documented under `:param infer_columns_from_data:`.

storey/targets.py
gtopper: It's related. Without this change, you may get a schema with only the index columns, which we do not want.

storey/targets.py
gtopper: No need. Refactoring accident.

assaf758 (approved):

PR #364, opened 2022-07-10
ML-2383: Make `Complete` NoOp when there is no awaitable to complete

gtopper:


storey/flow.py
```diff
@@ -608,9 +608,10 @@ async def _do(self, event):
         termination_result = await self._do_downstream(event)
         if event is not _termination_obj:
             result = self._get_event_or_body(event)
```
assaf758: you can move the "result" computation inside the awaitable_result "if" clause

assaf758 (approved):

PR #367, opened 2022-08-03
ML-1913: Redis based nosql driver

assaf758:
this replaces Alex PR at https://github.com/mlrun/storey/pull/352


integration/test_aggregation_with_redis.py
```diff
@@ -0,0 +1,1853 @@
+import asyncio
```
gtopper: Do we really have to duplicate `integration/test_aggregation_integration.py`? Can't we just parameterize the suite?

setup.py
```diff
@@ -27,6 +27,9 @@ def load_deps(file_name):
     "s3": ["s3fs~=0.5"],
     "az": ["adlfs~=0.5"],
     "kafka": ["kafka-python~=2.0"],
+    "redis": ["redis~=3.5"],
+    "redis-cluster": ["redis-py-cluster~=2.1.3"],
+    "lupa": ["lupa~=1.13"],
```
gtopper: Any reason that redis and lupa are pinned to a major version and redis-cluster is pinned to a minor version?

tests/test_flow.py
```diff
@@ -407,6 +408,9 @@ def test_write_parquet_read_parquet_partitioned(tmpdir):
 
 
 from unittest.mock import MagicMock
+from storey.redis_driver import RedisDriver
+from storey.aggregations import AggregateByKey
+from storey.dtypes import FieldAggregator, FixedWindows, SlidingWindows
```
gtopper: These imports are not used.

tests/test_flow.py
```diff
@@ -3324,3 +3328,68 @@ def test_rename():
     controller.terminate()
     termination_result = controller.await_termination()
     assert termination_result == [{'b': 1, 'c': 3, 'e': 5}]
+
+
+def test_redis_driver_write(redis):
```
gtopper: This file is for unit tests. These are actually integration tests when `REDIS_URL` is set.

storey/utils.py
```diff
@@ -304,3 +304,11 @@ def find_filters(partitions_time_attributes, start, end, filters, filter_column)
     # for start=1.2.2018 08:53:15, end=5.2.2018 16:24:31, this method will append to filters
     # [(year=2018, month=2,day<=5, filter_column<5.2.2018 16:24:31)]
     _find_filter_helper(partitions_time_attributes, end, "<", "<=", first_uncommon, filters, filter_column)
+
+
+def drop_reserved_columns(df):
```
gtopper: This function is added but never used.

storey/table.py
```diff
@@ -1075,6 +1090,9 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         self._set_value(self.value + value * value)
 
+    def aggregate_lua_script(self, vl1, vl2):
+        return f'{vl1}+{vl2}'
```
gtopper, line 1105: Hm, this doesn't look like the right implementation for sum of squares. Unless I'm missing something, we should add a test that catches this.

storey/table.py
```diff
@@ -1064,6 +1076,9 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         self._set_value(self.value + 1)
 
+    def aggregate_lua_script(self, vl1, vl2):
+        return f'{vl1}+{vl2}'
```
gtopper, line 1088: This doesn't look like the right implementation for count. Unless I'm missing something, we should add a test that catches this.

storey/table.py
```diff
@@ -1042,6 +1048,9 @@ def aggregate(self, time, value):
     def get_update_expression(self, old):
         return f'max({old}, {self.value})'
 
+    def aggregate_lua_script(self, vl1, vl2):
+        return f'type({vl1}) == "number" and math.max({vl1},{vl2}) or {vl2}'
```
gtopper, line 1060: Not sure I understand this... Is this supposed to be the value or a condition? Some of the implementations have what seems to be a value, e.g. `f'{vl1}+{vl2}'`, and some have what seems to be a condition. Also the signature here doesn't match the one in the super class.

setup.py
assaf758: no reason, fixed

tests/test_flow.py
assaf758: removed

tests/test_flow.py
assaf758: well, we do have V3io driver running on other tests in this file - what's the difference?
anyway, can move it to integration tests if you prefer.

storey/utils.py
assaf758: removed

storey/table.py
assaf758: the expected Sqr math function (x*x) is done in the aggregate function.
the aggregate_lua_script() is used as component for building the lua script. It is called with vl1 containing the old value and vl2 is the square of the value to be added.
so there is no issue here (and there are tests covering this functionality)

storey/table.py
assaf758: similar explanation to SqrValue()

storey/table.py
assaf758: neither a value or condition - it's part of lua script, to be joined with other parts of the script (see lua_script in _build_feature_store_lua_update_script).
Changed the super-class signature and added a comment explaining this method.

tests/test_flow.py
gtopper: There is no V3IO driver running in `test_flow.py`, because it's a unit test suite. Ideally fake redis will run in unit tests and real redis in integration tests, and without code duplication of course :).

integration/test_aggregation_with_redis.py
assaf758: Done

tests/test_flow.py
assaf758: moved redis specific tests to out of test_flow.py and into new integration file test_redis_specific.py

storey/table.py
```diff
@@ -970,6 +970,14 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
     def aggregate(self, time, value):
         raise NotImplementedError()
 
+    def aggregate_lua_script(self, vl1, vl2):
+        '''
```
gtopper:
```suggestion
        """
```

storey/table.py
```diff
@@ -970,6 +970,14 @@ def __init__(self, max_value=None, set_data=None, set_time=None):
...
+        The aggregate_lua_script() method is used for creating part of a lua script, to be sent
+        to the Redis DB (see _build_feature_store_lua_update_script).
+        The math function (for example Sqr: x*x, increment: x+1) is implemented in the aggregate method.'''
```
gtopper:
```suggestion
        The math function (for example Sqr: x*x, increment: x+1) is implemented in the aggregate method.
        """
```

integration/test_aggregation_integration.py
```diff
@@ -1550,7 +1550,11 @@ def map_multiply(x):
 
 
 def test_write_read_first_last(setup_teardown_test):
-    table = Table(setup_teardown_test, V3ioDriver(), flush_interval_secs=None)
+
+    if setup_teardown_test["driver_name"] == "RedisDriver":
+        pytest.skip() # TODO
```
gtopper: Let's not forget this one.

integration/test_flow_integration.py
```diff
@@ -58,13 +58,35 @@ async def get_shard_data(self, path):
         return data
 
 
+def _get_redis_kv_all_attrs(setup_teardown_test: dict, key:str):
+    from .integration_test_utils import get_redis_client
+    from storey import RedisDriver
+    table_name = setup_teardown_test["table_name"]
+    redis_key = ''.join(('storey-test:', table_name, key, ':static'))
```
```suggestion
    redis_key = f'storey-test:{table_name}{key}:static'
```

integration/test_flow_integration.py
```diff
@@ -58,13 +58,35 @@ async def get_shard_data(self, path):
...
+    redis_fake_server = setup_teardown_test["redis_fake_server"] if "redis_fake_server" in setup_teardown_test else None
+    values = get_redis_client(redis_fake_server= redis_fake_server).hgetall(redis_key)    
+    return {RedisDriver.convert_to_str(key): RedisDriver.convert_redis_value_to_python_obj(val) for key, val in values.items()}
+
+def get_key_all_attrs_test_helper(setup_teardown_test: dict, key:str):
+    result = None
```
gtopper: This line is redundant.

integration/test_flow_integration.py
```diff
@@ -58,13 +58,35 @@ async def get_shard_data(self, path):
...
+    else:
+        response = asyncio.run(get_kv_item(setup_teardown_test["table_name"], key))
+        assert response.status_code == 200
+        result = response.output.item
+    return result
+
 def test_join_with_v3io_table(setup_kv_teardown_test):
-    table_path = setup_kv_teardown_test
+    if setup_kv_teardown_test["driver_name"] == "RedisDriver":
+        pytest.skip(msg = 'test not relevant for Redis')
```
gtopper: So maybe we can have separate fixtures?

integration/test_flow_integration.py
```diff
@@ -58,13 +58,35 @@ async def get_shard_data(self, path):
...
+
+    table_path = setup_kv_teardown_test["table_name"]
     controller = build_flow([
         SyncEmitSource(),
         Map(lambda x: x + 1),
         Filter(lambda x: x < 8),
-        JoinWithV3IOTable(V3ioDriver(), lambda x: x, lambda x, y: y['age'], table_path),
+        JoinWithV3IOTable(get_driver(setup_kv_teardown_test), lambda x: x, lambda x, y: y['age'], table_path),
```
gtopper: Seems like an unnecessary change.

integration/test_flow_integration.py
```diff
@@ -320,7 +343,7 @@ def test_join_by_key(setup_kv_teardown_test):
 
 
 def test_join_by_key_specific_attributes(setup_kv_teardown_test):
-    table = Table(setup_kv_teardown_test, V3ioDriver())
+    table = Table(setup_kv_teardown_test["table_name"], get_driver(setup_kv_teardown_test))
```
gtopper, line 346: I think it would be nicer to have `setup_kv_teardown_test` return a tuple?
As in
```suggestion
    table_name, _ = setup_kv_teardown_test
```

gtopper: We use single quotes in storey, except for `"""` and when single quotes need to appear in the actual string.
```suggestion
    table = Table(setup_kv_teardown_test['table_name'], get_driver(setup_kv_teardown_test))
```

integration/test_redis_specific.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture()
+def redis():
+    import os
+    import fakeredis
+    import redis as r
+    REDIS_URL = os.environ.get('REDIS_URL')
+    if REDIS_URL:
+        yield r.Redis.from_url(REDIS_URL)
```
gtopper:
```suggestion
    redis_url = os.environ.get('REDIS_URL')
    if REDIS_URL:
        yield r.Redis.from_url(redis_url)
```

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
+    return f'{prefix}/{random_table}/'
 
+redis_server = None
 
-@pytest.fixture()
-def setup_teardown_test():
+def get_redis_client(redis_fake_server=None):
+    REDIS_URL = os.environ.get('REDIS_URL')
+    if REDIS_URL:
+        return r.Redis.from_url(REDIS_URL)
```
gtopper:
```suggestion
    redis_url = os.environ.get('REDIS_URL')
    if redis_url:
        return r.Redis.from_url(redis_url)
```

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
+    return RedisDriver(redis_client = get_redis_client(redis_fake_server),key_prefix="storey-test:", *args, **kwargs)
+
+def get_driver(setup_teardown_test, *args, **kwargs):
+    if setup_teardown_test["driver_name"] == "V3ioDriver":
+        return V3ioDriver(*args, **kwargs)
+    elif setup_teardown_test["driver_name"] == "RedisDriver":
+        redis_fake_server = setup_teardown_test["redis_fake_server"] if "redis_fake_server" in setup_teardown_test else None
+        return redis_driver(redis_fake_server = redis_fake_server, *args, **kwargs)
+    else:
+        assert 0
```
gtopper:
```suggestion
        driver_name = setup_teardown_test["driver_name"]
        ...
        raise ValueError(f'Unsupported driver name "{driver_name}"')
```

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
     table_name = _generate_table_name()
+    
+    test_params = {}
+    driver_name = request.param
+    test_params["driver_name"] = driver_name
+    test_params["table_name"] = table_name
+    if driver_name == "RedisDriver":
+        REDIS_URL = os.environ.get('REDIS_URL')
+        if not REDIS_URL:
+            test_params["redis_fake_server"] = fakeredis.FakeServer()
```
gtopper: I'm starting to think this code is repeated a lot... Can we extract it?
```suggestion
        redis_url = os.environ.get('REDIS_URL')
        if not redis_url:
            test_params["redis_fake_server"] = fakeredis.FakeServer()
```

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
+    table_name = _generate_table_name()
+    
+    test_params = {}
+    driver_name = request.param
+    test_params["driver_name"] = driver_name
+    test_params["table_name"] = table_name
+    if driver_name == "RedisDriver":
+        REDIS_URL = os.environ.get('REDIS_URL')
+        if not REDIS_URL:
+            test_params["redis_fake_server"] = fakeredis.FakeServer()
```
gtopper: Here as well.

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
+    yield test_params
 
     # Teardown
-    asyncio.run(recursive_delete(table_path, V3ioHeaders()))
+    if driver_name == "V3ioDriver":
+        asyncio.run(recursive_delete(table_name, V3ioHeaders()))
+    elif driver_name == "RedisDriver":
+        remove_redis_table(table_name)
+    else:
+        assert 0
```
gtopper: Here too we need to raise an error with what happened as mentioned above. Not just `assert False` (let alone 0...).

integration/integration_test_utils.py
```diff
@@ -68,40 +73,101 @@ def __init__(self, **kwargs):
...
+drivers_list = ["V3ioDriver", "RedisDriver"]
+@pytest.fixture(params=drivers_list)
+def setup_teardown_test(request):
     # Setup
     table_name = _generate_table_name()
+    
+    test_params = {}
+    driver_name = request.param
+    test_params["driver_name"] = driver_name
+    test_params["table_name"] = table_name
```
gtopper: I'm starting to think this way makes for too much clutter in test code... If we really need this as opposed to just a tuple result, can we create a class for it? It makes for more readable (and writable) client code.

storey/drivers.py
```diff
@@ -156,9 +156,11 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
         response = await self._v3io_client.kv.update(container, table_path, key, expression=update_expression,
                                                      condition=condition_expression,
                                                      raise_for_status=v3io.aio.dataplane.RaiseForStatus.never)
+
```
gtopper: Let's not change the formatting in this file.

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+
+class RedisType(Enum):
+    STANDALONE = 1
+    CLUSTER = 2
+
+
+class NeedsRedisAccess:
+    """Checks that params for access to Redis exist and are legal
+
+    :param webapi: URL to the web API (https or http). If not set, the REDIS_URL environment variable will be used.
```
gtopper: I don't think this comment is correct for redis. Redis URLs have the "redis" scheme rather than "http" or "https", and are not referred to as "web API". So this seems like accidental c&p from v3io.

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+
```
gtopper, line 27:
```suggestion
```

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+    """
+
+    def __init__(self, redis_url=None):
+        redis_url = redis_url or os.getenv('REDIS_URL')
+        if not redis_url:
+            self._redis_url = None
+            print('Missing redis_url parameter or REDIS_URL environment variable. Using fakeredit instead')
```
gtopper: Hm, we're using fakeredis in the implementation code? Isn't it only for tests?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        self._aggregation_time_attribute_prefix = aggregation_time_attribute_prefix
+        self._aggregation_prefixes = (self._aggregation_attribute_prefix,
+                                      self._aggregation_time_attribute_prefix)
+
+    def asyncify(self, fn):
+        """Run a synchronous function asynchronously."""
+        async def inner_fn(*args, **kwargs):
+            loop = asyncio.get_event_loop()
+            partial_hset = partial(fn, *args, **kwargs)
+            return await loop.run_in_executor(RedisDriver._thread_pool, partial_hset)
```
gtopper: Elsewhere in storey we pass `None` for the executor (thread pool is the implementation type) to get the default executor. Why do we need a dedicated parameterless executor here?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        self._key_prefix = key_prefix if key_prefix else self.DEFAULT_KEY_PREFIX
+        self._type = redis_type
+        self._mtime_name = '$_mtime_'
+
+        self._aggregation_attribute_prefix = aggregation_attribute_prefix
+        self._aggregation_time_attribute_prefix = aggregation_time_attribute_prefix
+        self._aggregation_prefixes = (self._aggregation_attribute_prefix,
+                                      self._aggregation_time_attribute_prefix)
+
+    def asyncify(self, fn):
```
gtopper: Method can be moved out of class or be made static.

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        """Run a synchronous function asynchronously."""
+        async def inner_fn(*args, **kwargs):
+            loop = asyncio.get_event_loop()
+            partial_hset = partial(fn, *args, **kwargs)
```
gtopper: Seems to me that `partial_hset` is a leftover from when this function was specific to the HSET operation.
```suggestion
            partial_func = partial(fn, *args, **kwargs)
```

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        if self._redis:
+            return self._redis
+        if self._type is RedisType.STANDALONE:
+            return redis.Redis.from_url(self._redis_url, decode_responses=True)
+        self._redis = rediscluster.RedisCluster.from_url(self._redis_url, decode_response=True)
+        return self._redis
+
+    @staticmethod
+    def make_key(key_prefix, *parts):
+        return f"{key_prefix}{''.join([str(p) for p in parts])}"
```
gtopper, line 99:
```suggestion
        key = ''.join([str(p) for p in parts])
        return f'{key_prefix}{key}'
```
but hm, is it even correct to join the keys without a separator?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+            if pd.isnull(value):
+                return None
+            return f'{cls.DATETIME_FIELD_PREFIX}{value.timestamp()}'
+        elif isinstance(value, timedelta):
+            return f'{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}'
+        else:
+            if isinstance(value, float):
+                if value == math.inf or value == -math.inf:
+                    # Use the shorter infinity form (inf, -inf) to save space.
+                    value = str(value)
```
gtopper:
```
>>> json.dumps(math.inf)
'Infinity'
>>> json.dumps(str(math.inf))
'"inf"'
```
The comment implies that `value = str(value)` is an optimization, but the code doesn't work without it. So need to remove or improve the comment.

What happens if the value is NaN?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        elif isinstance(value, timedelta):
+            return f'"{cls.TIMEDELTA_FIELD_PREFIX}{value.total_seconds()}"'
+        elif isinstance(value, bool):
+            return f'"{json.dumps(value)}"'
+        else:
+            if isinstance(value, float):
+                if value == math.inf:
+                    value = "math.huge"
+                elif value == -math.inf:
+                    value = "-math.huge"
```
gtopper, line 167: Here too, are we sure NaN works?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+
+        value = str_value.lower()
+        if value.startswith(cls.TIMEDELTA_FIELD_PREFIX):
+            ret = timedelta(seconds=float(value.split(":")[1]))
+        elif value.startswith(cls.DATETIME_FIELD_PREFIX):
+            ret = datetime.fromtimestamp(float(value.split(":")[1]), tz=timezone.utc)
+        elif value == "inf":
+            ret = math.inf
+        elif value == "-inf":
+            ret = -math.inf
```
gtopper, line 191: Here too, should we be concerned about NaN?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        elif value == "true":
+            ret = True
+        elif value == "false":
+            ret = False
+        else:
+            # if value is a number, convert type to int / float
+            if value[0] in ('-', '+'):
+                value = value[1:]
+            if value.replace('.', '', 1).isdigit():
+                ret = int(value) if int(float(value)) == float(value) else float(value)
```
gtopper: Am I missing something, or we just ignore the sign here?

tests/test_flow.py
```diff
@@ -23,6 +23,7 @@
     ParquetTarget, QueryByKey, \
     TSDBTarget, Extend, SendToHttp, HttpRequest, NoSqlTarget, NoopDriver, Driver, Recover, V3ioDriver, ParquetSource
 from storey.flow import _ConcurrentJobExecution, Context, ReifyMetadata, Rename
+from integration.integration_test_utils import append_return
```
gtopper: I'm pretty sure there's no need to change this file.

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+
+        return lua_script, condition_expression, pending_updates, redis_keys_involved
+
+    async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data):
+        redis_key_prefix = self._make_key(container, table_path, key)
+        update_expression, mtime_condition, pending_updates, redis_keys_involved = self._build_feature_store_lua_update_script(
+            redis_key_prefix, aggr_item, partitioned_by_key, additional_data)
+        if not update_expression:
+            return
+        current_time = int(time.time_ns() / 1000)
```
gtopper, line 321: Microseconds on purpose?

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        if mtime_condition is not None:
+            update_expression = f'if redis.call("HGET", "{redis_key_prefix}","{self._mtime_name}") == "{mtime_condition}" then \
+                {update_expression} redis.call("HSET","{redis_key_prefix}","{self._mtime_name}",{current_time});return 1;else return 0;end;'
+        else:
+            update_expression = f'{update_expression}redis.call("HSET","{redis_key_prefix}","{self._mtime_name}",{current_time});return 1;'
+
+        redis_keys_involved.append(redis_key_prefix)
+        update_expression = f"""{update_expression}"""
```
gtopper: This line seems redundant.

storey/redis_driver.py
```diff
@@ -0,0 +1,482 @@
...
+        update_ok = await self.asyncify(self.redis.eval)(update_expression, len(redis_keys_involved), *redis_keys_involved)
+        # should_raise_error = False
```
gtopper: Dead code ☠️

storey/table.py
assaf758: fixed

storey/table.py
assaf758: fixed

integration/test_flow_integration.py
assaf758: fixed

integration/test_flow_integration.py
assaf758: fixed

integration/test_flow_integration.py
assaf758: Since we are thinking about obsoleting/depreciating JoinWithV3IOTable, do we really want to a fixture that "escapes" the driver parametrization?

integration/test_aggregation_integration.py
assaf758: addressing this is still in progress.

integration/test_flow_integration.py
assaf758: fixed

integration/test_redis_specific.py
assaf758: fixed

integration/integration_test_utils.py
assaf758: fixed

integration/integration_test_utils.py
assaf758: done

integration/integration_test_utils.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/redis_driver.py
assaf758: fixed

storey/redis_driver.py
assaf758: made static

storey/redis_driver.py
assaf758: changed to partial_fn

storey/redis_driver.py
assaf758: it is - the table_path part contains '/' both at it's start and end.

storey/redis_driver.py
assaf758: it's an optimization. If you want to remove it you have to do the on the function that translate back to python.
Aded NaN handling - good catch.

storey/redis_driver.py
assaf758: good catch - fixed

tests/test_flow.py
assaf758: fixed

storey/redis_driver.py
assaf758: occurs twice in the code so I assume yes :-)
what is the concern? suggest keeping the mtime at nano resolution?

storey/redis_driver.py
assaf758: removed

storey/redis_driver.py
assaf758: removed

storey/redis_driver.py
assaf758: will be addressed in next commit

storey/redis_driver.py
assaf758: will be addressed in next commit

integration/integration_test_utils.py
assaf758: done

integration/integration_test_utils.py
assaf758: class created

integration/integration_test_utils.py
assaf758: fixed

storey/redis_driver.py
assaf758: fixed

storey/redis_driver.py
assaf758: it is, fixed logic.

integration/test_flow_integration.py
assaf758: removed by introducing TestContext getters.

storey/redis_driver.py
assaf758: since the redis driver 3.5 does not support asyncio, using multiple threads provided a mitigation.
I now moved to redis 4.3 which does support asyncio. I will evaluate moving to it later on (not in this PR).

storey/redis_driver.py
assaf758: fixed

storey/redis_driver.py
assaf758: fixed

storey/redis_driver.py
gtopper: Not really concerned about it, just wanted to make sure it's a deliberate choice :)

storey/redis_driver.py
gtopper: This doesn't answer my question though... Why not use the default executor as is done elsewhere in storey?

storey/drivers.py
```diff
@@ -244,7 +246,6 @@ def _build_feature_store_update_expression(self, aggregation_element, additional
                     expressions.append(f'{name}={self._convert_python_obj_to_expression_value(value)}')
                 else:
                     expressions.append(f'REMOVE {name}')
-
```
gtopper: No need to change formatting unnecessarily.

storey/drivers.py
```diff
@@ -168,10 +169,12 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
             response = await self._v3io_client.kv.update(container, table_path, key, expression=update_expression,
                                                          condition=condition_expression,
                                                          raise_for_status=v3io.aio.dataplane.RaiseForStatus.never)
+
```
gtopper:
```suggestion
```

storey/drivers.py
```diff
@@ -168,10 +169,12 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
...
             if response.status_code == 200 and aggr_item:
                 await self._fetch_state_by_key(aggr_item, container, table_path, key)
             else:
                 should_raise_error = True
+
```
gtopper:
```suggestion
```

storey/drivers.py
```diff
@@ -159,6 +159,7 @@ async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_
         if response.status_code == 200:
             if aggr_item:
                 aggr_item.storage_specific_cache[self._mtime_header_name] = response.headers[self._mtime_header_name]
+
```
gtopper:
```suggestion
```

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/drivers.py
assaf758: fixed

storey/redis_driver.py
assaf758: we can do that - but the code will run slower.
the reason for having multiple threads is to hide the latency of the call to Redis: the other thread will be blocked waiting for the redis call to complete while the main thread can handle other coroutines.

integration/test_aggregation_integration.py
assaf758: fixed

gtopper (approved):

PR #369, opened 2022-08-08
None aggregation option 

davesh0812:
Adding option to use table without full support for aggregation step in the query process. This change is supporting 2 kind of future drivers for now - mongoDB and SQL.


storey/aggregations.py
```diff
@@ -251,9 +251,17 @@ def __init__(self, features: List[str], table: Union[Table, str], key: Union[str
                  fixed_window_type: Optional[FixedWindowType] = FixedWindowType.CurrentOpenWindow, **kwargs):
         self._aggrs = []
         self._enrich_cols = []
+        self._table = None
         resolved_aggrs = {}
+        Flow.__init__(self, **kwargs)
```
gtopper: `AggregateByKey.__init__` already calls `Flow.__init__`. So this will cause it to be called twice.

storey/aggregations.py
```diff
@@ -251,9 +251,17 @@ def __init__(self, features: List[str], table: Union[Table, str], key: Union[str
...
+        if isinstance(table, str):
+            if not self.context:
+                raise TypeError("Table can not be string if no context was provided to the step")
+            self._table = self.context.get_table(table)
```
gtopper: This code is already in the super class. Afaict, this will cause the table to be instantiated and then cloned 17 lines below, which doesn't make much sense.

storey/table.py
```diff
@@ -24,7 +24,7 @@ class Table:
      """
 
     def __init__(self, table_path: str, storage: Driver, partitioned_by_key: bool = True,
-                 flush_interval_secs: Optional[int] = 300, max_updates_in_flight: int = 8):
+                 flush_interval_secs: Optional[int] = 300, max_updates_in_flight: int = 8, support_aggr=True):
```
gtopper: What happens if `support_aggr=False` in ingest rather than on querying? Looks like this parameter only has one specific flow in mind...

gtopper (changes requested):
Please improve the title and description to reflect the motivation behind this change. It's hard for me to make refactoring suggestions when I don't understand the purpose behind the code.

What flow does this serve? When does a table not support aggregations? Is this specific to a driver? If that's the case, why not get this information from the driver? Shouldn't it also affect the ingest flow rather than only the query flow?

davesh0812:
#372 replaced this one.

PR #375, opened 2022-08-29
Support wasbs schema

speretz-ig:


storey/utils.py
```diff
@@ -113,6 +113,8 @@ def url_to_file_system(url, storage_options):
         load_fs_dependencies(scheme)
         if scheme == 'v3io':
             remaining_path = parsed_url.path
+        elif scheme in ["wasb", "wasbs"]:
+            remaining_path = f'{parsed_url.netloc.split("@")[0]}{parsed_url.path}'
```
gtopper: Can you please add a test so it's clearer what kind of URL this is for? Because it's hard to tell if this is correct or not otherwise.

speretz-ig: test added

integration/test_filesystems_integration.py
```diff
@@ -571,3 +571,10 @@ def test_empty_filter_result(setup_teardown_test):
         pd.testing.assert_frame_equal(read_back_result, pd.DataFrame({}))
     finally:
         os.remove(path)
+
+
+def test_get_path_utils():
+    url = "wasbs://mycontainer@myaccount.blob.core.windows.net/pat/to/object.csv"
```
gtopper, line 577: pat -> path

storey/utils.py
```diff
@@ -118,17 +118,25 @@ def _split_path(path):
...
     remaining_path = url
     scheme = ""
     if "://" in url:
         parsed_url = urlparse(url)
         scheme = parsed_url.scheme.lower()
-        load_fs_dependencies(scheme)
         if scheme == 'v3io':
             remaining_path = parsed_url.path
+        elif scheme in ["wasb", "wasbs"]:
+            remaining_path = f'{parsed_url.netloc.split("@")[0]}{parsed_url.path}'
```
gtopper: Better replace this with `parsed_url.username`, so you don't have to do additional ad-hoc parsing.

storey/utils.py
```diff
@@ -118,17 +118,25 @@ def _split_path(path):
...
         else:
             remaining_path = f'{parsed_url.netloc}{parsed_url.path}'
+    return scheme, remaining_path
+
+
+def url_to_file_system(url, storage_options):
+    scheme, remaining_path = get_remaining_path(url)
+    if scheme != "":
```
```suggestion
    if scheme:
```

storey/utils.py
```diff
@@ -118,17 +118,25 @@ def _split_path(path):
...
         if scheme == 'v3io':
             remaining_path = parsed_url.path
+        elif scheme in ["wasb", "wasbs"]:
+            remaining_path = f'{parsed_url.netloc.split("@")[0]}{parsed_url.path}'
         else:
             remaining_path = f'{parsed_url.netloc}{parsed_url.path}'
+    return scheme, remaining_path
+
+
+def url_to_file_system(url, storage_options):
```
gtopper, line 136: The name implies that this function takes a URL and returns a file system object, but actually it returns nothing.

integration/test_filesystems_integration.py
gtopper:
```suggestion
    url = "wasbs://mycontainer@myaccount.blob.core.windows.net/path/to/object.csv"
```

speretz-ig: Done

storey/utils.py
speretz-ig: Done

storey/utils.py
speretz-ig: Done

integration/test_filesystems_integration.py
```diff
@@ -79,7 +80,6 @@ async def _delete_file(path):
     finally:
         await client_session.close()
 
-
```
gtopper: Please don't change the formatting.

gtopper (approved):
LGTM just please don't change the formatting (see above).

PR #376, opened 2022-09-13
ML-1246: Support `full_event` in `StreamTarget` and `KafkaTarget`

gtopper:
and read it back properly.

### DIFF
--- a/integration/test_flow_integration.py
+++ b/integration/test_flow_integration.py
@@ -189,7 +189,7 @@ async def async_test_write_to_v3io_stream_full_event_readback(setup_stream_teard
 
     controller = build_flow([
         AsyncEmitSource(),
-        Reduce([], lambda acc, x: append_return(acc, x)),
+        Reduce([], lambda acc, x: append_return(acc, x), full_event=True),
     ]).run()
     for record in (shard0_data + shard1_data):
         await controller.emit(Event(json.loads(record.decode('utf8')), id='some-new-id'))
@@ -197,7 +197,12 @@ async def async_test_write_to_v3io_stream_full_event_readback(setup_stream_teard
     await controller.terminate()
     result = await controller.await_termination()
 
-    assert result == [0, 2, 4, 6, 8, 1, 3, 5, 7, 9]
+    assert len(result) == 10
+    expected_bodies = [0, 2, 4, 6, 8, 1, 3, 5, 7, 9]
+    for i, record in enumerate(result):
+        assert record.body == expected_bodies[i]
+        assert record.id == str(expected_bodies[i])
+        assert record.time == event_time
 
 
 def test_async_test_write_to_v3io_stream_full_event_readback(assign_stream_teardown_test):
--- a/integration/test_flow_integration.py
+++ b/integration/test_flow_integration.py
@@ -26,7 +26,7 @@ import v3io_frames as frames
 
 from storey import Filter, JoinWithV3IOTable, SendToHttp, Map, Reduce, SyncEmitSource, HttpRequest, build_flow, \
     StreamTarget, V3ioDriver, TSDBTarget, Table, JoinWithTable, MapWithState, NoSqlTarget, DataframeSource, \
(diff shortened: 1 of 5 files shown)


integration/test_flow_integration.py
```diff
@@ -151,14 +153,70 @@ def test_write_to_v3io_stream(assign_stream_teardown_test):
...
+async def async_test_write_to_v3io_stream_full_event_readback(setup_stream_teardown_test):
+    stream_path = setup_stream_teardown_test
+    event_time = datetime(2022, 8, 8)
+
+    controller = build_flow([
+        AsyncEmitSource(),
+        StreamTarget(V3ioDriver(), stream_path, sharding_func=lambda event: int(event.body), batch_size=8, shard_count=2, full_event=True)
+    ]).run()
+    for i in range(10):
+        await controller.emit(Event(i, time=event_time, id=str(i)))
```
assaf758, line 165: i => body=i

integration/test_flow_integration.py
```diff
@@ -151,14 +153,70 @@ def test_write_to_v3io_stream(assign_stream_teardown_test):
...
+
+    await asyncio.sleep(5)
+
+    try:
+        shard0_data = await GetShardData().get_shard_data(f'{stream_path}/0')
+        assert shard0_data == [
+            b'{"full_event_wrapper": true, "body": 0, "time": "2022-08-08T00:00:00", "id": "0"}',
```
assaf758, line 172: "full_event_wrapper" =>  Event._serialize_event_marker

storey/dtypes.py
```diff
@@ -38,6 +38,9 @@ class Event:
     :type awaitable_result: AwaitableResult (Optional)
     """
 
+    _serialize_event_marker = 'full_event_wrapper'
```
assaf758, line 41: how about adding a '_' prefix or even '_story' prefix to make conflicts with the user data less probable..

integration/test_flow_integration.py
```diff
@@ -151,14 +153,70 @@ def test_write_to_v3io_stream(assign_stream_teardown_test):
...
+    finally:
+        await controller.terminate()
+        await controller.await_termination()
+
+    controller = build_flow([
+        AsyncEmitSource(),
+        Reduce([], lambda acc, x: append_return(acc, x), full_event=True),
+    ]).run()
+    for record in (shard0_data + shard1_data):
+        await controller.emit(Event(json.loads(record.decode('utf8')), id='some-new-id'))
```
assaf758: i understand the id ('some-new-id') is ignored when receiving full-event - please state this in a comment.

assaf758 (changes requested):

integration/test_flow_integration.py
gtopper: I don't mind changing it, but it's like that pretty much everywhere:
```
✗ grep -r 'Event(' * | wc -l 
     105
✗ grep -r 'Event(body=' * | wc -l
       4
```

integration/test_flow_integration.py
gtopper: This way we get:
* a readable example that's also
* less coupled with the implementation and
* will break if we make an incompatible change

I think the last one is particularly important.

storey/dtypes.py
gtopper: But... How is the user going to conflict with this? They can't insert anything into the wrapper.

integration/test_flow_integration.py
gtopper: I'll put it in the ID itself.

storey/dtypes.py
assaf758: I guess can happen if _build_event() receive an event with a user "full_event_wrapper" field - but I agree it's unlikely.

integration/test_flow_integration.py
assaf758: ok

assaf758 (approved):

PR #381, opened 2022-10-12
ML-2505: Support sharding by field

gtopper:


storey/targets.py
```diff
@@ -23,12 +23,13 @@
...
 import traceback
 import uuid
-from typing import Optional, Union, List, Callable, Tuple
+from typing import Optional, Union, List, Callable, Tuple, Any
 from urllib.parse import urlparse
 
 import pandas as pd
 import pyarrow
 import v3io_frames as frames
+import xxhash as xxhash
```
assaf758: this "as" looks redundant, no?

storey/targets.py
```diff
@@ -843,7 +877,12 @@ async def _do(self, event):
             record = json.dumps(record).encode("UTF-8")
             partition = None
             if self._sharding_func:
```
assaf758, line 881: seems like a KafkaTarget instance will always have a _sharding_func, if true the "if" clause is redundant - ie can be done always

storey/targets.py
```diff
@@ -801,10 +830,15 @@ def __init__(self, bootstrap_servers: str, topic: str, producer_options: Optiona
...
-        self._sharding_func = sharding_func
+        self._sharding_func = None
+        if isinstance(sharding_func, int):
+            self._sharding_func = lambda _: sharding_func
+        elif isinstance(sharding_func, str):
+            self._sharding_func = lambda event: event.body.get(sharding_func, None)
+        elif callable(sharding_func):
+            self._sharding_func = sharding_func
+        else:
+            raise TypeError(f'Expected an int, string, or callable, got {type(sharding_func)}')
```
assaf758: maybe add the value as well:
... got X of type Y

storey/targets.py
```diff
@@ -843,7 +877,12 @@ async def _do(self, event):
             record = json.dumps(record).encode("UTF-8")
             partition = None
             if self._sharding_func:
-                partition = self._sharding_func(event)
+                sharding_func_result = self._sharding_func(event)
+                partition = None
```
assaf758: you already set the partition to None (few lines above) - seems redundant

storey/targets.py
```diff
@@ -789,9 +812,15 @@ class KafkaTarget(Flow, _Writer):
...
     """
 
-    def __init__(self, bootstrap_servers: str, topic: str, producer_options: Optional[dict] = None,
-                 sharding_func: Optional[Callable[[Event], int]] = None, columns: Optional[List[str]] = None,
-                 infer_columns_from_data: Optional[bool] = None, full_event: Optional[bool] = None, **kwargs):
+    def __init__(self,
+                 bootstrap_servers: str,
+                 topic: str,
+                 producer_options: Optional[dict] = None,
+                 sharding_func: Union[None, int, str, Callable[[Event], Any]] = None,
```
assaf758, line 821: instead of Any, can be Union[int,str]

storey/targets.py
```diff
@@ -625,16 +627,23 @@ class StreamTarget(Flow, _Writer):
...
         if using a URL that will be parsed by fsspec, e.g., starting “s3://”, “gcs://”. Optional
     :type storage_options: dict
     """
 
-    def __init__(self, storage: Driver, stream_path: str, sharding_func: Optional[Callable[[Event], int]] = None, batch_size: int = 8,
-                 columns: Optional[List[str]] = None, infer_columns_from_data: Optional[bool] = None,
-                 shard_count: int = 1, retention_period_hours: int = 24, full_event: Optional[bool] = None, **kwargs):
+    def __init__(self,
+                 storage: Driver, stream_path: str,
+                 sharding_func: Union[None, int, str, Callable[[Event], Any]] = None,
```
assaf758, line 639: instead of Any, can be Union[int,str]

requirements.txt
```diff
@@ -13,3 +13,4 @@ v3io-frames>=0.10.2,<1
 # which breaks v3io-frames
 protobuf>=3.20.1,!=3.20.2,<4
 v3iofs~=0.1.9
+xxhash>=1
```
assaf758, line 16: i noticed that we do "from kafka import" but in mlrun we install kafka-python (and indeed it did not work for me with pip install kafka). so we should probably change to import kafka-python..

storey/targets.py
```diff
@@ -646,15 +655,21 @@ def __init__(self, storage: Driver, stream_path: str, sharding_func: Optional[Ca
...
-        self._sharding_func = sharding_func
+        self._sharding_func = None
+        if isinstance(sharding_func, int):
+            self._sharding_func = lambda _: sharding_func
+        elif isinstance(sharding_func, str):
+            self._sharding_func = lambda event: event.body.get(sharding_func, None)
+        elif callable(sharding_func):
+            self._sharding_func = sharding_func
+        else:
+            raise TypeError(f'Expected an int, string, or callable, got {type(sharding_func)}')
```
assaf758: maybe add the value as well:
... got X of type Y

storey/targets.py
```diff
@@ -708,13 +723,18 @@ async def _worker(self):
...
                             await self._handle_response(req)
-                        for shard_id in range(self._shard_count):
+                        for shard_id in range(self._shards):
                             if buffers[shard_id]:
                                 self._send_batch(buffers, in_flight_reqs, shard_id)
                         for req in in_flight_reqs:
                             await self._handle_response(req)
                         break
-                    shard_id = self._sharding_func(event) % self._shard_count
+                    shard_id = self._sharding_func(event)
```
assaf758: i prefer the kafka name you used (sharding_func_result) - since it's not (always) the shard_id at this point, and will be less confusing when we hash the shard_id down below

integration/test_flow_integration.py
```diff
@@ -318,6 +318,38 @@ def test_write_to_v3io_stream_unbalanced(assign_stream_teardown_test):
     assert shard1_data == []
 
 
+@pytest.mark.parametrize('sharding_func_type', ['func', 'partition_num', 'field'])
+def test_write_to_v3io_stream_unbalanced(assign_stream_teardown_test, sharding_func_type):
+    if sharding_func_type == 'func':
+        sharding_func = lambda event: 0
```
assaf758, line 324: seems the coverage of the case where the function return the partition key is missing

assaf758 (changes requested):
nice!
some comments inline.

storey/targets.py
gtopper: It shouldn't have to. Fixing the code in the constructor.

storey/targets.py
gtopper: It doesn't need to be an int or a string though.

storey/targets.py
gtopper: It doesn't need to be an int or a string though.

requirements.txt
gtopper: There is no `import kafka-python`. kafka-python is the name of library, kafka is the name of the main package in that library.

integration/test_flow_integration.py
gtopper: What do you mean? This line *is* the coverage.

storey/targets.py
assaf758: ok. so probably need to add similar "if" clause for the StreamTarget

storey/targets.py
assaf758: if i understand correct the code (and the class documentation) the function maps from event to partition (int) or sharding-key (str).. So how can it be other types?

storey/targets.py
gtopper: It already has one.

storey/targets.py
gtopper: Sharding key can be any hashable type basically, as defined by xxhash for `StreamTarget` and by the kafka library's serializer for `KafkaTarget`.

assaf758 (approved):

PR #384, opened 2022-10-18
SqlDB - sqldb driver + source

davesh0812:
This pr focus in implementation of source, driver for SqlDB.
SqlDriver is not fully supported in the aggregation query flow.

https://jira.iguazeng.com/browse/ML-2610


integration/integration_test_utils.py
```diff
@@ -111,6 +113,19 @@ def remove_redis_table(table_name):
...
+def remove_sql_tables():
+    import sqlalchemy as db
+    engine = db.create_engine(SQL_DB)
+    with engine.connect() as conn:
+        metadata = db.MetaData()
+        metadata.reflect(bind=engine)
+        # and drop them, if they exist
+        metadata.drop_all(bind=engine, checkfirst=True)
+        engine.dispose()
+        conn.close()
```
gtopper: Doesn't the `with` already close the connection?

integration/integration_test_utils.py
```diff
@@ -111,6 +113,19 @@ def remove_redis_table(table_name):
         redis_client.delete(key)
         count += 1
 
+
+def remove_sql_tables():
```
gtopper, line 118: This method is not in use.

integration/integration_test_utils.py
```diff
@@ -169,21 +199,26 @@ def setup_teardown_test(request):
         asyncio.run(recursive_delete(test_context.table_name, V3ioHeaders()))
     elif test_context.driver_name == "RedisDriver":
         remove_redis_table(test_context.table_name)
+    elif test_context.driver_name == "SQLDriver":
+        # remove_sql_tables()
```
gtopper: Dead code ☠️

requirements.txt
```diff
@@ -13,4 +13,5 @@ v3io-frames>=0.10.2,<1
 # which breaks v3io-frames
 protobuf>=3.20.1,!=3.20.2,<4
 v3iofs~=0.1.9
-xxhash>=1
+sqlalchemy~=1.3
```
gtopper: I think this should probably be an optional requirement. In storey, these can be found in `setup.py`.

integration/test_flow_integration.py
```diff
@@ -80,10 +81,40 @@ def _get_redis_kv_all_attrs(setup_teardown_test: TestContext, key: str):
     values = get_redis_client(redis_fake_server=redis_fake_server).hgetall(redis_key)
     return {RedisDriver.convert_to_str(key): RedisDriver.convert_redis_value_to_python_obj(val) for key, val in values.items()}
 
+def _get_sql_by_key_all_attrs(setup_teardown_test: TestContext, key: str, key_name):
```
gtopper: Not sure why `setup_teardown_test` and `key` have type annotations and `key_name` doesn't, but best be consistent if there isn't some logic to justify it.

integration/test_flow_integration.py
```diff
@@ -80,10 +81,40 @@ def _get_redis_kv_all_attrs(setup_teardown_test: TestContext, key: str):
...
+    import sqlalchemy as db
+    where_statement = ""
+    if isinstance(key, str) and '.' in key:
+        key = key.split('.')
+    if isinstance(key, List):
```
```suggestion
    if isinstance(key, list):
```

integration/test_flow_integration.py
```diff
@@ -80,10 +81,40 @@ def _get_redis_kv_all_attrs(setup_teardown_test: TestContext, key: str):
     values = get_redis_client(redis_fake_server=redis_fake_server).hgetall(redis_key)
     return {RedisDriver.convert_to_str(key): RedisDriver.convert_redis_value_to_python_obj(val) for key, val in values.items()}
 
+def _get_sql_by_key_all_attrs(setup_teardown_test: TestContext, key: str, key_name):
+    import sqlalchemy as db
+    where_statement = ""
+    if isinstance(key, str) and '.' in key:
```
gtopper: The type annotation says this is always a string...

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        self._aggregation_time_attribute_prefix = aggregation_time_attribute_prefix
+
+    def _lazy_init(self):
+        import sqlalchemy as db
+
+        if not self._sql_connection:
+            self._engine = db.create_engine(self._db_path)
+            self._sql_connection = self._engine.connect()
+
+    def table(self, table_path):
```
gtopper:
```suggestion
    def _table(self, table_path):
```

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        import sqlalchemy as db
+
+        metadata = db.MetaData()
+        return db.Table(
+            table_path[3:].split("/")[1],
```
gtopper: What's up with the `[3:]`? In a debugger I see that the table path is `'///test.db/JbNLxlQwvC'`. How did we end up with 3 slashes? This looks error prone.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            metadata,
+            autoload=True,
+            autoload_with=self._engine,
+        )
+
+    async def _save_schema(self, container, table_path, schema):
+        self._lazy_init()
+        return None
```
gtopper:
```suggestion
```

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+    async def _load_schema(self, container, table_path):
+        self._lazy_init()
+        return None
```
gtopper:
```suggestion
```

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+        self._lazy_init()
+
+        table = self.table(table_path)
+        return_val = None
+        try:
+            return_val = self._sql_connection.execute(table.insert(), [additional_data], autocommit=True)
+
+        except db.exc.IntegrityError:
+            pass
```
gtopper: Needs a comment to explain why this is ignored.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        await self.close()
+        return return_val
```
gtopper: Why return anything? `_save_key` isn't supposed to return anything, and the return value is never used.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        table = self.table(table_path)
+        try:
+            agg_val, values = await self._get_all_fields(key, table)
+            if not agg_val:
+                agg_val = None
+            if not values:
+                values = None
+            return [agg_val, values]
+        except Exception:
+            return [None, None]
```
gtopper: What are we ignoring here? Shouldn't errors be raised rather than silenced?

storey/sources.py
```diff
@@ -903,3 +903,111 @@ def _init(self):
             else:
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
             self._dfs.append(df)
+
+
+class SQLSource(_IterableSource, WithUUID):
+    """Use dql table as input source for a flow.
+
```
```suggestion
    """Use SQL table as input source for a flow.
```

storey/sources.py
```diff
@@ -903,3 +903,111 @@ def _init(self):
...
+    :parameter key_field: the primary key of the table.
+    :parameter time_field: column to be used as time for events.
+    :parameter id_field: column to be used as ID for events.
+    :parameter db_name: url string connection to sql database.
+    :parameter table_name: the name of the table to access,
+                                from the current database
+    :parameter start_filter: If not None, the results will be filtered by partitions and 'filter_column' > start_filter.
+                            Default is None (not in use)
+    :parameter end_filter:  If not None, the results will be filtered by partitions 'filter_column' <= end_filter.
```
gtopper: These parameters are not used.

storey/sources.py
```diff
@@ -903,3 +903,111 @@ def _init(self):
...
+
+    """
+
+    def __init__(
+        self,
+        db_path: str = None,
+        table_name: str = None,
+        key_field: Optional[Union[str, List[str]]] = None,
+        start_filter: Optional[datetime] = None,
+        end_filter: Optional[datetime] = None,
```
gtopper: These parameters are not used.

storey/sources.py
```diff
@@ -903,3 +903,111 @@ def _init(self):
...
+
+    async def _run_loop(self):
+        import sqlalchemy as db
+
+        engine = db.create_engine(self.db_path)
+        metadata = db.MetaData()
+        connection = engine.connect()
+        table = db.Table(self.table_name, metadata, autoload=True, autoload_with=engine)
+        results = connection.execute(db.select([table])).fetchall()
+        df = pandas.DataFrame(results)
```
gtopper: 1. Why not use the results directly? Why do we need an intermediate dataframe?
2. If we did need it, we shouldn't duplicate all the code for ingesting a dataframe (below).

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+        primary_key: str,
+        db_path: str,
+        aggregation_attribute_prefix: str = "aggr_",
+        aggregation_time_attribute_prefix: str = "_",
+    ):
+        self._db_path = db_path
+        self._sql_connection = None
+        self._primary_key = primary_key
+        self._mtime_name = "$_mtime_"
+        self._storey_key = "storey_key"
```
gtopper: These are never used.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+    :param db_path: url string connection to sql database.
+    :param primary_key: the primary key of the table.
+    """
+
+    def __init__(
+        self,
+        primary_key: str,
+        db_path: str,
+        aggregation_attribute_prefix: str = "aggr_",
+        aggregation_time_attribute_prefix: str = "_",
```
gtopper: I thought aggregations were not supported.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+            my_query = f"SELECT * FROM {table} where {where_statement}"
+            results = self._sql_connection.execute(my_query).fetchall()
+        except Exception as e:
+            raise RuntimeError(f"Failed to get key {key}. Response error was: {e}")
+
+        return None, {
+            results[0]._fields[i]: results[0][i] for i in range(len(results[0]))
+        }
+
+    async def _get_specific_fields(self, key: str, table, attributes: List[str]):
```
gtopper: This method is never used.

storey/drivers.py
```diff
@@ -518,3 +518,137 @@ async def _get_item(self, container, table_path, key, attributes):
...
+
+        return None, {
+            results[0]._fields[i]: results[0][i] for i in range(len(results[0]))
+        }
+
+    def supports_aggregations(self):
+        return False
+
+    def _get_where_statement(self, key):
+        where_statement = ""
```
gtopper:
```suggestion
    def _get_where_clause(self, key):
        where_clause = ""
```

integration/integration_test_utils.py
```diff
@@ -317,3 +352,32 @@ def _convert_nginx_to_python_type(typ, value):
...
+                col_type = db.Boolean
+            elif col_type == float:
+                col_type = db.Float
+            else:
+                raise TypeError(f"{col_type} unsupported type")
+            columns.append(db.Column(col, col_type, primary_key=(col in key)))
+
+        db.Table(table_name, metadata, *columns)
+        metadata.create_all(engine)
+        conn.close()
```
gtopper: Doesn't the `with` already close the connection?

gtopper (changes requested):

storey/drivers.py
```diff
@@ -660,3 +660,121 @@ async def _get_item(self, container, table_path, key, attributes):
             attribute_names=attributes,
             raise_for_status=v3io.aio.dataplane.RaiseForStatus.never,
         )
+
+
+class SQLDriver(Driver):
```
assaf758: Please introduce sql_driver.py (similar to redis_driver.py)

assaf758 (changes requested):
I see that only test_flow_integration.py is not skipped, and also in that test file there are a lot of skips.
is there a design/spec of this driver - so we can understand what it's limitations?
Also comments inline.

integration/conftest.py
```diff
@@ -41,6 +44,9 @@ def setup_teardown_test(request):
         asyncio.run(recursive_delete(test_context.table_name, V3ioHeaders()))
     elif test_context.driver_name == "RedisDriver":
         remove_redis_table(test_context.table_name)
+    elif test_context.driver_name == "SQLDriver":
+        remove_sql_tables()
+        pass
```
gtopper: This line has no effect.

integration/integration_test_utils.py
```diff
@@ -143,7 +166,7 @@ class AggregationlessRedisDriver(RedisDriver):
         def supports_aggregations(self):
             return False
 
-    def driver(self, IsAggregationlessDriver=False, *args, **kwargs):
+    def driver(self, IsAggregationlessDriver=False, primary_key=None, *args, **kwargs):
```
gtopper: I know this was there before, but
```suggestion
    def driver(self, *args, IsAggregationlessDriver=False, primary_key=None, **kwargs):
```

integration/integration_test_utils.py
```diff
@@ -155,12 +178,19 @@ def driver(self, IsAggregationlessDriver=False, *args, **kwargs):
...
+        elif self.driver_name == "SQLDriver" and IsAggregationlessDriver:
+            sql_driver_class = storey.sql_driver.SQLDriver
+            return sql_driver_class(db_path=SQL_DB, primary_key=primary_key)
+        elif self.driver_name == "SQLDriver" and not IsAggregationlessDriver:
+            pytest.skip("SQLDriver doesnt support aggregation")
         else:
             driver_name = self.driver_name
-            raise ValueError(f'Unsupported driver name "{driver_name}"')
+            raise ValueError(
+                f'Unsupported driver name "{driver_name}" with IsAggregationlessDriver = {IsAggregationlessDriver}'
```
gtopper: Why add this to the error message? Why does it matter?

integration/integration_test_utils.py
```diff
@@ -155,12 +178,19 @@ def driver(self, IsAggregationlessDriver=False, *args, **kwargs):
                 key_prefix="storey-test:",
                 **kwargs,
             )
+        elif self.driver_name == "SQLDriver" and IsAggregationlessDriver:
+            sql_driver_class = storey.sql_driver.SQLDriver
+            return sql_driver_class(db_path=SQL_DB, primary_key=primary_key)
+        elif self.driver_name == "SQLDriver" and not IsAggregationlessDriver:
+            pytest.skip("SQLDriver doesnt support aggregation")
```
gtopper:
```suggestion
        elif self.driver_name == "SQLDriver":
            if is_aggregationless_driver:
                sql_driver_class = storey.sql_driver.SQLDriver
                return sql_driver_class(db_path=SQL_DB, primary_key=primary_key)
            else:
                pytest.skip("SQLDriver does not support aggregation")
```

integration/integration_test_utils.py
gtopper: Also know this also precedes this PR, but `IsAggregationlessDriver` is not a proper parameter name in Python. Should be `is_aggregationless_driver`.

integration/integration_test_utils.py
```diff
@@ -294,3 +324,32 @@ def _convert_nginx_to_python_type(typ, value):
         return datetime.utcfromtimestamp(secs + nanosecs / 1000000000)
     else:
         raise V3ioError(f"Type {typ} in get item response is not supported")
+
+
+def create_sql_table(schema, table_name, sql_db_path, key):
+    import datetime
+
+    import pandas as pd
+    import sqlalchemy as db
```
gtopper: These should be imported at the top of the file.
```suggestion
    import sqlalchemy as db
```

integration/integration_test_utils.py
```diff
@@ -294,3 +324,32 @@ def _convert_nginx_to_python_type(typ, value):
...
+
+    engine = db.create_engine(sql_db_path, echo=True)
+    with engine.connect() as _:
```
gtopper: Looks to me like this line is completely unnecessary.
```suggestion
```

integration/integration_test_utils.py
```diff
@@ -294,3 +324,32 @@ def _convert_nginx_to_python_type(typ, value):
...
+            elif col_type == str:
+                col_type = db.String
+            elif col_type == datetime.datetime or col_type == pd.Timestamp or col_type == pd.Timedelta:
+                col_type = db.DateTime
+            elif col_type == bool:
+                col_type = db.Boolean
+            elif col_type == float:
+                col_type = db.Float
+            else:
+                raise TypeError(f"{col_type} unsupported type")
```
gtopper:
```suggestion
                raise TypeError(f"Unsupported type '{col_type}'")
```

integration/test_flow_integration.py
```diff
@@ -115,9 +117,44 @@ def _get_redis_kv_all_attrs(setup_teardown_test: TestContext, key: str):
...
+    where_statement = ""
+    if isinstance(key, str) and "." in key:
+        key = key.split(".")
+    if isinstance(key, list):
+        for i in range(len(key_name)):
+            if i != 0:
+                where_statement += " and "
+            where_statement += f'{key_name[i]}="{key[i]}"'
+    else:
+        where_statement += f'{key_name}="{key}"'
```
gtopper, line 134:
```suggestion
    if isinstance(key, str):
        key = key.split(".")
    if isinstance(key, list):
        for i in range(len(key_name)):
            if i != 0:
                where_statement += " and "
            where_statement += f'{key_name[i]}="{key[i]}"'
```

integration/test_flow_integration.py
```diff
@@ -115,9 +117,44 @@ def _get_redis_kv_all_attrs(setup_teardown_test: TestContext, key: str):
...
+        table_name,
+        metadata,
+        autoload=True,
+        autoload_with=engine,
+    )
+    my_query = f"SELECT * FROM {sql_table} where ({where_statement})"
+    with engine.connect() as conn:
+        results = conn.execute(my_query).fetchall()
+    values = {results[0]._fields[i]: results[0][i] for i in range(len(results[0]))}
+    return values
```
gtopper, line 148:
```suggestion
    return results[0]._mapping
```

requirements.txt
```diff
@@ -14,4 +14,4 @@ v3io-frames>=0.10.2,<1
 # which breaks v3io-frames
 protobuf>=3.20.1,!=3.20.2,<4
 v3iofs~=0.1.9
-xxhash>=1
+xxhash>=1
```
gtopper, line 17: Please don't change this line.

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+                else:
+                    key = body[self._key_field]
+                    if key is None:
+                        create_event = False
+            if create_event:
+                time = None
+                if self._time_field:
+                    time = self.get_val_from_multi_dictionary(body, self._time_field)
+                if self._id_field:
+                    _id = self.get_val_from_multi_dictionary(body, self._id_field)
```
gtopper: BUG. These two ifs are not covered by any tests. It they were, the test would fail because this code, instead of accessing field "my_id", will try to access fields "m", "y", "_", "i", and "d".

gtopper: Afaict, this code was copied from `DataframeSource` and then the bug was added. Please try to avoid duplicating this code.

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2020 Iguazio
```
gtopper:
```suggestion
# Copyright 2022 Iguazio
```

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+from typing import List
+
+import sqlalchemy as db
+
+from storey.drivers import Driver
+
+
+class SQLDriver(Driver):
+    """
+    Database connection to Sql data basw.
```
gtopper:
```suggestion
    SQL database connector.
```

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+    :param db_path: url string connection to sql database.
```
gtopper:
```suggestion
    :param db_url: database url
```

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+
+    async def _get_all_fields(self, key, table):
+        where_clause = self._get_where_clause(key)
+        try:
+            my_query = f"SELECT * FROM {table} where {where_clause}"
+            results = self._sql_connection.execute(my_query).fetchall()
+        except Exception as e:
+            raise RuntimeError(f"Failed to get key {key}. Response error was: {e}")
+
+        return None, {results[0]._fields[i]: results[0][i] for i in range(len(results[0]))}
```
gtopper:
```suggestion
        return None, results[0]._mapping
```
Copy `results[0]._mapping` if necessary (I did not look into this).

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+
+    async def _get_specific_fields(self, key: str, table, attributes: List[str]):
+        where_clause = self._get_where_clause(key)
+        try:
+            my_query = f"SELECT {','.join(attributes)} FROM {table} where {where_clause}"
+            results = self._sql_connection.execute(my_query).fetchall()
+        except Exception as e:
+            raise RuntimeError(f"Failed to get key {key}. Response error was: {e}")
+
+        return None, {results[0]._fields[i]: results[0][i] for i in range(len(results[0]))}
```
gtopper:
```suggestion
        return None, results[0]._mapping
```
Copy `results[0]._mapping` if necessary (I did not look into this).

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+        where_clause = ""
+        if isinstance(key, str) and "." in key:
+            key = key.split(".")
+        if isinstance(key, List):
+            for i in range(len(self._primary_key)):
+                if i != 0:
+                    where_clause += " and "
+                where_clause += f'{self._primary_key[i]}="{key[i]}"'
+        else:
+            where_clause += f'{self._primary_key}="{key}"'
```
gtopper:
```suggestion
        if isinstance(key, str):
            key = key.split(".")
        if isinstance(key, List):
            for i in range(len(self._primary_key)):
                if i != 0:
                    where_clause += " and "
                where_clause += f'{self._primary_key[i]}="{key[i]}"'
```

integration/test_flow_integration.py
```suggestion
    where_statement = ""
    if isinstance(key, str):
        key = key.split(".")
    if isinstance(key, list):
        for i in range(len(key_name)):
            if i != 0:
                where_statement += " and "
            where_statement += f'{key_name[i]}="{key[i]}"'
```

integration/test_flow_integration.py
```suggestion
    return results[0]._mapping
```

tests/test_flow.py
```diff
@@ -3915,3 +3916,26 @@ def test_rename():
...
+
+
+def test_read_sql_db():
+    import sqlalchemy as db
+
+    engine = db.create_engine("sqlite:///test.db", echo=True)
+    with engine.connect() as conn:
+        origin_df = pd.DataFrame({"string": ["hello", "world"], "int": [1, 2], "float": [1.5, 2.5]})
+        origin_df.to_sql("table_1", conn, if_exists="replace")
+        conn.close()
```
```suggestion
```

storey/sql_driver.py
```diff
@@ -0,0 +1,135 @@
...
+        table = self._table(table_path)
+
+        agg_val, values = await self._get_all_fields(key, table)
+        if not agg_val:
+            agg_val = None
+        if not values:
+            values = None
+        return [agg_val, values]
+
+    async def _load_by_key(self, container, table_path, key, attribute):
```
gtopper:
```suggestion
    async def _load_by_key(self, container, table_path, key, attributes):
```

storey/sql_driver.py
gtopper: Wait, why does this return None? Doesn't look like there's any reason for that.

storey/sql_driver.py
gtopper: Wait, why does this return None? Doesn't look like there's any reason for that.

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+class SQLSource(_IterableSource, WithUUID):
+    """Use SQL table as input source for a flow.
+
+
+    :parameter key_field: the primary key of the table.
+    :parameter time_field: column to be used as time for events.
+    :parameter id_field: column to be used as ID for events.
+    :parameter db_name: url string connection to sql database.
+    :parameter table_name: the name of the table to access,
+                                from the current database
```
gtopper, line 1026:
```suggestion
    :parameter table_name: the name of the table to access, from the current database
```

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+
```
```suggestion
    :parameter table_name: the name of the table to access, from the current database
```

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+            kwargs["key_field"] = key_field
+        if time_field is not None:
+            kwargs["time_field"] = time_field
+        if id_field is not None:
+            kwargs["id_field"] = id_field
+        _IterableSource.__init__(self, **kwargs)
+        WithUUID.__init__(self)
+
+        if not all([db_path, table_name]):
+            raise ValueError("cannot specify without db_path and table_name args")
```
```suggestion
```
Just make the parameters mandatory.

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+    :parameter db_name: url string connection to sql database.
+    :parameter table_name: the name of the table to access,
+                                from the current database
+
+    """
+
+    def __init__(
+        self,
+        db_path: str = None,
+        table_name: str = None,
```
```suggestion
        db_path: str,
        table_name: str,
```

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+        key_field: Optional[Union[str, List[str]]] = None,
```
gtopper: If you want the default to be None, then the type annotation should be
```suggestion
        key_field: Optional[Union[None, str, List[str]]] = None,
```
(or `NoneType` if you want to be pedantic)

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+
+        if not all([db_path, table_name]):
+            raise ValueError("cannot specify without db_path and table_name args")
+
+        self.table_name = table_name
+        self.db_path = db_path
+
+        self._key_field = key_field
+        if isinstance(time_field, str) and "." in time_field:
+            self._time_field = time_field.split(".")
```
gtopper: I'm pretty sure we don't support multiple time keys.

storey/sources.py
```diff
@@ -1012,3 +1012,99 @@ def _init(self):
...
+        else:
+            self._time_field = time_field
+        if isinstance(id_field, str) and "." in id_field:
```
gtopper: I'm pretty sure we don't support multiple IDs.

integration/integration_test_utils.py
davesh0812: @gtopper because the SQLDdriver  Is aggregation less.

integration/integration_test_utils.py
davesh0812: @gtopper I need to connect, to get the updated metadata of our db

storey/sources.py
davesh0812: @gtopper if you sure I will change it

storey/sources.py
davesh0812: @gtopper if you sure I will change it

integration/integration_test_utils.py
davesh0812: @gtopper your change made some bugs... so I revert it

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+        self._primary_key = primary_key
+
+    def _lazy_init(self):
+
+        if not self._sql_connection:
+            self._engine = db.create_engine(self._db_path)
+            self._sql_connection = self._engine.connect()
+
+    def _table(self, table_path):
+        import sqlalchemy as db
```
gtopper: This is already imported at the top level.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+
+    async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data):
+        self._lazy_init()
+
+        table = self._table(table_path)
+        try:
+            self._sql_connection.execute(table.insert(), [additional_data], autocommit=True)
+
+        except db.exc.IntegrityError:
+            raise RuntimeError(f"Failed to get insert {additional_data} to {table_path} table because IntegrityError")
```
gtopper: Don't we want to to know the specifics of the error? What does it mean "to get insert"? "because IntegrityError" should be "because of IntegrityError".

Seems to me that wrapping this exception is only hiding the original reason and traceback.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+        if not self._sql_connection:
+            self._engine = db.create_engine(self._db_path)
+            self._sql_connection = self._engine.connect()
+
+    def _table(self, table_path):
+        import sqlalchemy as db
+
+        metadata = db.MetaData()
+        while table_path.startswith("/"):
+            table_path = table_path[1:]
```
gtopper: We shouldn't be doing this every time we read or save an item. Why not get rid of the database part much earlier (up the stack)? We can call it the Table's "container".

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+    async def _save_key(self, container, table_path, key, aggr_item, partitioned_by_key, additional_data):
+        self._lazy_init()
+
+        table = self._table(table_path)
+        try:
+            self._sql_connection.execute(table.insert(), [additional_data], autocommit=True)
+
+        except db.exc.IntegrityError:
+            raise RuntimeError(f"Failed to get insert {additional_data} to {table_path} table because IntegrityError")
+        await self.close()
```
gtopper: Wait, why are we closing the connection every time we save a key to the DB?

requirements.txt
gtopper: This comment is still outstanding.

storey/sources.py
```diff
@@ -1012,3 +1012,82 @@ def _init(self):
...
+                else:
+                    key = body[self._key_field]
+                    if key is None:
+                        create_event = False
+            if create_event:
+                time = None
+                if self._time_field:
+                    time = body[self._time_field]
+                if self._id_field:
+                    _id = body[self._id_field]
```
gtopper: These two cases (where a bug was found in the previous iteration of this review) are still missing test coverage.

storey/sources.py
```diff
@@ -1012,3 +1012,82 @@ def _init(self):
...
+        engine = db.create_engine(self.db_path)
+        metadata = db.MetaData()
+        with engine.connect() as conn:
+            table = db.Table(self.table_name, metadata, autoload=True, autoload_with=engine)
+            results = conn.execute(db.select([table])).fetchall()
+
+        for namedtuple in results:
+            create_event = True
+            body = namedtuple._asdict()
+            body.pop("index")
```
gtopper: Why?

storey/sources.py
```diff
@@ -1012,3 +1012,82 @@ def _init(self):
...
+                    time = body[self._time_field]
+                if self._id_field:
+                    _id = body[self._id_field]
+                else:
+                    _id = self._get_uuid()
+                event = Event(body, key=key, time=time, id=_id)
+                await self._do_downstream(event)
+            else:
+                if self.context:
+                    self.context.logger.error(f"For {body} value of key {key_field} is None")
```
gtopper: `key_field` may not even be assigned at this point in the code.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+            autoload=True,
+            autoload_with=self._engine,
+        )
+
+    async def _save_schema(self, container, table_path, schema):
+        self._lazy_init()
+
+    async def _load_schema(self, container, table_path):
+        self._lazy_init()
+
```
gtopper:
```suggestion
```

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+        except db.exc.IntegrityError:
+            raise RuntimeError(f"Failed to get insert {additional_data} to {table_path} table because IntegrityError")
+        await self.close()
+
+    async def _load_aggregates_by_key(self, container, table_path, key):
+        self._lazy_init()
+        table = self._table(table_path)
+
+        agg_val, values = await self._get_all_fields(key, table)
```
gtopper: How does this not fail? `_get_all_fields()` does not return a tuple. It must not be tested.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+
+    async def close(self):
+        if self._sql_connection:
+            self._sql_connection.close()
+            self._sql_connection = None
+
+    async def _get_all_fields(self, key, table):
+        where_clause = self._get_where_clause(key)
+        try:
+            my_query = f"SELECT * FROM {table} where {where_clause}"
```
gtopper:
```suggestion
        my_query = f"SELECT * FROM {table} where {where_clause}"
        try:
```

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+            results = self._sql_connection.execute(my_query).fetchall()
+        except Exception as e:
+            raise RuntimeError(f"Failed to get key {key}. Response error was: {e}")
+
+        return results[0]._mapping
+
+    async def _get_specific_fields(self, key: str, table, attributes: List[str]):
+        where_clause = self._get_where_clause(key)
+        try:
+            my_query = f"SELECT {','.join(attributes)} FROM {table} where {where_clause}"
```
gtopper, line 98: Need to sanitize the inputs here and elsewhere in the code.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+        return results[0]._mapping
+
+    def supports_aggregations(self):
+        return False
+
+    def _get_where_clause(self, key):
+        where_clause = ""
+        if isinstance(key, str):
+            key = key.split(".")
+        if isinstance(key, List):
```
gtopper:
```suggestion
        if isinstance(key, list):
```

storey/sources.py
```diff
@@ -1012,3 +1012,82 @@ def _init(self):
             else:
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
             self._dfs.append(df)
+
+
+class SQLSource(_IterableSource, WithUUID):
+    """Use SQL table as input source for a flow.
+
```
gtopper, line 983:
```suggestion
```

integration/integration_test_utils.py
gtopper: @davesh0812 – this is still outstanding. Both the rename and `*args`. Not sure what bugs you're referring to...

integration/integration_test_utils.py
gtopper: But if that's the case, why in some other places we call `db.MetaData()` without first calling this? E.g. in `SQLSource._run_loop()`.

storey/sql_driver.py
```diff
@@ -0,0 +1,133 @@
...
+    """
+    SQL database connector.
+    :param db_url: database url
+    :param primary_key: the primary key of the table.
+    """
+
+    def __init__(
+        self,
+        primary_key: str,
+        db_path: str,
```
gtopper: The parameter name in the code and documentation should match.

storey/sources.py
```diff
@@ -1012,3 +1012,82 @@ def _init(self):
...
+    :parameter id_field: column to be used as ID for events.
+    :parameter db_name: url string connection to sql database.
+    :parameter table_name: the name of the table to access, from the current database
+    """
+
+    def __init__(
+        self,
+        db_path: str,
+        table_name: str,
+        key_field: Optional[Union[None, str, List[str]]] = None,
```
```suggestion
        key_field: Union[None, str, List[str]] = None,
```
Optional[X] == Union[None, X]

requirements.txt
davesh0812: the tests are failing without this requirement as you can see here https://github.com/mlrun/storey/pull/384/commits/650afa6bf2b7414cfd58c711e4e1433ca293bf74

tests/test_flow.py
```diff
@@ -3850,3 +3851,25 @@ def test_rename():
...
+        origin_df = pd.DataFrame({"string": ["hello", "world"], "int": [1, 2], "float": [1.5, 2.5]})
+        origin_df.to_sql("table_1", conn, if_exists="replace", index=False)
+    controller = build_flow(
+        [
+            SQLSource("sqlite:///test.db", "table_1", "string"),
+            Reduce([], append_and_return),
+        ]
+    ).run()
+
+    termination_result = controller.await_termination()
```
assaf758: add:
actual = termination_result
and below assert actual == expceted

tests/test_flow.py
```diff
@@ -3850,3 +3851,25 @@ def test_rename():
     controller.terminate()
     termination_result = controller.await_termination()
     assert termination_result == [{"b": 1, "c": 3, "e": 5}]
+
+
+def test_read_sql_db():
+    import sqlalchemy as db
```
assaf758, line 3858: define constant "sqlite:///test.db", use  it in the two places (better bring it from conftest.py)

integration/conftest.py
```diff
@@ -27,14 +27,20 @@
     get_redis_client,
     recursive_delete,
     remove_redis_table,
+    remove_sql_tables,
 )
 from storey import V3ioDriver
 from storey.redis_driver import RedisDriver
+from storey.sql_driver import SQLDriver
+
+SQL_DB = "sqlite:///test.db"
```
assaf758: change to SQLITE_DB for to make it explicit we are testing against sqlite

integration/integration_test_utils.py
```diff
@@ -30,6 +31,7 @@
 
 _non_int_char_pattern = re.compile(r"[^-0-9]")
 test_base_time = datetime.fromisoformat("2020-07-21T21:40:00+00:00")
+SQL_DB = "sqlite:///test.db"
```
assaf758: import from conftest

integration/test_flow_integration.py
```diff
@@ -760,7 +798,10 @@ def enrich(event, state):
 
 
 def test_write_table_metadata_columns(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
```
assaf758: apart from the skip of test_flow_aggregation (which is "by design" since you don't support aggregation with this driver), why are other tests  files skipped and only test_flow_integration.py is used?
Also in test_flow_integration.py, 8 tests are skipped with vague "not testing over SQLDriver".
Please see what tests you can add. If there is good/technical reason to skip a test please state it in the message.

integration/test_flow_integration.py
```diff
@@ -928,7 +980,20 @@ def test_writing_timedelta_key(setup_teardown_test):
 
 
 def test_write_two_keys_to_v3io_from_df(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
+        create_sql_table(
+            {"first_name": str, "last_name": str, "city": str},
+            setup_teardown_test._sql_table_name,
+            setup_teardown_test.sql_db_path,
+            ["first_name", "last_name"],
```
assaf758: move keys var above and use her and in the Table constructor

integration/test_flow_integration.py
```diff
@@ -928,7 +980,20 @@ def test_writing_timedelta_key(setup_teardown_test):
...
+        )
+        table = Table(
+            setup_teardown_test.table_name,
+            setup_teardown_test.driver(primary_key=["first_name", "last_name"], IsAggregationlessDriver=True),
```
assaf758, line 992: move outside the "if" clause and set  IsAggregationlessDriver based on driver_name

integration/test_flow_integration.py
```diff
@@ -892,7 +933,15 @@ async def get_kv_item(full_path, key):
 
 
 def test_writing_int_key(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
+        create_sql_table(
```
assaf758: the call to create_sql_table "spams" the pytest log with multiple lines - please see if you can remove / reduce:

`
INFO sqlalchemy.engine.Engine BEGIN (implicit)
2022-12-20 16:05:03,748 INFO sqlalchemy.engine.Engine PRAGMA main.table_info("ByvRKrMpEi")
2022-12-20 16:05:03,749 INFO sqlalchemy.engine.Engine [raw sql] ()
2022-12-20 16:05:03,750 INFO sqlalchemy.engine.Engine PRAGMA temp.table_info("ByvRKrMpEi")
2022-12-20 16:05:03,750 INFO sqlalchemy.engine.Engine [raw sql] ()
2022-12-20 16:05:03,752 INFO sqlalchemy.engine.Engine 
CREATE TABLE "ByvRKrMpEi" (
        num INTEGER NOT NULL, 
        color VARCHAR, 
        PRIMARY KEY (num)
)
2022-12-20 16:05:03,752 INFO sqlalchemy.engine.Engine [no key 0.00025s] ()
2022-12-20 16:05:03,986 INFO sqlalchemy.engine.Engine COMMIT
`

assaf758 (changes requested):

integration/test_flow_integration.py
```diff
@@ -636,7 +671,10 @@ def test_inner_join_by_key(setup_kv_teardown_test):
 
 
 def test_write_table_specific_columns(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
+        pytest.skip("test_write_table_specific_columns not testing over SQLDriver")
```
assaf758: pytest has "skipif" for these cases - please replace

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+
+
+class SQLSource(_IterableSource, WithUUID):
+    """Use SQL table as input source for a flow.
+
+
+    :parameter key_field: the primary key of the table.
+    :parameter time_field: column to be used as time for events.
+    :parameter id_field: column to be used as ID for events.
+    :parameter db_name: url string connection to sql database.
```
assaf758: db_name (in the comment) or db_path (in the constructor)

requirements.txt
assaf758: can you provide details what test is failing without this change?

integration/test_flow_integration.py
```diff
@@ -928,7 +980,20 @@ def test_writing_timedelta_key(setup_teardown_test):
...
 
 def test_write_two_keys_to_v3io_from_df(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
+        create_sql_table(
+            {"first_name": str, "last_name": str, "city": str},
+            setup_teardown_test._sql_table_name,
+            setup_teardown_test.sql_db_path,
+            ["first_name", "last_name"],
+        )
```
assaf758: also please see if you can reduce the boilerplate done in everytest:
create a Table constructor from a function and give that function sql_options, which will be used to call the create_sql_table if needed

requirements.txt
davesh0812: I didn't changed any thing- you can see that the lines are exactly the same

integration/test_flow_integration.py
davesh0812: I cant use skipif inside function

integration/integration_test_utils.py
davesh0812: I entered it inside also, it have to be there because as I said and understood, we have to connect to be able get the updated metadata

integration/test_flow_integration.py
davesh0812: all the tests in the integration are supported now for this driver except : 
1. kv tests
2. `test_writing_timedelta_key` - because sqlite issues
3. `test_write_three_keys_to_v3io_from_df` - waiting for your answer
4. `test_write_none_time` - when the time filed is none this test expected not to get the time field when querying the drivers - I don't think it something we want from sql driver.

assaf758: regarding test_write_three_keys_to_v3io_from_df, moshe.dc1e617eaae40eea3449baf3795bb5b50676c963 is the key used to store the first record in the table (user "moshe")
regarding test_write_none_time, why not?

integration/integration_test_utils.py
```diff
@@ -113,7 +115,19 @@ def remove_redis_table(table_name):
         count += 1
 
 
-drivers_list = ["V3ioDriver", "RedisDriver"]
+def remove_sql_tables():
+    import sqlalchemy as db
+
+    engine = db.create_engine(integration.conftest.SQLITE_DB)
+    with engine.connect() as _:
```
gtopper:
```suggestion
    with engine.connect():
```

integration/integration_test_utils.py
```diff
@@ -246,3 +260,31 @@ def _convert_nginx_to_python_type(typ, value):
         return datetime.utcfromtimestamp(secs + nanosecs / 1000000000)
     else:
         raise V3ioError(f"Type {typ} in get item response is not supported")
+
+
+def create_sql_table(schema, table_name, sql_db_path, key):
+    import sqlalchemy as db
+
+    engine = db.create_engine(sql_db_path)
+    with engine.connect() as _:
```
gtopper:
```suggestion
    with engine.connect():
```

integration/integration_test_utils.py
```diff
@@ -246,3 +260,31 @@ def _convert_nginx_to_python_type(typ, value):
...
+            elif col_type == timedelta or col_type == pd.Timedelta:
+                col_type = db.Interval
+            elif col_type == datetime or col_type == pd.Timestamp:
+                col_type = db.DateTime
+            elif col_type == bool:
+                col_type = db.Boolean
+            elif col_type == float:
+                col_type = db.Float
+            else:
+                raise TypeError(f"Unsupported type '{col_type}'")
```
gtopper:
```suggestion
                raise TypeError(f"Column '{col}' has unsupported type '{col_type}'")
```

integration/integration_test_utils.py
gtopper: Also I'm pretty sure you can move this line down to just before `engine` is actually needed.

integration/test_flow_integration.py
```diff
@@ -117,16 +119,67 @@ def _get_redis_kv_all_attrs(setup_teardown_test: ContextForTests, key: str):
...
+    sql_table = db.Table(
+        table_name,
+        metadata,
+        autoload=True,
+        autoload_with=engine,
+    )
+    my_query = f"SELECT * FROM {sql_table} where ({where_statement})"
+    with engine.connect() as conn:
+        results = conn.execute(my_query).fetchall()
+    return [r._asdict() for r in results][0]
```
gtopper: Isn't this the same as
```suggestion
    return results[0]._asdict()
``
?

File: integration/test_flow_integration.py [hunk 81]
```diff
@@ -937,23 +1032,25 @@ def test_write_two_keys_to_v3io_from_df(setup_teardown_test):
...
             NoSqlTarget(table),
         ]
     ).run()
     controller.await_termination()
-
     expected = {"city": "tel aviv", "first_name": "moshe", "last_name": "cohen"}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe.cohen")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe.cohen", keys)
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1042,20 +1141,20 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
     controller.await_termination()
 
     expected = {"n1": 1, "n2": 2, "n3": 3}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "1.2")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "1.2", ["n1", "n2"])
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1042,20 +1141,20 @@ def test_write_multiple_keys_to_v3io_from_csv(setup_teardown_test):
...
 
     expected = {"n1": 4, "n2": 5, "n3": 6}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "4.5")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "4.5", ["n1", "n2"])
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1067,16 +1166,20 @@ def test_write_multiple_keys_to_v3io(setup_teardown_test):
     controller.await_termination()
 
     expected = {"n1": 1, "n2": 2, "n3": 3}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "1.2")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "1.2", keys)
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1067,16 +1166,20 @@ def test_write_multiple_keys_to_v3io(setup_teardown_test):
...
 
     expected = {"n1": 4, "n2": 5, "n3": 6}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "4.5")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "4.5", keys)
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1101,20 +1204,19 @@ def set_moshe_time_to_none(data):
     controller.await_termination()
 
     expected = {"first_name": "yosi", "color": "yellow"}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "yosi")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "yosi", "first_name")
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

integration/test_flow_integration.py
```diff
@@ -1101,20 +1204,19 @@ def set_moshe_time_to_none(data):
...
 
     expected = {"first_name": "moshe", "color": "blue"}
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe")
-    assert actual == expected
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe", "first_name")
+    assert expected == actual
```
```suggestion
    assert actual == expected
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
             else:
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
             self._dfs.append(df)
+
+
+class SQLSource(_IterableSource, WithUUID):
+    """Use SQL table as input source for a flow.
+
+
```
```suggestion
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+        self._id_field = id_field
+
+    async def _run_loop(self):
+        import sqlalchemy as db
+
+        engine = db.create_engine(self.db_path)
+        with engine.connect() as conn:
+            metadata = db.MetaData()
+            table = db.Table(self.table_name, metadata, autoload=True, autoload_with=engine)
+            results = conn.execute(db.select([table])).fetchall()
```
gtopper: This will fetch the entire into memory if I understand it correctly. We probably want to avoid that.

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+
+        for namedtuple in results:
```
```suggestion
        for row in results:
```
(not a named tuple)

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+                    if key is None:
+                        self.context.logger.error(f"For {body} value there is no {self._key_field} field (key_field)")
+            time = None
+            if self._time_field:
+                time = body[self._time_field]
+            if self._id_field:
+                _id = body[self._id_field]
+            else:
+                _id = self._get_uuid()
+            event = Event(body, key=key, id=_id, processing_time=time)
```
```suggestion
            event = Event(body, key=key, id=_id)
```
Event time has been removed. Please don't set `processing_time` instead.

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+        if id_field is not None:
+            kwargs["id_field"] = id_field
+        _IterableSource.__init__(self, **kwargs)
+        WithUUID.__init__(self)
+
+        self.table_name = table_name
+        self.db_path = db_path
+
+        self._key_field = key_field
+        self._time_field = time_field
```
```suggestion
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
             self._dfs.append(df)
+
+
+class SQLSource(_IterableSource, WithUUID):
+    """Use SQL table as input source for a flow.
+
+
+    :parameter key_field: the primary key of the table.
+    :parameter time_field: column to be used as time for events.
```
```suggestion
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+    :parameter db_path: url string connection to sql database.
+    :parameter table_name: the name of the table to access, from the current database
+    """
+
+    def __init__(
+        self,
+        db_path: str,
+        table_name: str,
+        key_field: Union[None, str, List[str]] = None,
+        time_field: str = None,
```
```suggestion
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+                        if key_field not in body or pandas.isna(body[key_field]):
+                            self.context.logger.error(
+                                f"For {body} value there is no {self._key_field} " f"field (key_field)"
+                            )
+                            break
+                        key.append(body[key_field])
+                else:
+                    key = body[self._key_field]
+                    if key is None:
+                        self.context.logger.error(f"For {body} value there is no {self._key_field} field (key_field)")
```
gtopper: This isn't going to work as the error message implies it will. If the key is missing, an error will be raised. The error will only be logged if the key exists but is `None`.

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+            time = None
+            if self._time_field:
+                time = body[self._time_field]
```
```suggestion
```

storey/sources.py
```diff
@@ -976,3 +976,78 @@ def _init(self):
...
+            if self._id_field:
+                _id = body[self._id_field]
+            else:
+                _id = self._get_uuid()
```
```suggestion
                event_id = body[self._id_field]
            else:
                event_id = self._get_uuid()
```

storey/sql_driver.py
```diff
@@ -0,0 +1,123 @@
...
+import pandas as pd
+import sqlalchemy as db
+
+from storey.drivers import Driver
+
+
+class SQLDriver(Driver):
+    """
+    SQL database connector.
+    :param primary_key: the primary key of the table, format <str>.<str>..
```
gtopper: Need to explain that a list can also be passed. And remove the misleading trailing periods. Also, I see that `_extract_list_of_keys` supports an int key, but this documentation and the type annotation don't reflect that.

storey/sql_driver.py
```diff
@@ -0,0 +1,123 @@
...
+
+        return results[0]._mapping
+
+    async def _get_specific_fields(self, key: str, table, attributes: List[str]):
+        where_clause = self._get_where_clause(key)
+        try:
+            my_query = f"SELECT {','.join(attributes)} FROM {table} where {where_clause}"
+            results = self._sql_connection.execute(my_query).fetchall()
+        except Exception as e:
+            raise RuntimeError(f"Failed to get key {key}. Response error was: {e}")
```
gtopper:
```suggestion
            raise RuntimeError(f"Failed to get key '{key}'") from e
```

gtopper (changes requested):

storey/sql_driver.py
davesh0812: `_extract_list_of_keys ` function extract also the key value and this why it support int also..

integration/test_flow_integration.py
```diff
@@ -117,16 +119,67 @@ def _get_redis_kv_all_attrs(setup_teardown_test: ContextForTests, key: str):
...
+    import sqlalchemy as db
+
+    where_statement = ""
+    if isinstance(key, str):
+        key = key.split(".")
+    if isinstance(key, list):
+        for i in range(len(key_name)):
+            if i != 0:
+                where_statement += " and "
+            where_statement += f'{key_name[i]}="{key[i]}"'
```
gtopper: As discussed, we should only wrap values with quotes if the type is string. Not sure why the test passes, but it may be that sqlite is permissive beyond the SQL standard.

storey/sql_driver.py
```diff
@@ -0,0 +1,123 @@
...
+    def supports_aggregations(self):
+        return False
+
+    def _get_where_clause(self, key):
+        where_clause = ""
+        key = self._extract_list_of_keys(key)
+        for i in range(len(self._primary_key)):
+            if i != 0:
+                where_clause += " and "
+            where_clause += f'{self._primary_key[i]}="{key[i]}"'
```
gtopper: As discussed, we should only wrap values with quotes if the type is string. Not sure if we have test coverage of this, since the code is duplicated in test code.

storey/sql_driver.py
```diff
@@ -0,0 +1,123 @@
...
+import pandas as pd
+import sqlalchemy as db
+
+from storey.drivers import Driver
+
+
+class SQLDriver(Driver):
+    """
+    SQL database connector.
+    :param primary_key: the primary key of the table, format <str>.<str>... or [<str>, <str>,..]
```
gtopper:
```suggestion
    :param primary_key: the primary key of the table, format <str>.<str>... or [<str>, <str>, ...]
```

integration/integration_test_utils.py
gtopper: See my second comment.

gtopper (approved):
LGTM

integration/test_flow_integration.py
```diff
@@ -937,23 +1035,25 @@ def test_write_two_keys_to_v3io_from_df(setup_teardown_test):
...
-    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe.cohen")
+    actual = get_key_all_attrs_test_helper(setup_teardown_test, "moshe.cohen", keys)
     assert actual == expected
 
 
 # ML-775
 def test_write_three_keys_to_v3io_from_df(setup_teardown_test):
-    table = Table(setup_teardown_test.table_name, setup_teardown_test.driver())
+    if setup_teardown_test.driver_name == "SQLDriver":
+        pytest.skip("test_write_three_keys_to_v3io_from_df not testing over SQLDriver")
```
assaf758: testing => tested

integration/test_flow_integration.py
```diff
@@ -1101,20 +1205,23 @@ def set_moshe_time_to_none(data):
...
 
 
 def test_cache_flushing(setup_teardown_test):
-    table = Table(
-        setup_teardown_test.table_name,
-        setup_teardown_test.driver(),
-        flush_interval_secs=3,
-    )
+    if setup_teardown_test.driver_name == "SQLDriver":
+        pytest.skip("test_cache_flushing not testing over SQLDriver, SQL is not kv")
```
assaf758: test_cache_flushing is not tested over SQLDriver since SQL is not kv

assaf758 (changes requested):

assaf758 (approved):

PR #389, opened 2022-10-31
ML-2704: Fix redis NaN conversion

alxtkr77:


storey/redis_driver.py
```diff
@@ -194,7 +194,7 @@ def convert_redis_value_to_python_obj(cls, value):
             ret = math.inf
         elif value == "-inf":
             ret = -math.inf
-        elif value == "nan" or value == "-nan":
```
gtopper, line 197: Any chance we will also get `"-0/0"`?

storey/redis_driver.py
```diff
@@ -194,7 +194,7 @@ def convert_redis_value_to_python_obj(cls, value):
...
+        elif value == "nan" or value == "-nan" or value == "0/0":
```
gtopper: Any chance we will also get "-0/0"?

alxtkr77: Don't see how, but on the other hand Assaf added "-nan". I will add -0/0 as well to keep things symmetric.

gtopper (approved):
LGTM ✅

PR #393, opened 2022-11-24
 ML-2793: Fix float conversion in redis driver

alxtkr77:


integration/test_aggregation_integration.py
```diff
@@ -4085,3 +4085,57 @@ def test_aggregate_and_query_by_key_with_holes(setup_teardown_test):
     assert (
         actual == expected_results
     ), f"actual did not match expected. \n actual: {actual} \n expected: {expected_results}"
+
+
+def test_float_format(setup_teardown_test):
+    t0 = pd.Timestamp(test_base_time)
+    floats_array = [1.5089695129344164e-05, 1.07174329054775652e05, 1]
```
assaf758: not sure why, but when I tried this test without the fix it still passed.. do you understand why?
btw, changing floats_array to `floats_array = [1.5089695129344164e05,  1, 1.07174329054775652e-05]` does fail on the old code.

alxtkr77: I know why - because the fault happens on the last element of the floats_array. When '1' the old code handles it well. Will fix the test

storey/redis_driver.py
```diff
@@ -202,10 +202,13 @@ def convert_redis_value_to_python_obj(cls, value):
...
             # if value is a number, convert type to int / float
-            no_sign_value = value[1:] if value[0] in ("-", "+") else value
-            if no_sign_value.replace(".", "", 1).isdigit():
-                ret = int(value) if int(float(value)) == float(value) else float(value)
-            else:
+            try:
+                ret = float(value)
+                try:
+                    ret = int(value)
+                except ValueError:
```
gtopper, line 209: I wonder, what is a float, not nan/inf, and cannot be converted to int?

assaf758: >>> int('1.2')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.2'

gtopper: But it's already converted to float first...

gtopper: Oh ok, `int()` takes the original value, not the result of `float()`.

gtopper: But then, doesn't it make more sense to call `int()` first, and only call `float()` if the former fails?

assaf758: both options are correct but i think float-then-int is slightly better from perf perspective if we count the number of exceptions for each data-type:

float-then-int:
type      exception-count
float        0
int           1
string     1

int-then-float:
type      exception-count
float        1
int           0
string     2

assaf758 (approved):

integration/test_aggregation_integration.py
gtopper: Test still does not fail without the fix.

storey/redis_driver.py
gtopper: The count is a bit off but good enough for me.

integration/test_aggregation_integration.py
assaf758: will fix this on top of alex commits (would like to "catch" the next rc.. :-))

gtopper: Just make your own branch and PR? It's free :)

PR #394, opened 2022-11-28
ML-2502: Take time from event body as needed, replacing event time metadata

gtopper:
See [ML-2502](https://jira.iguazeng.com/browse/ML-2502) for full specification.


storey/aggregations.py
```diff
@@ -107,16 +110,27 @@ def f(element, features):
...
-            elif isinstance(key, list):
-                self.key_extractor = lambda element: [element.get(single_key) for single_key in key]
+        self._key_extractor = None
+        if key_field:
+            if callable(key_field):
+                self._key_extractor = key_field
+            elif isinstance(key_field, str):
+                self._key_extractor = lambda event: event.get(key_field)
+            elif isinstance(key_field, list):
+                self._key_extractor = lambda event: [event.get(single_key) for single_key in key_field]
```
theSaarco, line 120: The `list` option doesn't appear in the type hint for the `key_field` parameter (only `str` and `Callable`)

storey/aggregations.py
```diff
@@ -107,16 +110,27 @@ def f(element, features):
...
             else:
-                raise TypeError(f"key is expected to be either a callable or string but got {type(key)}")
+                raise TypeError(
+                    f"key_field is expected to be either a callable or " f"string but got {type(key_field)}"
```
theSaarco:
```suggestion
                    f"key_field is expected to be either a callable or " f"string or list of strings but got {type(key_field)}"
```

storey/dtypes.py
```diff
@@ -56,14 +56,15 @@ def __init__(
...
-                time = datetime.fromisoformat(time)
-            elif isinstance(time, int):
-                time = datetime.utcfromtimestamp(time)
+        if processing_time is not None and not isinstance(processing_time, datetime):
+            if isinstance(processing_time, str):
+                processing_time = datetime.fromisoformat(processing_time)
             else:
-                raise TypeError(f"Event time parameter must be a datetime, string, or int. Got {type(time)} instead.")
-        self.time = time or datetime.now(timezone.utc)
+                raise TypeError(
```
theSaarco, line 65: What happened to `int` processing? Seems you miss those lines now.

storey/flow.py
```diff
@@ -395,6 +395,22 @@ async def _do(self, event):
             await self._do_internal(event, fn_result)
 
 
+class DropColumns(Flow):
+    def __init__(self, columns, **kwargs):
+        super().__init__(**kwargs)
+        if not isinstance(columns, list):
+            columns = [columns]
+        self._columns = columns
```
theSaarco, line 403: Don't you need to force `full_event` here? If this step is added without `full_event` the `_do` function will fail, won't it?

storey/flow.py
```diff
@@ -932,18 +948,32 @@ def _init(self):
...
+        elif isinstance(key_field, str):
+            if key_field == "$key":
+                return lambda event: event.key
+            else:
+                return lambda event: event.body[key_field]
+        else:
+            raise ValueError(f"Unsupported key_field type {type(key_field)}")
+
+    @staticmethod
+    def _create_time_extractor(key_field) -> Callable:
```
theSaarco: This doesn't seem to be used anywhere.

storey/sources.py
```diff
@@ -821,21 +795,15 @@ def _blocking_io_loop(self):
...
-                                time_as_datetime = datetime.now()
                             if self._id_field:
                                 id_field = self._id_field
                                 if self._with_header and isinstance(id_field, str):
                                     id_field = field_name_to_index[id_field]
                                 id = parsed_line[id_field]
                             else:
                                 id = self._get_uuid()
-                            event = Event(element, key=key, time=time_as_datetime, id=id)
+                            # TODO FIXME
```
theSaarco: Can you explain the TODO FIXME here? What is the issue with this line?

storey/windowed_store.py
```diff
@@ -74,7 +78,9 @@ async def _do(self, event):
 
         element = event.body
         key = event.key
-        event_time = event.time
+        event_time = event.processing_time
+        if self._time_field:
+            event_time = event.body.pop(self._time_field)
```
theSaarco, line 82: How does that work if `_time_field` is an `EmitPolicy`?

theSaarco (changes requested):
Looks good, though I admit I don't know Storey enough to make sure all the relevant flows are covered. I did see some issues that need fixing, didn't go through the tests yet.

storey/windowed_store.py
gtopper: Why would `_time_field` be an `EmitPolicy`? They're unrelated things.

storey/flow.py
gtopper: Using `full_event` or not is up to the step implementation. This step does not use `full_event` at all.

storey/windowed_store.py
theSaarco: See line 39 above: 
```
time_field: Union[str, EmitPolicy, None] = None,
```
Maybe the issue is there then?

gtopper: Yup, not sure how that happened...

storey/dtypes.py
```diff
@@ -56,14 +56,17 @@ def __init__(
...
-        if time is not None and not isinstance(time, datetime):
-            if isinstance(time, str):
-                time = datetime.fromisoformat(time)
-            elif isinstance(time, int):
-                time = datetime.utcfromtimestamp(time)
+        if processing_time is not None and not isinstance(processing_time, datetime):
+            if isinstance(processing_time, str):
+                processing_time = datetime.fromisoformat(processing_time)
+            elif isinstance(processing_time, (int, float)):
+                processing_time = datetime.fromtimestamp(processing_time)
```
theSaarco: Using `utcfromtimestamp` (as was done before) is safer here since it always creates a UTC timestamp, regardless of local timezone. Since we always use UTC we should stick to it.

theSaarco (changes requested):
Looks good - one very small change needed.

theSaarco (approved):
LGTM

assaf758 (approved):

PR #398, opened 2022-12-12
ML-1167: Support multiple upstreams

gtopper:


yaronha (approved):

storey/flow.py
```diff
@@ -158,6 +168,7 @@ def to(self, outlet):
         if outlet._legal_first_step:
             raise ValueError(f"{outlet.name} can only appear as the first step of a flow")
         self._outlets.append(outlet)
+        outlet._inlets.append(self)
```
assaf758, line 171: can you explain what does the _inlets list holds?
i would think its list of steps leading to a specific test, and in that case the outlet._inlets should have "self" appended.

gtopper: If by "test" you mean "step" then yeah :)

assaf758 (approved):

PR #403, opened 2022-12-18
Remove documentation related to time metadata

gtopper:
Time metadata was deleted in #394.


storey/dataframe.py
```diff
@@ -26,8 +26,6 @@ class ReduceToDataFrame(Flow):
     :param columns: List of column names to be passed as-is to the DataFrame constructor. Optional.
     :param insert_key_column_as: Name of the column to be inserted for event keys. Optional.
         If not set, event keys will not be inserted into the DataFrame.
-    :param insert_time_column_as: Name of the column to be inserted for event times. Optional.
```
assaf758, line 29: I think should still be there - just need to change to insert_processing_time_column_as

assaf758 (approved):

PR #407, opened 2023-01-01
Parse timestamp if necessary in Writer trait

gtopper:
This fixes mlrun test `TestFeatureStore::test_preview_saves_changes`

### DIFF
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -46,6 +46,7 @@ class _Writer:
         index_cols: Union[str, List[Union[str, Tuple[str, str]]], None] = None,
         partition_cols: Union[str, List[str], None] = None,
         time_field: Union[None, str, int] = None,
+        time_format: Optional[str] = None,
         retain_dict: bool = False,
         storage_options: Optional[dict] = None,
     ):
@@ -135,6 +136,7 @@ class _Writer:
                 )
 
         self._time_field = time_field
+        self._time_format = time_format
 
     _type_string_to_pyarrow_type = {
         "str": pyarrow.string(),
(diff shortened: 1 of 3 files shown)


storey/targets.py
```diff
@@ -179,6 +181,15 @@ def _get_event_time(self, event):
             if isinstance(event.body, list) and isinstance(time_field, str):
                 time_field = self._col_to_index[time_field]
             event_time = event.body[time_field]
+            if isinstance(event_time, str):
+                if self._time_format:
+                    event_time = datetime.datetime.strptime(event_time, self._time_format)
+                else:
+                    event_time = datetime.datetime.fromisoformat(event_time)
+                event.body[time_field] = event_time
+            if isinstance(event_time, int):
```
assaf758: matter of style - but I would do elif - to stress it's a "switch" case.

storey/targets.py
```diff
@@ -509,6 +520,7 @@ def __init__(
         columns: Union[str, Union[List[str], List[Tuple[str, str]]], None] = None,
         partition_cols: Union[str, Union[List[str], List[Tuple[str, int]]], None] = None,
         time_field: Union[None, str, int] = None,
+        time_format: Optional[str] = None,
```
assaf758, line 526: add to ParquetTarget documentation

storey/targets.py
```diff
@@ -560,6 +572,7 @@ def __init__(
             index_cols,
             partition_cols,
             time_field,
+            time_format,
```
assaf758, line 578: what about other derivatives of Writer - do they need to add time_format as param? (CSVTarget?)

storey/targets.py
gtopper: Man, I knew I was going to forget this...

storey/targets.py
gtopper: Of all our targets, only parquet supports a timestamp type. CSV of course writes all values as strings. But I think `TSDBTarget` needs this too. I'll check and add if necessary.

gtopper: `CSVTarget` does not support partitioning (and writes everything as strings of course). I'm adding time format to `TSDBTarget` though (even though it doesn't support partitioning), as it needs to convert the time column because v3io-frames doesn't do it.

assaf758 (approved):

PR #408, opened 2023-01-03
SQLDriver - update data option (CreateOrReplaceAttributes) and Convert time fields from sql by pandas pd.read_sql

davesh0812:
1. lint bug
2. Convert time fields from sql by pandas `pd.read_sql`
3. CreateOrReplaceAttributes

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -136,7 +136,7 @@ class ContextForTests:
         def supports_aggregations(self):
             return False
 
-    def driver(self, *args, primary_key=None, is_aggregationless_driver=False, **kwargs):
+    def driver(self, *args, primary_key=None, is_aggregationless_driver=False, time_fields=None, **kwargs):
         if self.driver_name == "V3ioDriver":
             v3io_driver_class = ContextForTests.AggregationlessV3ioDriver if is_aggregationless_driver else V3ioDriver
             return v3io_driver_class(*args, **kwargs)
@@ -153,7 +153,7 @@ class ContextForTests:
         elif self.driver_name == "SQLDriver":
             if is_aggregationless_driver:
                 sql_driver_class = SQLDriver
-                return sql_driver_class(db_path=SQLITE_DB, primary_key=primary_key, time_fields=kwargs["time_fields"])
+                return sql_driver_class(db_path=SQLITE_DB, primary_key=primary_key, time_fields=time_fields)
             else:
                 pytest.skip("SQLDriver does not support aggregation")
         else:
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -14,6 +14,7 @@
 #
 import asyncio
 import os
+from datetime import datetime
 
 import fakeredis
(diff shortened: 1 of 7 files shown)


assaf758 (changes requested):
Looks good. Please add test coverage for this.
for example add to test_write_table_specific_columns() another table object, such that _load_by_key will be exercised (instead of using the cache)

storey/sql_driver.py
```diff
@@ -26,16 +26,14 @@ class SQLDriver(Driver):
     SQL database connector.
     :param primary_key: the primary key of the table, format <str>.<str>... or [<str>, <str>,...]
     :param db_path: database url
+    :param time_fields:
```
assaf758: add description

integration/test_flow_integration.py
```diff
@@ -714,13 +721,15 @@ def test_write_table_specific_columns(setup_teardown_test):
             "Avg": int,
         },
         keys,
+        time_fields,
     )
 
+    test_base_time_ = test_base_time if setup_teardown_test.driver_name != "SQLDriver" else test_base_time_sqlite
```
assaf758: why not add to the fixture

integration/test_flow_integration.py
```diff
@@ -163,19 +164,24 @@ def get_key_all_attrs_test_helper(
         result = _get_sql_by_key_all_attrs(setup_teardown_test, key, key_name)
         if time_fields is not None:
             for time_field in time_fields:
-                result[time_field] = datetime.strptime(result[time_field] + " UTC-0000", "%Y-%m-%d %H:%M:%S.%f UTC%z")
+                result[time_field] = (
```
assaf758: probably switched "if" and "else" clause?

integration/test_flow_integration.py
```diff
@@ -959,18 +997,44 @@ def enrich(event, state):
...
         "my_key": "tal",
     }
-    actual_cache = get_key_all_attrs_test_helper(
-        setup_teardown_test, "tal", keys, ["sometime", "first_activity", "last_event"]
-    )
+    actual_cache = get_key_all_attrs_test_helper(setup_teardown_test, "tal", keys, time_fields)
     assert expected_cache == actual_cache
 
+    # delete cache - testing load__by_key from driver
+    table._attrs_cache = {}
```
assaf758, line 1009: why not a new table object?

assaf758 (changes requested):
add to description the time-zone thing.
also comments inline.

integration/conftest.py
```diff
@@ -34,6 +35,8 @@
 from storey.sql_driver import SQLDriver
 
 SQLITE_DB = "sqlite:///test.db"
+test_base_time = datetime.fromisoformat("2020-07-21T21:40:00+00:00")
```
assaf758: move both into the ContextForTests class

assaf758 (approved):

PR #413, opened 2023-01-17
ML-3192: Parse string event timestamps in `AggregateByKey`

gtopper:


storey/aggregations.py
```diff
@@ -60,6 +60,8 @@ class AggregateByKey(Flow):
         and added to every event. (Optional)
     :param aliases: Dictionary specifying aliases for enriched or aggregate columns, of the
      format `{'col_name': 'new_col_name'}`. (Optional)
+    :param time_format: If the value of the time field is of type string, use this format, as defined in
```
assaf758: :param time_format: When the value of the time field is of type string, parsing will be done using this param (as defined in datetime.strptime()) if it's specified. Otherwise, ISO-8601 will be used for parsing.

tests/test_aggregate_by_key.py
```diff
@@ -3784,24 +3784,31 @@ def test_aggregation_unique_fields():
         ).run()
 
 
-def test_fixed_window_aggregation_with_first_and_last_aggregates():
+@pytest.mark.parametrize("timestamp", [True, False])
```
assaf758, line 3787: i'm not following, where is the timestamp parameter being used?

gtopper: One line below.

assaf758 (approved):

PR #426, opened 2023-03-12
ML-3566: Fix error when data or filters have tz info but the other doesn't

gtopper:
[ML-3566](https://jira.iguazeng.com/browse/ML-3566)

This error:
```
pyarrow.lib.ArrowInvalid: Cannot compare timestamp with timezone to timestamp without timezone
```
was introduced in pyarrow 7.


davesh0812 (approved):
LGTM :)

PR #429, opened 2023-03-26
ML-3654: Set result when using error handler in graph

davesh0812:
see also mlrun pr - https://github.com/mlrun/mlrun/pull/3341

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -228,9 +228,9 @@ class Flow:
                 return
             elif self.context and hasattr(self.context, "push_error"):
                 if event._awaitable_result:
-                    # set event as result to prevent double push to error stream
-                    none_or_coroutine = event._awaitable_result._set_result(
-                        Event(body={"error_msg": message, "origin_body": event.body})
+                    # set event as result to prevent double push to error stream ?
+                    none_or_coroutine = event._awaitable_result._set_error(
+                        ex
                     )
                     if none_or_coroutine:
                         await none_or_coroutine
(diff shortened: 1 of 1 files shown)


storey/flow.py
```diff
@@ -215,22 +215,30 @@ async def _do_and_recover(self, event):
...
-                else:
-                    raise ex
-            event.origin_state = self.name
-            event.error = ex
-            return await recovery_step._do(event)
+            message = traceback.format_exc()
+            if recovery_step:
+                event.origin_state = self.name
+                if event._awaitable_result:
+                    event._awaitable_result._set_result(Event(body={"error_msg": message, "origin_body": event.body}))
```
gtopper, line 222: You really can't just make up an event to put there. You're thinking about serving, but storey isn't just used for serving.

gtopper: Furthermore, there was already a call to `_set_error` there, which calls `_set_result`.

gtopper: It also makes no sense to set the result and then do the recovery. The recovery should lead somewhere that completes the result.

storey/flow.py
```diff
@@ -215,22 +215,30 @@ async def _do_and_recover(self, event):
...
+                if self.logger:
+                    self.logger.error(
+                        f"step {self.name} got error {ex} when processing an event:\n {event.body}\n" f"{message}"
+                    )
+                await recovery_step._do(event)
+                return
```
gtopper, line 228: The result of `await recovery_step._do(event)` should be returned as before, not `None`. This matters for flow termination.

gtopper (changes requested):
Please see my comments on [ML-3654](https://jira.iguazeng.com/browse/ML-3654) for RCA and proposed action points.

davesh0812:
will redesign as discussed

PR #431, opened 2023-04-09
ML-3603: `CSVSource` from pandas

tomerm-iguazio:
Changed CSVSource to work with pandas.read_csv.

There are a few changes that may affect the user:

###CSVSource:

1 ) When parsing a DataFrame with an empty value in a line, the result is now changed from None to numpy.nan, as we rely on pandas' "read_csv" methods.
2) Users can now use integers as indexes when parsing dates, even when there is a header. Parsing of no-header 3) DataFrames with a string column name is blocked.
4) Catching every error while building the events has been canceled.

###DataframeSource / ParquetSource:

1) An error is now raised if a key/id does not exist (including time_field/parse_dates in CSVSource). Previously, we only wrote a log error if the key did not match for each line.
2) The validator now raises a custom exception with proper messages, such as "ValueError" for missing keys or "IndexError" for missing indexes in CSVSource.


tests/test_flow.py
```diff
@@ -2915,37 +3011,6 @@ def test_csv_reader_parquet_write_nanosecs(tmpdir):
...
-
-        def push_error(self, event, message, source):
-            self.errors.append({"event": event, "message": message, "source": source})
-
-    context = PushErrorContext()
-
-    controller = build_flow(
-        [
-            CSVSource(
-                "tests/test-type-inference.csv",
```
gtopper, line 2931: Need to delete the file too

tests/test_flow.py
```diff
@@ -359,6 +360,43 @@ def test_csv_reader_with_key_and_timestamp():
...
+                    "tests/test-with-timestamp.csv",
+                    header=True,
+                    key_field="t",
+                    **datetime_field,
+                    timestamp_format="%d/%m/%Y %H:%M:%S",
+                )
+            ]
+        ).run()
+        controller.await_termination()
+    assert str(value_error.value) == "Missing column provided to 'parse_dates': 'non_existent_column'"
```
gtopper, line 379: Better reformulate the error message, I think:
```suggestion
    assert str(value_error.value) == "Column 'non_existent_column' provided to parse_dates does not exist in the data"
```

tests/test_flow.py
```diff
@@ -359,6 +360,43 @@ def test_csv_reader_with_key_and_timestamp():
     assert termination_result[1].body == ["m2", datetime(2020, 2, 16, 2, 0), 14, False]
 
 
+@pytest.mark.parametrize(
+    "datetime_field", [{"parse_dates": ["t", "non_existent_column"]}, {"time_field": "non_existent_column"}]
```
```suggestion
    "csv_source_kwargs", [{"parse_dates": ["t", "non_existent_column"]}, {"time_field": "non_existent_column"}]
```

storey/sources.py
```diff
@@ -622,69 +620,94 @@ def __init__(
...
+
+    def _is_nan_validator(self, result, body, field_type, field):
+        if pandas.isna(result) or result is None:
+            raise self.NoneKeyException(f"For {body} value of {field_type} {field} is None")
+
+
+class CSVSource(DataframeSource):
+    """Reads Csv files as input source for a flow.
+
+    :parameter paths: paths to Parquet files
```
gtopper: Parquet? What happened to the original documentation? This whole docstring looks completely wrong.

tests/test_flow.py
```diff
@@ -3440,7 +3505,10 @@ def test_csv_source_with_none_values():
         "2021-04-21 15:56:53.385444",
     ]
     assert termination_result[1].key == "b"
-    assert termination_result[1].body == ["b", True, None, math.nan, math.nan, None]
+    assert all(
```
gtopper: Need to break down into multiple asserts, or we won't know what actually failed.

tests/test_flow.py
```diff
@@ -3517,6 +3585,40 @@ def test_none_key_num_is_not_written():
...
+        controller = build_flow(
+            [
+                DataframeSource(dfs=data, key_field="non_existent_column"),
+            ]
+        ).run()
+        controller.await_termination()
+
+    assert (
+        str(value_error.value)
+        == "\"KeyError occurred: keys ['non_existent_column'] missing from df. Df path: file path was not provided.\""
```
gtopper: If you have to chain exceptions, there is a way to do that (with the `from` keyword). We don't want to get quotes inside of the string for some reason.

Also, as we discussed, there are only file paths, no dataframe paths. And let's call them dataframes or DataFrames, not df.

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
-            self._event_buffer.put(ex)
-        self._event_buffer.put(_termination_obj)
-
-    def _get_event(self):
-        event = self._event_buffer.get()
-        if isinstance(event, BaseException):
-            raise event
-        return event
+    def _get_element(self, body: OrderedDict):
+        if self._build_dict and self._with_header:
```
gtopper: What happens if `self._build_dict and not self._with_header`? We just ignore the `build_dict` setting? That doesn't seem right.

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
-            if field == "":
-                return None
-            return self._datetime_from_timestamp(field)
-        if typ == "n":
-            return None
-        raise TypeError(f"Unknown type: {typ}")
+        self._dfs = []
+        for path in self._paths:
+            header = 0 if self._with_header else None
+            #  with 'header = None' it will automatically put indexes as columns.
```
```suggestion
            #  with header=None, columns will be indexed by position (0, 1, 2, ...)
            header = 0 if self._with_header else None
```

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
-                    self._dates_indices = parse_dates
-        if isinstance(self._time_field, int):
-            if self._with_header:
-                raise ValueError("time field can be int only when there is no header")
+            if not self._with_header and any([isinstance(f, str) for f in self._parse_dates]):
+                raise ValueError("parse_dates can be list of str only when there is header")
+            self._dates_indices.extend(parse_dates)
+        if self._time_field is not None:
+            if not header and isinstance(self._time_field, str):
+                raise ValueError("time_field can only be set to an integer when with_header is false")
```
```suggestion
                raise ValueError("time_field parameter can only be an integer in conjunction with with_header=False")
```

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
             self._dates_indices.append(self._time_field)
-
-        if not header and isinstance(key_field, str):
+        if not header and (
+            isinstance(key_field, str)
+            or (isinstance(key_field, list) and any([isinstance(key, str) for key in key_field]))
```
gtopper: I really hope we're not accepting lists of mixed types here... It's probably enough to test just the first element.

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
-            if isinstance(parse_dates, List):
-                if self._with_header and any([isinstance(f, int) for f in self._parse_dates]):
-                    raise ValueError("parse_dates can be list of int only when there is no header")
-                if not self._with_header and all([isinstance(f, int) for f in self._parse_dates]):
-                    self._dates_indices = parse_dates
-        if isinstance(self._time_field, int):
-            if self._with_header:
-                raise ValueError("time field can be int only when there is no header")
+            if not self._with_header and any([isinstance(f, str) for f in self._parse_dates]):
+                raise ValueError("parse_dates can be list of str only when there is header")
```
```suggestion
                raise ValueError("parse_dates can be list of strings when with_header=True")
```

storey/sources.py
```diff
@@ -730,212 +754,80 @@ def __init__(
...
-        raise TypeError(f"Unknown type: {typ}")
+        self._dfs = []
+        for path in self._paths:
+            header = 0 if self._with_header else None
+            #  with 'header = None' it will automatically put indexes as columns.
+            df = pandas.read_csv(
+                path,
+                header=header,
+                parse_dates=self._dates_indices,
+                date_parser=lambda x: self._datetime_from_timestamp(x),
```
```suggestion
                date_parser=self._datetime_from_timestamp,
```

storey/sources.py
tomerm-iguazio: I disagree, I think we should check every key.

storey/sources.py
tomerm-iguazio: You are right, but this what we had in the pervious version so I did not change it.
what do you suggest? we do not have the headers to make it as dict.

tests/test_flow.py
tomerm-iguazio: fixed tnx

tests/test_flow.py
tomerm-iguazio: This is the default pandas.read_csv error message, I did not write it.

storey/sources.py
tomerm-iguazio: we can get mixed types with header =True as before. We need to talk about that...

storey/sources.py
```diff
@@ -1112,3 +1017,10 @@ async def _run_loop(self):
                     event = Event(body, key=key, id=event_id)
                     await self._do_downstream(event)
         return await self._do_downstream(_termination_obj)
+
+
+def get_path_message(path):
```
gtopper: Need to make this a private of `DataframeSource`, not a public top-level API. You can make it a `@staticmethod` though.

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
+                self._is_nan_validator(result=result, body=body, field_type=field_type, field=field)
+        return result
 
-        while True:
-            events = await asyncio.get_running_loop().run_in_executor(None, get_multiple)
-            for event in events:
-                res = await self._do_downstream(event)
-                if event is _termination_obj:
-                    return res
+    def _field_validator(self, df, key_field, id_field, path="file path was not provided."):
```
gtopper: The default value for `path` is wrong.

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
+        str_id_field = None
+        if id_field is not None:
+            if isinstance(id_field, str):
+                str_id_field = id_field
+            elif isinstance(id_field, int) and (id_field < 0 or id_field >= len(df.columns)):
+                raise IndexError(f"IndexError: id {id_field} is int and isn't in df index range.{path_message}")
+        if key_field:
+            key_field = [key_field] if not isinstance(key_field, list) else key_field
+            str_key_field = [key for key in key_field if isinstance(key, str)]
+            int_key_field = [key for key in key_field if isinstance(key, int)]
```
gtopper: This has gone way too far I think. We don't need to support a mix of ints and strings in `key_field`. It serves no useful purpose to the user.

storey/sources.py
gtopper: But actually better just inline this method. It's too short and only used in 2 places.

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
-    async def _run_loop(self):
-        asyncio.get_running_loop().run_in_executor(None, self._blocking_io_loop)
-
-        def get_multiple():
-            events = [self._get_event()]
-            while not self._event_buffer.empty() and len(events) < 128:
-                events.append(self._get_event())
-            return events
+    def _get_by_field_or_index(self, field, body: OrderedDict, field_type: str, raise_exception=False):
+        result = None
```
gtopper: This line has no effect.

storey/sources.py
gtopper: I guess just make it raise an error in `__init__`.

storey/sources.py
gtopper: Not to nitpick, but method names are generally verbs, like `_validate_fields` (plural).

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
+        if self._with_header and isinstance(field, str):
```
gtopper: Shouldn't we be checking just `self._build_dict`? Anyway, the condition doesn't seem right. If `self._with_header` is false and `field` is a string, we will end up trying to access a list using a string index (in the else case).

storey/sources.py
```diff
@@ -622,49 +620,85 @@ def __init__(
...
+
+        if id_field and id_field not in df.columns:
+            raise ValueError(f"id field '{id_field}' missing from dataframe.{path_message}")
+
+    class NoneKeyException(Exception):
+        pass
 
-class CSVSource(_IterableSource, WithUUID):
+    def _is_nan_validator(self, result, body, field_type, field):
+        if pandas.isna(result) or result is None:
```
gtopper: `pandas.isna(None)` returns `True`, so the second check is redundant.
```suggestion
        if pandas.isna(result):
```

storey/sources.py
```diff
@@ -622,49 +620,85 @@ def __init__(
...
+                raise ValueError(f"keys {missing_keys} missing from dataframe.{path_message}")
+
+        if id_field and id_field not in df.columns:
+            raise ValueError(f"id field '{id_field}' missing from dataframe.{path_message}")
+
+    class NoneKeyException(Exception):
+        pass
 
-class CSVSource(_IterableSource, WithUUID):
+    def _is_nan_validator(self, result, body, field_type, field):
```
gtopper: This method is too small and the way that the call hierarchy is structured is somewhat of an antipattern. I suggest that, instead of passing all the information down and raising an exception, just return the `NaN`/`None` result and raise an exception where appropriate (in `_get_by_field_or_index` – it's better to just delete `_is_nan_validator`).

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
-            if isinstance(parse_dates, List):
-                if self._with_header and any([isinstance(f, int) for f in self._parse_dates]):
-                    raise ValueError("parse_dates can be list of int only when there is no header")
-                if not self._with_header and all([isinstance(f, int) for f in self._parse_dates]):
-                    self._dates_indices = parse_dates
-        if isinstance(self._time_field, int):
-            if self._with_header:
-                raise ValueError("time field can be int only when there is no header")
+            if not header and any([isinstance(f, str) for f in self._parse_dates]):
+                raise ValueError("parse_dates can be list of strings when with_header=True")
```
gtopper: I think you meant "can only".

storey/sources.py
gtopper: Well, the key field list comes from the user. We don't have to support a mixed list of strings and ints. That's not useful to the user and only adds needless complexity to our code.

gtopper (changes requested):

storey/sources.py
gtopper: Also, "type" has a meaning that does not make sense here. Need a better name for this parameter.

storey/sources.py
```diff
@@ -622,49 +620,85 @@ def __init__(
...
                     await self._do_downstream(event)
-                else:
+                except self.NoneKeyException as key_error:
                     if self.context:
-                        self.context.logger.error(f"For {body} value of key {key_field} is None")
+                        self.context.logger.error(str(key_error))
+
         return await self._do_downstream(_termination_obj)
 
+    def _get_by_field_or_index(self, field, body: OrderedDict, field_type: str, raise_exception=False):
```
gtopper: "type" has a meaning that does not make sense here. Need a better name for this parameter.

storey/sources.py
```diff
@@ -730,212 +765,81 @@ def __init__(
...
+            elif isinstance(id_field, int) and (id_field < 0 or id_field >= len(df.columns)):
+                raise IndexError(f"IndexError: id {id_field} is int and isn't in df index range.{path_message}")
+        if key_field:
+            key_field = [key_field] if not isinstance(key_field, list) else key_field
+            str_key_field = [key for key in key_field if isinstance(key, str)]
+            int_key_field = [key for key in key_field if isinstance(key, int)]
+            out_of_range_keys = [int_key for int_key in int_key_field if int_key < 0 or int_key >= len(df.columns)]
+            if out_of_range_keys:
+                raise IndexError(
+                    f"IndexError: keys {out_of_range_keys} are int and are not in df index range.{path_message}"
```
gtopper: * df -> DataFrame
* Please don't repeat the error type in the error message.

storey/sources.py
tomerm-iguazio: ok :)

storey/sources.py
tomerm-iguazio: Okay, but strings and integers are still supported, so we cannot cancel this support.
Mixing integers and strings was not supported previously, so we may need to create a check to ensure that we have either all integers or all strings. However, I don't think it is necessary at this point.

storey/sources.py
tomerm-iguazio: you are right! That's why I added a validator in __init__ in CSVSource:


      
```
        if not header and (
            isinstance(key_field, str)
            or (isinstance(key_field, list) and any([isinstance(key, str) for key in key_field]))
        ):
            raise ValueError("key_field can only be set to an integer when with_header is false")
```

tomerm-iguazio: do you think I need to check it here too?

storey/sources.py
tomerm-iguazio: you are right fixed

storey/sources.py
tomerm-iguazio: Yes, but then we would be raising the same NoneKeyException with the same message in two different places.

tests/test_flow.py
```diff
@@ -359,6 +360,43 @@ def test_csv_reader_with_key_and_timestamp():
...
+def test_parse_dates_index_error(csv_source_kwargs):
+    with pytest.raises(IndexError):
+        controller = build_flow(
+            [
+                CSVSource(
+                    "tests/test-with-timestamp.csv",
+                    header=True,
+                    key_field="k",
+                    **csv_source_kwargs,
+                    timestamp_format="%d/%m/%Y %H:%M:%S",
```
```suggestion
                    timestamp_format="%d/%m/%Y %H:%M:%S",
                    **csv_source_kwargs,
```

tests/test_flow.py
```diff
@@ -3517,6 +3576,34 @@ def test_none_key_num_is_not_written():
     assert result == expected
 
 
+def test_field_validator_key():
```
```suggestion
def test_dataframe_source_missing_key_column():
```

tests/test_flow.py
```diff
@@ -3517,6 +3576,34 @@ def test_none_key_num_is_not_written():
...
+            [
+                DataframeSource(dfs=data, key_field="non_existent_column"),
+            ]
+        ).run()
+        controller.await_termination()
+
+    assert str(value_error.value) == "keys ['non_existent_column'] missing from dataframe."
+
+
+def test_field_validator_id():
```
```suggestion
test_dataframe_source_missing_id_column
```

tests/test_flow.py
```diff
@@ -3517,6 +3576,34 @@ def test_none_key_num_is_not_written():
...
+        data = pd.DataFrame({"id": [1, 2, 3], "some_data": ["data", "random_data", "my_data"]})
+
+        controller = build_flow(
+            [
+                DataframeSource(dfs=data, key_field="non_existent_column"),
+            ]
+        ).run()
+        controller.await_termination()
+
+    assert str(value_error.value) == "keys ['non_existent_column'] missing from dataframe."
```
gtopper: The common case is to have just one key column, so let's make it user friendly. And please pay attention to the distinction between a key and a key _column_ everywhere.
```suggestion
    assert str(value_error.value) == "key column 'non_existent_column' is missing from dataframe"
```

tests/test_flow.py
```diff
@@ -3545,6 +3632,39 @@ def test_none_key_date_is_not_written():
     assert result == expected
 
 
+def test_str_id_with_no_header():
```
gtopper, line 3635: Please mention csv_source in relevant test names. The ID here is an integer, not a string, so it seems that the test name is wrong (this also seems to apply to multiple tests, not just this one).

tests/test_flow.py
```diff
@@ -389,6 +427,58 @@ def test_csv_reader_none_in_keyfield_should_send_error_log():
     assert "value of key k is None" in logger.logs[0][1][0]
 
 
+def test_csv_reader_key_error():
```
gtopper: Since you're adding a bunch of tests, please rename csv_reader -> csv_source. This rename happened a long time ago, but I guess some of the tests weren't changed.

storey/sources.py
```diff
@@ -730,212 +759,80 @@ def __init__(
...
-                    return res
+    def _validate_fields(self, df, key_field, id_field, path=""):
+        path_message = f" File path: {path}." if path else ""
+        str_key_field = []
+        str_id_field = None
+        if id_field is not None:
+            if isinstance(id_field, str):
+                str_id_field = id_field
+            elif isinstance(id_field, int) and (id_field < 0 or id_field >= len(df.columns)):
+                raise IndexError(f"Id {id_field} is int and isn't in df index range.{path_message}")
```
```suggestion
                raise IndexError(f"id column '{id_field}' is of type int and is not in dataframe index range.{path_message}")
```

storey/sources.py
gtopper: What about this one?

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
-                        id = self._get_uuid()
-                    event = Event(body, key=key, id=id)
+                        line_id = self._get_uuid()
+                    element = self._get_element(body=body)
+                    event = Event(element, key=key, id=line_id)
                     await self._do_downstream(event)
                 else:
                     if self.context:
-                        self.context.logger.error(f"For {body} value of key {key_field} is None")
+                        self.context.logger.error(f"value of key {none_key} is None For {body}")
```
```suggestion
                        self.context.logger.error(f"value of key column '{none_key}' is None for {body}")
```
better yet, replace None with the actual value, which may be NaN rather than None.

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
         if isinstance(dfs, pandas.DataFrame):
             dfs = [dfs]
+        if dfs:
+            for df in dfs:
+                self._validate_fields(df=df, key_field=key_field, id_field=id_field)
         self._dfs = dfs
         self._key_field = key_field
         self._id_field = id_field
 
+    def _get_key(self, body):
```
gtopper: The contract of this method is rather unintuitive. Needs to be refactored or documented.

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
+            missing_keys = list(set(key_field) - set(df.columns))
+            if missing_keys:
+                raise ValueError(f"keys {missing_keys} missing from dataframe.{path_message}")
+
+        if id_field and id_field not in df.columns:
+            raise ValueError(f"id field '{id_field}' missing from dataframe.{path_message}")
 
-class CSVSource(_IterableSource, WithUUID):
+    class NoneKeyException(Exception):
+        pass
```
```suggestion
```

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
                     await self._do_downstream(event)
                 else:
                     if self.context:
-                        self.context.logger.error(f"For {body} value of key {key_field} is None")
+                        self.context.logger.error(f"value of key {none_key} is None For {body}")
         return await self._do_downstream(_termination_obj)
 
+    def _get_by_field_or_index(self, field, body: OrderedDict):
+        result = body[field]
+        return result
```
```suggestion
        return body[field]
```

storey/sources.py
```diff
@@ -730,212 +759,80 @@ def __init__(
...
-        def get_multiple():
-            events = [self._get_event()]
-            while not self._event_buffer.empty() and len(events) < 128:
-                events.append(self._get_event())
-            return events
+    def _get_by_field_or_index(self, field, body: OrderedDict):
+        if self._with_header and isinstance(field, str):
+            result = super()._get_by_field_or_index(field=field, body=body)
+        else:
+            result = list(body.items())[field][1]
```
gtopper: I'm not sure how we ended up accessing a dictionary by position, but if it does make sense,
```suggestion
            result = list(body.values())[field]
```

storey/sources.py
```diff
@@ -730,212 +759,80 @@ def __init__(
...
-                if not self._with_header and all([isinstance(f, int) for f in self._parse_dates]):
-                    self._dates_indices = parse_dates
-        if isinstance(self._time_field, int):
-            if self._with_header:
-                raise ValueError("time field can be int only when there is no header")
-            self._dates_indices.append(self._time_field)
-
-        if not header and isinstance(key_field, str):
+            if not header and any([isinstance(f, str) for f in self._parse_dates]):
+                raise ValueError("parse_dates can be only list of strings when with_header=True")
```
```suggestion
                raise ValueError("parse_dates can only be a list of strings when with_header=True")
```

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
+
+    def _get_element(self, body: OrderedDict):
+        return dict(body)
+
     async def _run_loop(self):
         for df in self._dfs:
             for namedtuple in df.itertuples():
-                create_event = True
-                body = namedtuple._asdict()
+                body = OrderedDict(namedtuple._asdict())
```
gtopper: Do we have to use an `OrderedDict`? Is this because of the weird support for accessing dictionaries by position? Because I think we should drop it.

If we end up keeping it though, I hope that there's a more efficient way to construct it than constructing a dictionary, populating an ordered dictionary from the dictionary we just constructed, and throwing the first dictionary away.

storey/sources.py
```diff
@@ -622,49 +621,78 @@ def __init__(
...
+                    key.append(single_key)
+
+            else:
+                key = self._get_by_field_or_index(field=self._key_field, body=body)
+                if pd.isna(key):
+                    return key, self._key_field
+        return key, None
+
+    def _get_element(self, body: OrderedDict):
+        return dict(body)
```
gtopper: As I wrote, I want to drop `OrderedDict`, but if we end up keeping it, why does it need to be converted to a regular dictionary here?

tests/test_flow.py
```diff
@@ -3545,6 +3632,39 @@ def test_none_key_date_is_not_written():
...
+
+def test_str_key_field_with_no_header():
+    with pytest.raises(ValueError) as value_error:
+        build_flow(
+            [
+                CSVSource("tests/test.csv", header=False, key_field="n1"),
+            ]
+        ).run()
+
+    assert str(value_error.value) == "key_field can only be set to an integer when with_header is false"
```
```suggestion
    assert str(value_error.value) == "key_field can only be set to an integer when with_header is False"
```

tests/test_flow.py
```diff
@@ -3545,6 +3632,39 @@ def test_none_key_date_is_not_written():
...
+
+def test_mixed_key_types():
+    with pytest.raises(ValueError) as value_error:
+        build_flow(
+            [
+                CSVSource("tests/test.csv", header=True, key_field=["n1", 2]),
+            ]
+        ).run()
+
+    assert str(value_error.value) == "Keys should not contain both integer and string values"
```
gtopper: To be consistent with other error messages:
```suggestion
    assert str(value_error.value) == "key_field should not contain both integer and string values"
```

storey/sources.py
tomerm-iguazio: added a mixed type validator in _validate_fields.I do not think I should add a test for first element because in that solution I will need also to check if the list is not empty.

storey/sources.py
tomerm-iguazio: we need to save the order of the dict, so _get_element in CSVSource will be able to access to the indexes of the columns as keys.

PR #433, opened 2023-04-13
ML-3527: Sanitizing all queries with sqlalchemy objects

davesh0812:


storey/sql_driver.py
```diff
@@ -87,17 +87,23 @@ async def close(self):
...
 
     async def _get_all_fields(self, key, table):
-        where_clause = self._get_where_clause(key, table)
-        query = f"SELECT * FROM {table} where {where_clause}"
-        results = pd.read_sql(query, con=self._sql_connection, parse_dates=self._time_fields).to_dict(orient="records")
+        key = self._extract_list_of_keys(key)
+        select_object = db.select(table).where(
+            db.and_(getattr(table.c, self._primary_key[i]) == key[i] for i in range(len(self._primary_key)))
+        )
+        results = pd.read_sql(select_object, con=self._sql_connection, parse_dates=self._time_fields).to_dict(
```
gtopper, line 94: So you went with full DSL – cool. We need to do this in `SQLSource` too. It still uses string substitution. For the tests I suppose it doesn't matter that much, but still nice to have if it's not too much work.

storey/sql_driver.py
```diff
@@ -108,23 +114,12 @@ async def _get_specific_fields(self, key: str, table, attributes: List[str]):
...
-    def _update_by_key(self, key, data, table):
-        where_clause = self._get_where_clause(key, table)
-        update_clause = " ,".join([f'{key}="{value}"' for key, value in data.items() if key not in self._primary_key])
-        sql_statement = f"UPDATE {table} SET {update_clause} where {where_clause}"
-        self._sql_connection.execute(sql_statement)
+    def _update_by_key(self, key, data, sql_table):
+        self._sql_connection.execute(
+            db.update(sql_table)
+            .values({getattr(sql_table.c, k): v for k, v in data.items() if k not in self._primary_key})
+            .where(db.and_(getattr(sql_table.c, self._primary_key[i]) == key[i] for i in range(len(self._primary_key))))
```
gtopper, line 121: I see a warning for `self._primary_key[i]` in PyCharm:
```
Expected type 'str', got 'int' instead 
```
Seems like there is a case where that can happen and that it's probably not correct.

storey/sql_driver.py
```diff
@@ -108,23 +114,12 @@ async def _get_specific_fields(self, key: str, table, attributes: List[str]):
...
-
-    def _update_by_key(self, key, data, table):
-        where_clause = self._get_where_clause(key, table)
-        update_clause = " ,".join([f'{key}="{value}"' for key, value in data.items() if key not in self._primary_key])
-        sql_statement = f"UPDATE {table} SET {update_clause} where {where_clause}"
-        self._sql_connection.execute(sql_statement)
+    def _update_by_key(self, key, data, sql_table):
+        self._sql_connection.execute(
+            db.update(sql_table)
+            .values({getattr(sql_table.c, k): v for k, v in data.items() if k not in self._primary_key})
```
gtopper, line 120:
```suggestion
            .values({getattr(sql_table.c, k): k for k in data.keys() if k not in self._primary_key})
```

storey/sql_driver.py
gtopper: Speaking of tests, it would be great if we could add a test that tries to do SQL injection.

gtopper (approved):

PR #440, opened 2023-05-02
ML-3782: Fix overly broad interpretation of features as aggregations

gtopper:


storey/aggregations.py
```diff
@@ -338,7 +339,12 @@ def __init__(
                 raise TypeError("Table can not be string if no context was provided to the step")
             table = kwargs["context"].get_table(table)
         for feature in features:
-            if table.supports_aggregations() and re.match(r".*_[a-z]+_[0-9]+[smhd]$", feature):
+            is_aggregation = False
+            if table.supports_aggregations():
```
assaf758: suggest the below - will deduct one indentation
```suggestion
            if not table.supports_aggregations():
                self._enrich_cols.append(feature)
                continue
```
also i'm not sure we need is_aggregation - simply do
```
if match and is_aggregation_name(match.group(1)):
```

assaf758 (commented):
see my comments inline.

assaf758 (approved):

PR #441, opened 2023-05-08
ML-3603: `CSVSource` reimplementation using `pandas.read_csv`

tomerm-iguazio:
Changed `CSVSource` to work with `pandas.read_csv`, and on top of `DataframeSource`.

There are a few changes that may affect the user:

### CSVSource:

1. When encountering an empty value in the CSV file, this will be read `numpy.nan` instead of `None` previously, as we rely on pandas' `read_csv` method which reads it like that.
2. Previously, it was possible to specify the key or id fields using integers to denote the position of the column. This was never documented and is no longer supported.
3. `header=False` is not longer supported. The parameter itself is now deprecated rather than removed for compatibility, but  header will always be treated as being `True`.

### DataframeSource / ParquetSource:

1. A `ValueError` is now raised if a key or id column does not exist in the headers. Previously, this would cause an error per line.
2. Header with space or special characters (such as `*`) is now valid to use as key or id. Previously, this misbehaved due to the internal use of namedtuples.


storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
+
 
-class CSVSource(_IterableSource, WithUUID):
+class CSVSource(DataframeSource):
     """
     Reads CSV files as input source for a flow.
 
     :parameter paths: paths to CSV files
-    :parameter header: whether CSV files have a header or not. Defaults to False.
+    :parameter header: deprecated! whether CSV files have a header or not. Defaults to False.
```
gtopper: Please be consistent with other deprecated fields.

storey/sources.py
```diff
@@ -699,7 +728,17 @@ def __init__(
...
     ):
+
         kwargs["paths"] = paths
+        if isinstance(paths, str):
+            paths = [paths]
+        self._paths = paths
+        self._build_dict = build_dict
+        if header is False:
+            warnings.warn(
+                "header=False is deprecated",
```
gtopper: Need to let the user what this means. It means that we will treat it as if `header=True`, and that the parameter will be removed in a future version.

storey/sources.py
```diff
@@ -729,212 +763,38 @@ def __init__(
...
+        self._dfs = []
+        for path in self._paths:
+            df = pandas.read_csv(
+                path,
+                header=0,
+                parse_dates=self._dates_indices,
+                date_parser=self._datetime_from_timestamp,
+                storage_options=self._storage_options,
+            )
+            self._validate_fields(df=df, path=path)
```
```suggestion
            self._validate_fields(df, path)
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
-                        key = body[self._key_field]
-                        if key is None:
-                            create_event = False
-                if create_event:
+            columns = list(df.columns)
+            if not df.index.empty and not (len(df.index.names) == 1 and df.index.names[0] is None):
+                df = df.reset_index(drop=False)
+            for body in df.to_dict("records"):
+                key, none_key_column = self._get_key(body=body)
+                if not none_key_column:
```
gtopper: I think this is clearer in code and for the user:
```python
keys = self._get_keys(body)
if any([pd.isna(key) for key in keys]):
    if self.context:
        self.context.logger.error(
            f"Encountered null values in key fields. Key values were: {', '.join(keys)}."
        )
else:
    if self._id_field:
        line_id = body[self._id_field]
    else:
        line_id = self._get_uuid()
    element = self._get_element(body, columns)
    keys = keys[0] if len(keys) == 1 else keys
    event = Event(element, keys, id=line_id)
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
+                for key_field in self._key_field:
+                    single_key = body[key_field]
+                    if pd.isna(single_key):
+                        return single_key, key_field
+                    key.append(single_key)
+            else:
+                key = body[self._key_field]
+                if pd.isna(key):
+                    return key, self._key_field
+        return key, None
```
gtopper: I think this is clearer in code and for the user:
```python
def _get_keys(self, body: dict):
    keys = []
    if self._key_field:
        if isinstance(self._key_field, list):
            for key_field in self._key_field:
                single_key = body[key_field]
                keys.append(single_key)
        else:
            keys.append(body[self._key_field])
    return keys
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
-                                create_event = False
-                                break
-                            key.append(body[key_field])
-                    else:
-                        key = body[self._key_field]
-                        if key is None:
-                            create_event = False
-                if create_event:
+            columns = list(df.columns)
+            if not df.index.empty and not (len(df.index.names) == 1 and df.index.names[0] is None):
```
gtopper: Probably clearer interntion:
```suggestion
            is_df_index_nonempty = not df.index.empty and not (len(df.index.names) == 1 and df.index.names[0] is None)
            if is_df_index_nonempty:
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
             kwargs["id_field"] = id_field
         _IterableSource.__init__(self, **kwargs)
         WithUUID.__init__(self)
+        #  in order to raise exception also for key_field=0
+        if key_field is not None:
+            key_fields = [key_field] if not isinstance(key_field, list) else key_field
+            if any([not isinstance(single_key_field, str) for single_key_field in key_fields]):
+                raise ValueError("key_field should be string/list of strings only!")
```
```suggestion
                raise ValueError("key_field must be a string or list of strings")
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
+        if id_field and not isinstance(id_field, str):
+            raise ValueError("id_field should be in string type only!")
+        self._key_field = key_field
+        self._id_field = id_field
         if isinstance(dfs, pandas.DataFrame):
             dfs = [dfs]
+        if dfs:
+            for df in dfs:
+                self._validate_fields(df=df)
```
```suggestion
                self._validate_fields(df)
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
             kwargs["id_field"] = id_field
         _IterableSource.__init__(self, **kwargs)
         WithUUID.__init__(self)
+        #  in order to raise exception also for key_field=0
+        if key_field is not None:
+            key_fields = [key_field] if not isinstance(key_field, list) else key_field
+            if any([not isinstance(single_key_field, str) for single_key_field in key_fields]):
+                raise ValueError("key_field should be string/list of strings only!")
+        if id_field and not isinstance(id_field, str):
+            raise ValueError("id_field should be in string type only!")
```
```suggestion
            raise ValueError("id_field must be a string")
```

storey/sources.py
```diff
@@ -619,56 +617,87 @@ def __init__(
...
-                        self.context.logger.error(f"For {body} value of key {key_field} is None")
+                        self.context.logger.error(f"value of key {none_key_column} is {key} For {body}")
         return await self._do_downstream(_termination_obj)
 
+    def _validate_fields(self, df, path=""):
+        path_message = f" File path: {path}." if path else ""
+        df = df.reset_index()
+        if self._key_field:
+            key_fields = [self._key_field] if not isinstance(self._key_field, list) else self._key_field
+            missing_keys = list(set(key_fields) - set(df.columns))
```
gtopper: I think it better be something like
```suggestion
missing_keys = [key_field for key_field in key_fields if key_field in df.columns]
```
Because sets are unordered and may mess up the order of the keys and confuse the user.

storey/sources.py
```diff
@@ -1039,6 +899,7 @@ def _init(self):
                 df = self._read_filtered_parquet(path)
             else:
                 df = pandas.read_parquet(path, columns=self._columns, storage_options=self._storage_options)
+            self._validate_fields(df=df, path=path)
```
```suggestion
            self._validate_fields(df, path)
```

tests/test_flow.py
```diff
@@ -3448,7 +3489,9 @@ def test_csv_source_with_none_values():
         "2021-04-21 15:56:53.385444",
     ]
     assert termination_result[1].key == "b"
-    assert termination_result[1].body == ["b", True, None, math.nan, math.nan, None]
+    excepted_result = ["b", True, math.nan, math.nan, math.nan, math.nan]
+    for x, y in zip(termination_result[1].body, excepted_result):
+        assert (isinstance(x, float) and isinstance(y, float) and math.isnan(x) and math.isnan(y)) or x == y
```
gtopper: I think it's better split to an if and two separate asserts, because if this fails, all you'll see is that the condition was false.

storey/sources.py
tomerm-iguazio: added Deprecated also to type_inference.

storey/sources.py
tomerm-iguazio: * not in

storey/sources.py
```diff
@@ -619,56 +617,86 @@ def __init__(
...
+            if is_df_index_nonempty:
+                df = df.reset_index(drop=False)
+            for body in df.to_dict("records"):
+                keys = self._get_keys(body)
+                key_fields = self._key_field if isinstance(self._key_field, list) else [self._key_field]
+                none_keys = [key for key, value in zip(key_fields, keys) if pd.isna(value)]
+                if none_keys:
+                    if self.context:
+                        self.context.logger.error(
+                            f"Encountered null values in key fields. null Key values were:"
```
gtopper, line 660:
```suggestion
                            f"Encountered null values in the following key fields:"
```

```suggestion
                            "Encountered null values in the following key fields:"
```

storey/sources.py
```diff
@@ -619,56 +617,86 @@ def __init__(
...
+
 
-class CSVSource(_IterableSource, WithUUID):
+class CSVSource(DataframeSource):
     """
     Reads CSV files as input source for a flow.
 
     :parameter paths: paths to CSV files
-    :parameter header: whether CSV files have a header or not. Defaults to False.
+    :parameter header: Deprecated! whether CSV files have a header or not. Defaults to False.
```
```suggestion
    :parameter header: Deprecated. whether CSV files have a header or not. Defaults to False.
```

storey/sources.py
```diff
@@ -679,8 +707,8 @@ class CSVSource(_IterableSource, WithUUID):
         datetime.fromisoformat().
     :parameter id_field: the CSV field to be used as the ID for events. May be an int (field index) or string (field
         name) if with_header is True. Defaults to None (random ID will be generated per event).
-    :parameter type_inference: Whether to infer data types from the data (when True), or read all fields in as strings
-        (when False). Defaults to True.
+    :parameter type_inference: Deprecated! Whether to infer data types from the data (when True), or read all fields in
```
```suggestion
    :parameter type_inference: Deprecated. Whether to infer data types from the data (when True), or read all fields in
```

gtopper: So I guess we need a deprecation warning here as well.

tests/test_flow.py
```diff
@@ -3312,7 +3355,7 @@ def test_reader_writer_to_code():
 
     reconstructed_code = flow.to_code()
     print(reconstructed_code)
-    expected = """c_s_v_source0 = CSVSource(paths='mycsv.csv', header=False, build_dict=False, type_inference=True)
+    expected = """c_s_v_source0 = CSVSource(paths='mycsv.csv', header=True, build_dict=False, type_inference=True)
```
gtopper, line 3358: `type_inference` is deprecated.

tests/test_flow.py
```diff
@@ -3448,7 +3489,16 @@ def test_csv_source_with_none_values():
         "2021-04-21 15:56:53.385444",
     ]
     assert termination_result[1].key == "b"
-    assert termination_result[1].body == ["b", True, None, math.nan, math.nan, None]
+    excepted_result = ["b", True, math.nan, math.nan, math.nan, math.nan]
+    for x, y in zip(termination_result[1].body, excepted_result):
```
gtopper, line 3494: If you do it like this, you need to check the length as well. Otherwise, an empty result will also pass this test.

tests/test_flow.py
```diff
@@ -379,14 +403,38 @@ def test_csv_reader_none_in_keyfield_should_send_error_log():
...
-            CSVSource("tests/test-none-in-keyfield.csv", header=True, key_field="k", context=context),
+            CSVSource("tests/test-none-in-keyfield.csv", key_field="k", context=context),
         ]
     ).run()
 
     controller.await_termination()
 
     assert "error" == logger.logs[0][0]
-    assert "value of key k is None" in logger.logs[0][1][0]
+    assert "Encountered null values in the following key fields:: k" in logger.logs[0][1][0]
```
gtopper, line 413: Please fix this one. It got messed up somehow.

storey/sources.py
```diff
@@ -619,56 +617,86 @@ def __init__(
...
+            if is_df_index_nonempty:
+                df = df.reset_index(drop=False)
+            for body in df.to_dict("records"):
+                keys = self._get_keys(body)
+                key_fields = self._key_field if isinstance(self._key_field, list) else [self._key_field]
+                none_keys = [key for key, value in zip(key_fields, keys) if pd.isna(value)]
+                if none_keys:
+                    if self.context:
+                        self.context.logger.error(
+                            f"Encountered null values in the following key fields::"
```
```suggestion
                            f"Encountered null values in the following key fields:"
```

tests/test_flow.py
```suggestion
    assert "Encountered null values in the following key fields: k" in logger.logs[0][1][0]
```

storey/sources.py
```diff
@@ -689,17 +717,34 @@ class CSVSource(_IterableSource, WithUUID):
...
+        self._build_dict = build_dict
+        if header is False:
+            warnings.warn(
+                "header=False is deprecated, will be treated as header=True."
+                " The parameter will be removed in a future version.",
+                DeprecationWarning,
+            )
+        if type_inference is not None:
+            warnings.warn(
+                "type_inference is deprecated, and will not be treated."
```
```suggestion
                "type_inference is deprecated, and does not have any effect."
```

tomerm-iguazio: I change it to be treated as True

gtopper (approved):
Congrats 👏

PR #442, opened 2023-05-24
ML-3880: Fix key extraction in `DataframeSource`

tomerm-iguazio:
Fix case where the `key_field` argument is a list of size 1.

### DIFF
--- a/storey/sources.py
+++ b/storey/sources.py
@@ -666,7 +666,7 @@ class DataframeSource(_IterableSource, WithUUID):
                     else:
                         line_id = self._get_uuid()
                     element = self._get_element(body, columns)
-                    keys = keys[0] if len(keys) == 1 else (None if not keys else keys)
+                    keys = None if not keys else keys
                     event = Event(element, keys, id=line_id)
                     await self._do_downstream(event)
         return await self._do_downstream(_termination_obj)
--- a/storey/sources.py
+++ b/storey/sources.py
@@ -666,11 +666,10 @@ class DataframeSource(_IterableSource, WithUUID):
                     else:
                         line_id = self._get_uuid()
                     element = self._get_element(body, columns)
-                    keys = (
-                        keys[0]
-                        if len(keys) == 1 and not isinstance(self._key_field, list)
-                        else (None if not keys else keys)
-                    )
+                    if len(keys) == 0:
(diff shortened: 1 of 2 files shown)


storey/sources.py
```diff
@@ -666,7 +666,11 @@ async def _run_loop(self):
                     else:
                         line_id = self._get_uuid()
                     element = self._get_element(body, columns)
-                    keys = keys[0] if len(keys) == 1 else (None if not keys else keys)
+                    keys = (
+                        keys[0]
+                        if len(keys) == 1 and not isinstance(self._key_field, list)
+                        else (None if not keys else keys)
+                    )
```
gtopper: Can we simplify and clarify as follows?
```suggestion
                    if len(keys) == 0:
                        keys = None
                    elif not isinstance(self._key_field, list):
                        keys = keys[0]
```

tests/test_flow.py
```diff
@@ -472,7 +472,10 @@ def test_indexed_dataframe_source():
     assert termination_result == expected
 
 
-def test_dataframe_source_with_metadata():
+@pytest.mark.parametrize(
+    "key_field ,key1, key2", [("my_key", "key1", "key2"), (["my_key"], ["key1"], ["key2"])]  # Test case 2
```
```suggestion
    "key_field,key1, key2", [("my_key", "key1", "key2"), (["my_key"], ["key1"], ["key2"])]
```

tests/test_flow.py
```diff
@@ -490,16 +493,19 @@ def test_dataframe_source_with_metadata():
...
-            key="key2",
+            key=key2,
             id="id2",
         ),
     ]
-    assert termination_result == expected
+    assert len(termination_result) == len(expected)
+    for result, expected_event in zip(termination_result, expected):
+        assert result == expected_event
+        assert result.key == expected_event.key  # __eq__ method does not check key.
```
gtopper: Maybe better to check like this, so that we get a more useful error message (not e.g. `AssertionError: 1 != 2`)
```suggestion
    assert termination_result == expected
    assert list(map(lambda event: event.key, termination_result)) == [key1, key2]
```

tomerm-iguazio: great idea!

gtopper (approved):

PR #443, opened 2023-06-08
ML-3903: `QueryByKey` - input aliasing

davesh0812:
Associated with [ML-3903](https://jira.iguazeng.com/browse/ML-3903), https://github.com/mlrun/mlrun/pull/3672.

Open the option to aliasing he incoming event, part of the implementation of the online merger that using the relation properties.


gtopper (commented):
There already exists a step for this, called `Rename`. Why not just place that right before `QueryByKey` as needed?

PR #444, opened 2023-06-08
ML-3895: Fix date parser

tomerm-iguazio:
Date parser should handle `na` values in `pandas` as `None` - like `Nan`/ `Nat`(not a time).

The test make sure we get a `NaT` for empty values.


storey/sources.py
```diff
@@ -793,7 +793,7 @@ def _init(self):
             self._dfs.append(df)
 
     def _datetime_from_timestamp(self, timestamp):
-        if timestamp == "" or timestamp is None:
+        if timestamp == "" or pd.isna(timestamp):
```
gtopper, line 796: It turns out that the tests pass even without this fix, so I looked into it in a debugger and it looks like this function always raises an error, because the argument it gets is not what it expects. Need to fix that, because right now it seems to be doing nothing. And fix the tests to fail before the actual fix.

gtopper (changes requested):

storey/sources.py
gtopper: Apparently the error raised is part of the flow of `pandas.read_csv`. It tries different types of arguments one after the other. So that part is ok. Only need to understand the first part of my previous comment, about test coverage. According to @tomerm-iguazio, he can only reproduce this with Azure, which is worrisome because where the CSV was read from shouldn't have an effect on how we read it.

PR #445, opened 2023-06-13
ML-3895: Fix date parser

tomerm-iguazio:
Date parser should handle `na` values in `pandas` as `None` - like `Nan`/ `Nat`(not a time).

Additionally, modify `test_csv_source_with_none_values` to attempt parsing an empty date. (The code previously mentioned the date column as a string).


gtopper (approved):
Looks good now. Original PR at #444.

PR #449, opened 2023-07-17
ML-4171: Redis target: move to one redis-key per entity record

alxtkr77:


storey/redis_driver.py
```diff
@@ -526,25 +547,35 @@ async def _load_aggregates_by_key(self, container, table_path, key):
...
+            values = {}
+            while True:
+                cursor, v = await RedisDriver.asyncify(self.redis.hscan)(
+                    redis_key, cursor, match=f"{self.AGGREGATION_ATTRIBUTE_PREFIX}*"
+                )
+                values.update(v)
+                if cursor == 0:
+                    break
+        except redis.ResponseError as e:
+            raise RedisError(f"Failed to get key {redis_key}. Response error was: {e}")
```
gtopper, line 536: You can add `from e`:
```suggestion
            raise RedisError(f"Failed to get key {redis_key}. Response error was: {e}") from e
```
In theory that means you can remove the "Response error was" part, but in practice, chained exceptions are often not printed where you might expect them to be, so up to you whether to drop that or not.

storey/redis_driver.py
```diff
@@ -526,25 +547,35 @@ async def _load_aggregates_by_key(self, container, table_path, key):
...
-        for aggr_key in all_aggr_keys:
-            aggr_key = RedisDriver.convert_to_str(aggr_key)
-            value = await RedisDriver.asyncify(self.redis.hget)(aggr_key_prefix, aggr_key)
+        redis_key = self._static_data_key(redis_key_prefix)
+        try:
+            cursor = 0
+            values = {}
+            while True:
+                cursor, v = await RedisDriver.asyncify(self.redis.hscan)(
+                    redis_key, cursor, match=f"{self.AGGREGATION_ATTRIBUTE_PREFIX}*"
```
gtopper: Maybe we can parallelize these calls? i.e. with `asyncio.create_task` and `asyncio.gather`.

storey/redis_driver.py
```diff
@@ -526,25 +547,35 @@ async def _load_aggregates_by_key(self, container, table_path, key):
...
-        aggr_key_prefix = f"{redis_key_prefix}:{RedisDriver.AGGREGATION_ATTRIBUTE_PREFIX}"
-        all_aggr_keys = self.redis.hkeys(aggr_key_prefix)
-        for aggr_key in all_aggr_keys:
-            aggr_key = RedisDriver.convert_to_str(aggr_key)
-            value = await RedisDriver.asyncify(self.redis.hget)(aggr_key_prefix, aggr_key)
+        redis_key = self._static_data_key(redis_key_prefix)
+        try:
+            cursor = 0
+            values = {}
+            while True:
```
gtopper, line 530: This pattern appears 3 times by my count (not including tests). I think it can be extracted out.

gtopper (commented):
Reviewed incidental aspect of this PR – not the redis changes themselves.

storey/redis_driver.py
alxtkr77: We cannot, since 'cursor' is returned and used in following call

storey/redis_driver.py
alxtkr77: ok

storey/redis_driver.py
alxtkr77: will refactor

integration/test_flow_integration.py
```diff
@@ -111,7 +111,16 @@ def _get_redis_kv_all_attrs(setup_teardown_test: ContextForTests, key: str):
     hash_key = RedisDriver.make_key("storey-test:", table_name, key)
     redis_key = RedisDriver._static_data_key(hash_key)
     redis_fake_server = setup_teardown_test.redis_fake_server
-    values = get_redis_client(redis_fake_server=redis_fake_server).hgetall(redis_key)
+
+    cursor = 0
+    values = {}
+    while True:
+        cursor, v = get_redis_client(redis_fake_server=redis_fake_server).hscan(
+            redis_key, cursor, match=f"[^{chr(0x1)}]*"
```
assaf758: use INTERNAL_FIELD_PREFIX

assaf758: use  INTERNAL_FIELD_PREFIX

assaf758 (changes requested):

assaf758 (approved):

PR #455, opened 2023-08-03
ML-2296: Add support for datastore profiles

alxtkr77:
[ML-2296](https://jira.iguazeng.com/browse/ML-2296)

### DIFF
--- a/integration/test_filesystems_integration.py
+++ b/integration/test_filesystems_integration.py
@@ -22,6 +22,7 @@ import aiohttp
 import pandas as pd
 import pytest
 import v3io
+from fsspec.implementations.local import LocalFileSystem
 
 from integration.integration_test_utils import V3ioHeaders, _generate_table_name
 from storey import (
@@ -38,7 +39,7 @@ from storey import (
     build_flow,
 )
 from storey.dtypes import V3ioError
-from storey.utils import get_remaining_path
+from storey.utils import get_remaining_path, url_to_file_system
 
 
 @pytest.fixture()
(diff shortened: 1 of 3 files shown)


storey/utils.py
```diff
@@ -138,6 +138,12 @@ def get_remaining_path(url):
 
 def url_to_file_system(url, storage_options):
     scheme, remaining_path = get_remaining_path(url)
+    if url.startswith("ds://"):
+        parsed_url = urlparse(url)
+        if parsed_url.password:
+            scheme = parsed_url.password
```
gtopper, line 144: Can we add a test to show what this looks like?

storey/utils.py
```diff
@@ -138,6 +138,12 @@ def get_remaining_path(url):
...
+        else:
+            raise ValueError("Datastore profile URL expecting to have underlying scheme embedded as a password")
```
```suggestion
            raise ValueError("Datastore profile URL is expected to have underlying scheme embedded as password")
```

storey/utils.py
alxtkr77: done

integration/test_filesystems_integration.py
```diff
@@ -672,3 +673,10 @@ def test_get_path_utils():
     schema, path = get_remaining_path(url)
     assert path == "mycontainer/path/to/object.csv"
     assert schema == "wasbs"
+
+
+def test_ds_get_path_utils():
```
gtopper: I think this is basically a unit test, so better move it there.

alxtkr77: Hmm, you probably right , and test_get_path_utils() should be moved as well. But there is no unit tests for the utils.py module. Where would you put one?

gtopper: Feel free to create a `test_utils.py` file under `tests/` and move anything that's appropriate to move in there.

gtopper (approved):
LGTM

PR #461, opened 2023-09-18
ML-2956: Make all aggregations return `NaN` when there is no data except for count

alxtkr77:
https://jira.iguazeng.com/browse/ML-2956


storey/table.py
```diff
@@ -1408,6 +1408,7 @@ def __init__(
         self.name = name
         self._explicit_raw_aggregations = explicit_raw_aggregations
         self._hidden_raw_aggregations = hidden_raw_aggregations
+        self._hidden_raw_aggregations.append("count")
```
gtopper, line 1428: As we discussed, I think it would be better to change `SumValue`, potentially decoupling initial values from default values if necessary.

alxtkr77: Actually you cannot do that. Min and Max are handled at table.py:1658 :
`                    if value == math.inf or value == -math.inf:
                        value = math.nan
`
For `sum` you cannot do this trick, because the value '0' is a valid value

alxtkr77: But you raised a good point - at the bug description the min and max are set to `inf` and not `nan`, means there is another path that should be handled as well. Looking into it

storey/table.py
```diff
@@ -852,7 +861,11 @@ def get_features(self, timestamp, windows=None):
         # In case our pre aggregates already have the answer
         for win in windows:
             result[f"{self.name}_{self.aggregation}_{win[1]}"] = self._current_aggregate_values[win].value
-
+            if (
+                self.aggregation != "count"
+                and (count_features.get(f"{self.name}_count_{win[1]}", 0) if count_features else 1) == 0
```
gtopper: Isn't that more simply
```suggestion
                and count_features
                and count_features.get(f"{self.name}_count_{win[1]}", 0) == 0
```

storey/table.py
```diff
@@ -852,7 +861,11 @@ def get_features(self, timestamp, windows=None):
...
+            ):
+                result[f"{self.name}_{self.aggregation}_{win[1]}"] = math.nan
```
gtopper: And more generally to also avoid duplication (refers to 1 line above also).
```suggestion
            if (
                self.aggregation != "count"
                and count_features
                and count_features.get(f"{self.name}_count_{win[1]}", 0) == 0
            ):
                value = math.nan
            else:
                value = self._current_aggregate_values[win].value
            result[f"{self.name}_{self.aggregation}_{win[1]}"] = value
```

storey/table.py
gtopper: I think this line should move to the caller, and there should be deduplication.

gtopper: I still think this would better be done by reworking the `AggregationValue` hierarchy, decoupling "initial value" from "default value", but it's not that straightforward and I dont' know how much more effort we want to put into this issue.

gtopper (approved):
Looks good, but please run some mlrun feature store system tests on it before we merge and release this.

PR #471, opened 2023-11-23
Bump pyarrow version to 14.0.1

alxtkr77:
Before upgrading pyarrow to version 14.0.1, we raised the minimum supported Python version in our continuous integration (CI) from Python 3.7 to Python 3.9. This change in the CI configuration automatically upgraded the pandas library to version 2.xxx because pandas 2.0 is supported starting with Python 3.9. As a result, we had to make adjustments to our tests to accommodate the new time resolutions introduced in pandas 2.0.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ v3io~=0.5.14
 # and 1.5.* due to https://github.com/pandas-dev/pandas/issues/49203
 pandas>=1, !=1.5.*, <3
 numpy>=1.16.5,<1.23
+# <15 is just a safeguard - no tests performed with pyarrow higher then 14
 pyarrow>=1,<15
 v3io-frames~=0.10.3
 v3iofs~=0.1.17
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ v3io~=0.5.14
 # and 1.5.* due to https://github.com/pandas-dev/pandas/issues/49203
 pandas>=1, !=1.5.*, <3
 numpy>=1.16.5,<1.23
-# <15 is just a safeguard - no tests performed with pyarrow higher then 14
+# <15 is just a safeguard - no tests performed with pyarrow higher than 14
 pyarrow>=1,<15
 v3io-frames~=0.10.3
 v3iofs~=0.1.17
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,7 @@ v3io~=0.5.14
 # and 1.5.* due to https://github.com/pandas-dev/pandas/issues/49203
 pandas>=1, !=1.5.*, <3
 numpy>=1.16.5,<1.23
-# pyarrow 13 and over cause test failures
-pyarrow>=1,<13
+pyarrow>=1,<15
 v3io-frames~=0.10.3
 v3iofs~=0.1.17
 xxhash>=1
--- a/tests/test_flow.py
+++ b/tests/test_flow.py
@@ -3235,6 +3235,44 @@ def test_csv_reader_parquet_write_microsecs(tmpdir):
     assert_frame_equal(expected, read_back_df, check_dtype=version.parse(pd.__version__) < version.parse("2.0.0"))
 
 
+def test_csv_reader_parquet_write_nanosecs_truncation(tmpdir):
+    out_file = f"{tmpdir}/test_csv_reader_parquet_write_nanosecs_{uuid.uuid4().hex}/"
+    columns = ["k", "t"]
+
+    time_format = "%d/%m/%Y %H:%M:%S.%f"
+    controller = build_flow(
+        [
+            CSVSource(
+                "tests/test-with-timestamp-nanosecs.csv",
+                header=True,
+                key_field="k",
+                parse_dates="t",
+                timestamp_format=time_format,
+            ),
+            ParquetTarget(
+                out_file,
+                columns=columns,
+                partition_cols=["$year", "$month", "$day", "$hour"],
+                max_events=2,
+            ),
+        ]
+    ).run()
+
+    expected = pd.DataFrame(
+        [
+            ["m1", datetime.strptime("15/02/2020 02:03:04.123456", time_format)],
+            ["m2", datetime.strptime("16/02/2020 02:03:04.123456", time_format)],
+        ],
+        columns=columns,
+    )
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -4,8 +4,7 @@ v3io~=0.5.14
 # and 1.5.* due to https://github.com/pandas-dev/pandas/issues/49203
 pandas>=1, !=1.5.*, <3
 numpy>=1.16.5,<1.23
-# pyarrow 13 and over cause test failures
-pyarrow>=1,<13
+pyarrow>=1,<15
```
gtopper, line 8: Maybe just mention that this is a safeguard. I.e. that this limit isn't because there's something wrong with pyarrow 15 (which doesn't even exist yet).

tests/test_flow.py
```diff
@@ -2721,7 +2722,9 @@ def test_write_to_parquet_partition_by_hash(tmpdir):
     read_back_df = pd.read_parquet(out_file, columns=columns)
     read_back_df.sort_values("my_int", inplace=True)
     read_back_df.reset_index(drop=True, inplace=True)
-    assert read_back_df.equals(expected), f"{read_back_df}\n!=\n{expected}"
+    # with the introduction of s, ms, us time resolutions in pandas-2.0, the dtype of the parquet data
```
gtopper, line 2725: I'm pretty sure the time resolution exists in pandas 1 too.

tests/test_flow.py
```diff
@@ -3225,43 +3230,9 @@ def test_csv_reader_parquet_write_microsecs(tmpdir):
...
-
-def test_csv_reader_parquet_write_nanosecs(tmpdir):
-    out_file = f"{tmpdir}/test_csv_reader_parquet_write_nanosecs_{uuid.uuid4().hex}/"
-    columns = ["k", "t"]
-
-    time_format = "%d/%m/%Y %H:%M:%S.%f"
-    controller = build_flow(
-        [
-            CSVSource(
-                "tests/test-with-timestamp-nanosecs.csv",
```
gtopper: Naturally, you need to delete the resource if you delete the test.

tests/test_flow.py
```diff
@@ -3225,43 +3230,9 @@ def test_csv_reader_parquet_write_microsecs(tmpdir):
     controller.await_termination()
     read_back_df = pd.read_parquet(out_file, columns=columns)
 
-    assert read_back_df.equals(expected), f"{read_back_df}\n!=\n{expected}"
-
-
-def test_csv_reader_parquet_write_nanosecs(tmpdir):
```
gtopper, line 3231: @assaf758, are we ok with dropping support for this? This was actually requested a couple of years back.

assaf758: Gal, can you elaborate more - what was requested exactly, (and by whom, for what reason, if you remember)
Alex, can you explain why you are removing the test?

requirements.txt
assaf758: Actually I would to revisit our practice of setting an upper limit on some later major version.
With the current approach, frequently we will get a request to upgrade from the field (like now) before we understand there is an issue in supporting the later version. In some cases - the user will need to wait for the next mlrun version.
On the other hand if we remove the upper limit, chances are we will hit the bug before the user, and the user will have an mlrun that support the new version by the time it needs it.
what do you think?

gtopper: Only not sure it's true that we'll hit this before the user... And if we do, we won't be able to add the limit after the fact. Without the limit, a new release could immediately introduce an issue for the user, as well as break CI (which wouldn't be so bad, but it's sometimes hard to realize that CI was broken by a new version of a dependency and not by a PR or sporadic failure or env issue).

Does the user actually need to wait for a new release? They should be able to install whatever library they want and only get a warning regarding the incompatibility.

tests/test_flow.py
gtopper: Ability to read CSV with nanosecond precision, iirc. I think it was related to work done by OrZ (like I said, a few years back). I don't remember any details beyond that.

requirements.txt
assaf758: I see, I guess will make more sense once we have dependencies freeze.

tests/test_flow.py
alxtkr77: https://levelup.gitconnected.com/welcoming-pandas-2-0-194094e4275b

tests/test_flow.py
alxtkr77: Will double check

requirements.txt
alxtkr77: @gtopper , what about padnas<3? Also a comment?  Should be more generic approach

gtopper: In principle you're right @alxtkr77. In practice, pyarrow releases are much more frequent, which is why the limit is there. A general guideline would be great, but maybe the dependency freeze takes care of it anyway.

tests/test_flow.py
gtopper: Turns out it was supported in numpy `datetime` but not pandas `Timestamp`.

gtopper: https://numpy.org/doc/stable/reference/arrays.datetime.html

tests/test_flow.py
alxtkr77: Since Python doesn't inherently support nanosecond resolution, the purpose of this test is to assess Storey's capacity to read values in nanosecond format from a CSV file and convert them to microseconds. I will reinstate the test and update its name to more accurately describe this capability.

requirements.txt
alxtkr77: Will add as requested

requirements.txt
```diff
@@ -4,8 +4,8 @@ v3io~=0.5.14
 # and 1.5.* due to https://github.com/pandas-dev/pandas/issues/49203
 pandas>=1, !=1.5.*, <3
 numpy>=1.16.5,<1.23
-# pyarrow 13 and over cause test failures
-pyarrow>=1,<13
+# <15 is just a safeguard - no tests performed with pyarrow higher then 14
```
gtopper:
```suggestion
# <15 is just a safeguard - no tests performed with pyarrow higher than 14
```

gtopper (approved):

PR #478, opened 2023-12-17
Fix `date_parser` deprecation warning

gtopper:
```
The argument 'date_parser' is deprecated and will be removed in a future version. Please use 'date_format' instead, or read your data in as 'object' dtype and then call 'to_datetime'.
```

### DIFF
--- a/storey/sources.py
+++ b/storey/sources.py
@@ -26,6 +26,7 @@ from collections import defaultdict
 from datetime import datetime, timezone
 from typing import Callable, Coroutine, Iterable, List, Optional, Union
 
+import packaging.version
 import pandas
 import pandas as pd
 import pyarrow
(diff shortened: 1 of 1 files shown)


storey/sources.py
```diff
@@ -944,20 +944,12 @@ def _init(self):
                 path,
                 header=0,
                 parse_dates=self._dates_indices,
-                date_parser=self._datetime_from_timestamp,
+                date_format=self._timestamp_format,
```
assaf758: according to [this](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) date_format is new in pandas 2.0.0, but what if a user is running pandas <2.0 ?

gtopper: Thanks!

assaf758 (approved):

PR #479, opened 2023-12-20
`ParquetTarget`: ignore batching parameters in single file mode

gtopper:
[ML-5119](https://jira.iguazeng.com/browse/ML-5119)


tomerm-iguazio (approved):
Looks good:)

PR #482, opened 2023-12-26
Fix reuse of various steps

gtopper:
[ML-5351](https://jira.iguazeng.com/browse/ML-5351)

### DIFF
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -1077,7 +1077,7 @@ class KafkaTarget(Flow, _Writer):
             record = self._event_to_writer_entry(event)
             if self._full_event:
                 record = Event.wrap_for_serialization(event, record)
-            record = json.dumps(record).encode("UTF-8")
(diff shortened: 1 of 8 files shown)


storey/targets.py
```diff
@@ -1124,19 +1123,24 @@ def __init__(
...
         self._write_missing_fields = False
 
     def _init(self):
         Flow._init(self)
         _Writer._init(self)
+        if self._table_name:
+            if not self.context:
+                raise TypeError("Table can not be string if no context was provided to the step")
+            self._table = self.context.get_table(self._table_name)
+            self._closeables = [self._table]
```
gtopper: Will add a test to cover this use case (reuse of `NoSqlTarget`.

gtopper: Reverted that change for now. It's more involved and risky than the rest of this PR, and doesn't relate to the ticket in question.

assaf758 (approved):

PR #498, opened 2024-01-24
[Driver] Fix V3IO driver handling of single quote in strings where double quote is not present

tomerm-iguazio:
If both are present, the UpdateItem operation will still fail with an expression syntax error.

[ML-5576](https://jira.iguazeng.com/browse/ML-5576)

### DIFF
--- a/integration/test_flow_integration.py
+++ b/integration/test_flow_integration.py
@@ -1120,9 +1120,9 @@ def test_writing_quote_in_value(setup_teardown_test):
         ]
     ).run()
     controller.await_termination()
-    assert get_key_all_attrs_test_helper(setup_teardown_test, "0", keys) == {'color': "gre'en", 'num': 0}
-    assert get_key_all_attrs_test_helper(setup_teardown_test, "1", keys) == {'color': 'bl"ue', 'num': 1}
-    assert get_key_all_attrs_test_helper(setup_teardown_test, "2", keys) == {'color': 'red', 'num': 2}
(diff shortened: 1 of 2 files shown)


alxtkr77 (approved):

integration/test_flow_integration.py
```diff
@@ -1107,6 +1107,24 @@ def test_writing_int_key(setup_teardown_test):
...
+    controller = build_flow(
+        [
+            DataframeSource(df, key_field="num"),
+            NoSqlTarget(table),
+        ]
+    ).run()
+    controller.await_termination()
+    assert get_key_all_attrs_test_helper(setup_teardown_test, "0", keys) == {"color": "gre'en", "num": 0}
+    assert get_key_all_attrs_test_helper(setup_teardown_test, "1", keys) == {"color": 'bl"ue', "num": 1}
+    assert get_key_all_attrs_test_helper(setup_teardown_test, "2", keys) == {"color": "red", "num": 2}
```
assaf758: suggest adding a test that "gre'en" + 'bl"ue' fails.

storey/drivers.py
```diff
@@ -549,6 +549,10 @@ def _build_simplified_feature_store_request(self, aggregation_element):
     @staticmethod
     def _convert_python_obj_to_expression_value(value):
         if isinstance(value, str):
+            # in order to handle ' in value.
+            # may impact performance
```
assaf758: please remove comment on performance (already discussed).
please add a comment that a string with both " and ' are not supported

assaf758 (changes requested):

assaf758 (approved):

PR #511, opened 2024-04-11
Add `ConcurrentExecution` step

gtopper:
[ML-6015](https://iguazio.atlassian.net/browse/ML-6015)

Based on the preexisting internal `_ConcurrentJobExecution` class. With support for three concurrency mechanisms: asyncio, threading, and multiprocessing.

[ML-6015]: https://iguazio.atlassian.net/browse/ML-6015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


storey/flow.py
```diff
@@ -804,7 +806,8 @@ def __init__(self, max_in_flight=None, retries=None, backoff_factor=None, **kwar
         self.retries = retries
         self.backoff_factor = backoff_factor
 
-        self._queue_size = max_in_flight - 1 if max_in_flight else 8
+        max_in_flight = max_in_flight or 8
```
alxtkr77: why not self.max_in_flight = max_in_flight or 8 ?

alxtkr77 (commented):
Aside from the issue with serializable context, everything else in the pull request looks good. Do we want to address this issue within the current pull request? It seems that the 'pass_context' flag may not be the best solution for the problem.

storey/flow.py
gtopper: Because it would be an unused attribute.

gtopper:
> Do we want to address this issue within the current pull request?

To make mlrun context serializable isn't something we can do in storey anyway.

>  It seems that the 'pass_context' flag may not be the best solution for the problem

Why is it the flag's fault? :) Do you have some change in mind?

storey/flow.py
alxtkr77: It's used in your new ConcurrentExecution class

gtopper: Right. Fixed now.

alxtkr77 (approved):

assaf758:
> > Do we want to address this issue within the current pull request?
> 
> To make mlrun context serializable isn't something we can do in storey anyway.
> 
> > It seems that the 'pass_context' flag may not be the best solution for the problem
> 
> Why is it the flag's fault? :) Do you have some change in mind?

Can you check if indeed https://github.com/uqfoundation/dill can serialize lambdas?

gtopper:
@assaf758, I don't believe multiprocessing allows for replacing pickle with dill or anything else. Do you have a solution in mind?

gtopper:
@assaf758, it can, but it's not obvious how to use that to solve the problem. I'll try to find a way.

gtopper:
It turns out that mlrun `GraphContext` contains the entire graph, which even dill fails to serialize. Ironically, it's there for `ParallelRun`.

assaf758 (approved):

PR #512, opened 2024-04-15
Update black formatter

gtopper:
To resolve https://github.com/mlrun/storey/security/dependabot/13

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,5 +1,5 @@
 build~=0.1.0
-black~=22.0
+black~=24.3
 flake8~=5.0
 flake8-bugbear~=22.9
 isort~=5.7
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -1,5 +1,5 @@
 build~=0.1.0
-black~=22.0
+black~=24.3
```
jond01, line 2: The newer 24.4.0 is available:
https://pypi.org/project/black/24.4.0/

jond01 (approved):
Looks good :)
I would consider moving to Ruff in the longer term

dev-requirements.txt
gtopper: Nice. The lower bound is only there because of the vulnerability though. It's not supposed to be set to the latest version.

PR #514, opened 2024-04-18
[Sources] Add filters to `ParquetSource`

tomerm-iguazio:
1) Added support for filters parquetSource.
2) Added validators to wrong usage.
3) Added tests for both scenarios: one with the combine filter_column(timestamp filter) parameter and one without combining it.

[ML-6163](https://iguazio.atlassian.net/browse/ML-6163)

[ML-6163]: https://iguazio.atlassian.net/browse/ML-6163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/test_flow.py
```diff
@@ -4580,3 +4581,53 @@ def test_empty_filter_result():
...
+            source_kwargs["filter_column"] = "my_time"
+            expected_df = pd.DataFrame([tel_aviv_data[0]], columns=columns)
+        expected_df.set_index("my_string", inplace=True)
+
+        controller = build_flow([ParquetSource(temp_dir, **source_kwargs), ReduceToDataFrame(index="my_string")]).run()
+        read_back_result = controller.await_termination()
+        pd.testing.assert_frame_equal(read_back_result, expected_df)
+
+
+def test_filters_validators():
```
gtopper: Please break up to separate tests with descriptive names. Preferably ones that don't contain the word "validator", since there is no such abstraction in storey.

tests/test_flow.py
```diff
@@ -4580,3 +4581,53 @@ def test_empty_filter_result():
...
+    with pytest.raises(ValueError, match="Cannot use the same column as both the filter_column and in the filters."):
+        ParquetSource(
+            "/my_dir", filters=[("start_time", ">", pd.Timestamp("2020-12-31 14:00:00"))], filter_column="start_time"
+        )
+
+    with pytest.raises(ValueError, match="ParquetSource supports filters only as a list of tuples!"):
+        ParquetSource(
+            "/my_dir", filters=[[("city", "=", "Tel Aviv")], [("age", ">=", "40")]], filter_column="start_time"
```
```
Expected type 'Optional[list[tuple]]', got 'list[list[tuple[str, str, str]]]' instead 
```

storey/sources.py
```diff
@@ -1009,7 +1019,23 @@ def __init__(
 
             if filter_column is None:
                 raise TypeError("Filter column is required when passing start/end filters")
-
+        filters = filters if filters else []
+
+        if not all(isinstance(item, Tuple) for item in filters):
+            raise ValueError(f"ParquetSource supports filters only as a list of tuples!" f"Current filters: {filters}")
```
gtopper: 1. Not sure why this is broken up to two separate strings.
2. No shouting at the user please :)

storey/sources.py
```diff
@@ -1009,7 +1019,23 @@ def __init__(
 
             if filter_column is None:
                 raise TypeError("Filter column is required when passing start/end filters")
-
+        filters = filters if filters else []
+
+        if not all(isinstance(item, Tuple) for item in filters):
```
```suggestion
        if not all(isinstance(item, tuple) for item in filters):
```

storey/sources.py
```diff
@@ -1009,7 +1019,23 @@ def __init__(
 
             if filter_column is None:
                 raise TypeError("Filter column is required when passing start/end filters")
-
+        filters = filters if filters else []
```
```suggestion
        filters = filters or []
```

storey/sources.py
```diff
@@ -1009,7 +1019,23 @@ def __init__(
...
+            if filter_column and filter_column in filters_tuple:
+                raise ValueError(
+                    f"Cannot use the same column as both the filter_column and in the filters."
+                    f" Column: {filter_column}."
+                )
+            for filter_argument in filters_tuple:
+                if isinstance(filter_argument, (pd.Timestamp, datetime)):
+                    raise ValueError(
+                        f"Cannot use datetime values in filters. For these types, "
+                        f"use start_filter and end_filter parameters. Current filters: {filters}"
```
gtopper: Why not? There may be timestamp columns other than the filter column.

storey/sources.py
```diff
@@ -993,6 +1002,7 @@ def __init__(
         start_filter: Optional[datetime] = None,
         end_filter: Optional[datetime] = None,
         filter_column: Optional[str] = None,
+        filters: Optional[list[tuple]] = None,
```
gtopper: This is confusing because we already have a parameter called `filter_column`. Better name this `additional_filters`.

storey/sources.py
```diff
@@ -1018,6 +1044,7 @@ def __init__(
         self._end_filter = end_filter
         self._filter_column = filter_column
         self._storage_options = kwargs.get("storage_options")
+        self.filters = filters
```
```suggestion
        self._filters = filters
```

storey/sources.py
```diff
@@ -982,8 +987,12 @@ class ParquetSource(DataframeSource):
     :parameter end_filter: datetime. If not None, the results will be filtered by partitions
         'filter_column' <= end_filter. Default is None.
     :parameter filter_column: Optional. if not None, the results will be filtered by this column and before and/or after
+                datetime column.
     :param key_field: column to be used as key for events. can be list of columns
     :param id_field: column to be used as ID for events.
+    :param filters: other filters to use while reading the parquet.
+                    Supported operators: '=', '>=', '<=', '>', '<'.
+                    Example: ('Product', '=', 'Computer')]
```
gtopper: Should probably reference the fact that these are [pyarrow filters](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html).

storey/sources.py
```diff
@@ -1009,7 +1019,23 @@ def __init__(
...
-
+        filters = filters if filters else []
+
+        if not all(isinstance(item, Tuple) for item in filters):
+            raise ValueError(f"ParquetSource supports filters only as a list of tuples!" f"Current filters: {filters}")
+
+        for filters_tuple in filters:
+            if filter_column and filter_column in filters_tuple:
+                raise ValueError(
+                    f"Cannot use the same column as both the filter_column and in the filters."
```
```suggestion
                    f"Cannot use the same column as both filter_column and in filters."
```

storey/utils.py
```diff
@@ -344,3 +345,25 @@ def find_filters(partitions_time_attributes, start, end, filters, filter_column)
...
     )
+
+
+def get_combined_filters(datetime_filters=None, filters=None):
+    datetime_filters = copy.deepcopy(datetime_filters) or [[]]
+    filters = copy.deepcopy(filters) or []
+    or_filter_exists = False
+    filter_tuple_exists = False
+
+    for or_filter in datetime_filters:
```
gtopper: Wouldn't a member of `datetime_filters` be a `datetime_filter`? Why is this called an `or_filter`?

storey/utils.py
```diff
@@ -344,3 +345,25 @@ def find_filters(partitions_time_attributes, start, end, filters, filter_column)
...
+                if or_filter:
+                    or_filter_exists = True
+                    or_filter.append(tuple_filter)
+            elif or_filter:
+                or_filter_exists = True
+
+    #  datatime filters doesn't exists or [[],[]] :
+    if not or_filter_exists and filter_tuple_exists:
+        return [filters]
+    return datetime_filters
```
gtopper: Can't all of this just be replaced with
```python
    if datetime_filters and filters:
        datetime_filters[0] += filters
    return datetime_filters
```
?

storey/utils.py
```diff
@@ -344,3 +345,25 @@ def find_filters(partitions_time_attributes, start, end, filters, filter_column)
         filters,
         filter_column,
     )
+
+
+def get_combined_filters(datetime_filters=None, filters=None):
```
```suggestion
def combine_filters(datetime_filters, filters):
```

gtopper (changes requested):

tests/test_flow.py
tomerm-iguazio: This is the purpose of the test.
In pandas filters you can pass list of lists of tuples, but here I forbidden it and raise an error if it happens:)

storey/sources.py
tomerm-iguazio: renamed to self.additional_filters

storey/sources.py
tomerm-iguazio: tnx!

storey/utils.py
tomerm-iguazio: nope, if datetime_filters are not exist, and filter exists, we will need to return filters.
moreover, we want to add `additional_filters` to each list (=or_filter).

storey/sources.py
tomerm-iguazio: There is timezone issue, let's talk:)

storey/utils.py
tomerm-iguazio: I prefer to remain the None as default value. Name changed.

storey/sources.py
tomerm-iguazio: removed limitation.

tests/test_flow.py
gtopper: Screenshot 2024-04-24 at 19 24 36

storey/sources.py
gtopper: Actually renamed to `self._additional_filters`, which is good.

storey/utils.py
gtopper: But there is only ever one outer filter that we create for `datetime_filters`.

storey/sources.py
```diff
@@ -982,8 +982,13 @@ class ParquetSource(DataframeSource):
     :parameter end_filter: datetime. If not None, the results will be filtered by partitions
         'filter_column' <= end_filter. Default is None.
     :parameter filter_column: Optional. if not None, the results will be filtered by this column and before and/or after
+                datetime column.
     :param key_field: column to be used as key for events. can be list of columns
     :param id_field: column to be used as ID for events.
+    :param additional_filters: other filters to use while reading the parquet.
+                    Supported operators: '=', '>=', '<=', '>', '<'.
+                    Example: ('Product', '=', 'Computer')]
+    https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html
```
gtopper, line 992:
```suggestion
    For all supported filters, please see:
    https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html
```

storey/sources.py
```diff
@@ -1009,6 +1015,20 @@ def __init__(
...
+                f"ParquetSource supports additional_filters only as a list of tuples."
+                f" Current additional_filters: {additional_filters}"
+            )
+
+        for filters_tuple in additional_filters:
+            if filter_column and filter_column in filters_tuple:
+                raise ValueError(
+                    f"Cannot use the same column as both filter_column and in additional_filters."
+                    f" Column: {filter_column}."
+                )
```
gtopper: Why do we have to police this though? If the user really wants to, I don't see why we have to preempt them.

tests/test_flow.py
gtopper: Never mind, it's the error case.

storey/utils.py
tomerm-iguazio: look at `test_filter_before_after_partitioned_random`

tomerm-iguazio: and also there are situations were there are not datetime_filters.

gtopper: Why? It passes with the 3 line implementation I posted above.

tomerm-iguazio: fixed.

gtopper (approved):

PR #516, opened 2024-05-06
Add `TDEngineTarget`

gtopper:
[ML-6367](https://iguazio.atlassian.net/browse/ML-6367)

[ML-6367]: https://iguazio.atlassian.net/browse/ML-6367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -4,6 +4,7 @@ from datetime import datetime
 import pytest
 import taosrest
 from taosrest import ConnectError
+from taosws import QueryError
 
 from storey import SyncEmitSource, build_flow
 from storey.targets import TDEngineTarget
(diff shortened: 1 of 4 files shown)


Eyal-Danieli (approved):
cool cool cool

Eyal-Danieli:
Note that TDEngine best practice is using WebSocket connection and not REST API : 

_"For a REST connection, users do not need to install the client driver taosc, providing the advantage of cross-platform ease of use. However, functions such as data subscription and binary data types are not available. Additionally, compared to Native and Websocket connections, a REST connection has the worst performance. For a Websocket connection, users also do not need to install the client driver taosc... Normally we recommend using Websocket connection."_

Any reason why not using WebSocket connection here?

assaf758:
The websocket connection API seems virtually identical (apart from installing taos-ws-py and importing taosws).
I would go with it - as they state it's more performant and also it's added features (data subscription and binary data types) might be useful down the road.

integration/test_tdengine.py
```diff
@@ -0,0 +1,116 @@
...
+from taosrest import ConnectError
+from taosws import QueryError
+
+from storey import SyncEmitSource, build_flow
+from storey.targets import TDEngineTarget
+
+url = os.getenv("TDENGINE_URL")
+user = os.getenv("TDENGINE_USER")
+password = os.getenv("TDENGINE_PASSWORD")
+has_tdengine_credentials = all([url, user, password]) or url.startswith("taosws")
```
Eyal-Danieli: it should fix the test failure as well 
```suggestion
has_tdengine_credentials = all([url, user, password]) or (url and url.startswith("taosws"))
```

Eyal-Danieli (commented):
a single comment that should fix the failure as well

Eyal-Danieli (approved):

PR #524, opened 2024-05-21
[TDEngine] Remove REST support

Eyal-Danieli:
Install `taospy[ws]` that includes python connector to TDEngine using either REST API or WebSocket also includes installation of unnecessary dependencies such as  `pytest-cov`.
Therefore, at the moment we are going to support only the WebSocket connection that can be implemented using `taos-ws-py` which doesn't include the unnecessary dependencies.

### DIFF
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -870,7 +870,6 @@ class TDEngineTarget(_Batching, _Writer):
         self._url = url
         self._database = database
 
-
     def _init(self):
         import taosws
 
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -781,6 +781,8 @@ class TDEngineTarget(_Batching, _Writer):
     :param time_col: Name of the time column.
     :param columns: List of column names to be passed to the DataFrame constructor. Use = notation for renaming fields
         (e.g. write_this=event_field). Use $ notation to refer to metadata ($key, event_time=$time).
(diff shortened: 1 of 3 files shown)


storey/targets.py
```diff
@@ -885,33 +877,17 @@ def __init__(
...
         _Batching._init(self)
         _Writer._init(self)
-
-        parsed_url = urlparse(self._url)
-
-        if parsed_url.scheme == "taosws":
-            import taosws
-
-            self._using_websocket = True
+        if self._url.startswith("taosws://"):
```
assaf758, line 887: this will be true whether u/p was provided or not, right? if that's the case need to improve the logic.

storey/targets.py
```diff
@@ -819,13 +816,9 @@ def __init__(
         table_col: Optional[str] = None,
         supertable: Optional[str] = None,
         tag_cols: Union[str, List[str], None] = None,
-        timeout: Optional[int] = None,
         time_format: Optional[str] = None,
         **kwargs,
     ):
```
assaf758, line 822: why did you remove the check that it's taosws?

Eyal-Danieli: it's only relevant to REST connection, there is no timeout param with the websocket connection

assaf758 (approved):

PR #534, opened 2024-09-04
Fix storey distribution - don't include `tests`/`integration` in packages

jond01:
Fixes [ML-7758](https://iguazio.atlassian.net/browse/ML-7758).

In addition:
* General Docker cleaning
* Add a proper `.gitignore` file
* Fix Python support labels and requirement in the distribution

[ML-7758]: https://iguazio.atlassian.net/browse/ML-7758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.dockerignore
```diff
@@ -1,3 +0,0 @@
-venv
-*.coverage
-coverage_reports
```
gtopper, line 3: First line seems like a legacy from project inception. Any reason you added lines 2 and 3 @tomerm-iguazio?

.gitignore.template
```diff
@@ -1,5 +0,0 @@
-.gitignore
```
gtopper, line 1: The reason to not check in a gitignore file is to avoid forcing the same gitignore on everyone.

.python-version
```diff
@@ -0,0 +1 @@
+3.9.19
```
gtopper: storey needs to support multiple python versions, not just this specific one. And I wouldn't want to have to keep bumping this for every patch release.

dev-requirements.txt
```diff
@@ -6,6 +6,8 @@ isort~=5.7
 pytest~=5.4.3
 coverage~=7.5
 pytest-benchmark~=3.2.3
+# Note: you might need a Lua installation to install lupa on Mac.
```
gtopper, line 9: might, or will?

setup.py
```diff
@@ -58,7 +59,8 @@ def load_deps(file_name):
     author_email="yaronh@iguazio.com",
     license="Apache",
     url="https://github.com/mlrun/storey",
-    packages=find_packages(),
+    packages=find_packages(include=["storey*"]),
+    python_requires=">=3.9",
```
gtopper, line 63: I guess we might as well because mlrun dropped older versions and 3.8 EOL is very near anyway.

gtopper (changes requested):
Thanks for the cleanup 🙂. As mentioned below, gitignore is not check in on purpose. Other than that and `.python-version`, it looks good.

setup.py
jond01: I didn't understand your comment here - we might as well what?

.python-version
jond01: You don't need to bump this on every release.
This file is for development purposes, for tools such as `pyenv`, `uv`, and GitHub Actions.

dev-requirements.txt
jond01: At the moment Lua is needed, since the relevant version has no macOS&Arm distribution:
https://pypi.org/project/lupa/1.14.1/#files
When `lupa` is upgraded to `>=2.2` - there will be no need, as there's a dedicated dist. for the mentioned platform:
https://pypi.org/project/lupa/2.2/#files

But I think the upgrade should be done separately.

.dockerignore
jond01: Refer to:
https://github.com/github/gitignore/blob/main/Python.gitignore
`.coverage` is for Coverage.py artifacts.
`coverage_reports` - not sure about this one.
`venv` is also very common.

.gitignore.template
jond01: It's actually helpful to have a shared `.gitignore` for all the contributors, and quite surprising that it's missing 🙂
If a private file is needed, one should use `.git/info/exclude`.
If something is missing from the shared file - please let me know.

dev-requirements.txt
jond01: By the way, do you happen to know what it is even used for? I didn't see any usage of it in the project

gtopper: Pretty sure it's used by the redis emulator. @alxtkr77 would be able to tell you more.

.dockerignore
gtopper: Yeah, I know what they are, I just want to know if @tomerm-iguazio added them for a specific reason.

tomerm-iguazio: I added them for generated coverage reports

.gitignore
```diff
@@ -0,0 +1,9 @@
+__pycache__/
```
tomerm-iguazio, line 1: please add test.db

tomerm-iguazio (changes requested):

.dockerignore
tomerm-iguazio: If we are not building docker image anymore, it is irrelevant.
I agree that It is required for gitignore file.

.python-version
jond01: Removed in 1ad4021

.gitignore
jond01: Done in 6d6f827

.gitignore
```diff
@@ -0,0 +1,12 @@
...
+*.py[cod]
```
tomerm-iguazio, line 2: Please add `bench-results.json`.
You can see it's usage in the makefile under `bench` section.

gtopper (approved):

.gitignore
```diff
@@ -0,0 +1,12 @@
...
+*.egg-info/
+dist/
+.idea
+venv/
+*.coverage
+coverage_reports/
+.python-version
+test.db
+
+.DS_Store
```
tomerm-iguazio, line 28: Please add `/docs/_build`
This path is generated when running `make docs`

tomerm-iguazio (changes requested):
Looks good, we need to add few paths to gitignore according to makefile :)

tomerm-iguazio (approved):
looks good:)

PR #536, opened 2024-09-10
Use parameter binding in TDEngine target insertions

jond01:
See:
https://docs.tdengine.com/develop/insert-data/sql-writing/#insert-with-parameter-binding
https://docs.tdengine.com/reference/connectors/python/#writing-data-via-parameter-binding

Fixes [ML-7662](https://iguazio.atlassian.net/browse/ML-7662).

Improve performance and security, e.g. protection against SQL injections, with parameter-bound `INSERT` transaction.

[ML-7662]: https://iguazio.atlassian.net/browse/ML-7662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -1,17 +1,16 @@
 import os
-from datetime import datetime
+from datetime import datetime, timezone
 
 import pytest
-import pytz
 import taosws
 
 from storey import SyncEmitSource, build_flow
 from storey.targets import TDEngineTarget
 
-url = os.getenv("TDENGINE_URL")
+url = os.getenv("TDENGINE_URL")  # e.g.: taosws://root:taosdata@localhost:6041
 user = os.getenv("TDENGINE_USER")
 password = os.getenv("TDENGINE_PASSWORD")
-has_tdengine_credentials = all([url, user, password]) or (url and url.startswith("taosws"))
+has_tdengine_credentials = all([url, user, password]) or (url and url.startswith("taosws://"))
 
 
 @pytest.fixture()
@@ -19,15 +18,10 @@ def tdengine():
     db_name = "storey"
     supertable_name = "test_supertable"
 
-    if url.startswith("taosws"):
+    if url.startswith("taosws://"):
         connection = taosws.connect(url)
     else:
-
-        connection = taosws.connect(
-            url=url,
-            user=user,
-            password=password,
-        )
+        connection = taosws.connect(url=url, user=user, password=password)
 
     try:
         connection.execute(f"DROP DATABASE {db_name};")
@@ -116,7 +110,7 @@ def test_tdengine_target(tdengine, table_col):
             if typ == "TIMESTAMP":
                 t = datetime.fromisoformat(row[field_index])
                 # websocket returns a timestamp with the local time zone
-                t = t.astimezone(pytz.UTC).replace(tzinfo=None)
+                t = t.astimezone(timezone.utc).replace(tzinfo=None)
                 row[field_index] = t
         result_list.append(row)
     if table_col:
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -20,7 +20,7 @@ TDEngineData = tuple[taosws.Connection, str, Optional[str], Optional[str], str, 
 
 
 @pytest.fixture(params=[10])
-def tdengine(request: pytest.FixtureRequest) -> Iterator[TDEngineData]:
+def tdengine(request: "pytest.FixtureRequest") -> Iterator[TDEngineData]:
     db_name = "storey"
     supertable_name = "test_supertable"
 
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -1,5 +1,7 @@
 import os
+from collections.abc import Iterator
 from datetime import datetime, timezone
+from typing import Optional
 
 import pytest
 import taosws
@@ -14,9 +16,11 @@ has_tdengine_credentials = all([url, user, password]) or (url and url.startswith
 
 pytestmark = pytest.mark.skipif(not has_tdengine_credentials, reason="Missing TDEngine URL, user, and/or password")
 
+TDEngineData = tuple[taosws.Connection, str, Optional[str], Optional[str], str, str]
+
 
 @pytest.fixture()
-def tdengine():
+def tdengine() -> Iterator[TDEngineData]:
     db_name = "storey"
     supertable_name = "test_supertable"
 
@@ -51,7 +55,7 @@ def tdengine():
 
 
 @pytest.mark.parametrize("table_col", [None, "$key", "table"])
-def test_tdengine_target(tdengine, table_col):
+def test_tdengine_target(tdengine: TDEngineData, table_col: Optional[str]) -> None:
     connection, url, user, password, db_name, supertable_name = tdengine
     time_format = "%d/%m/%y %H:%M:%S UTC%z"
 
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -158,7 +158,6 @@ def test_sql_injection(tdengine: TDEngineData) -> None:
                 user=user,
                 password=password,
                 database=db_name,
-                table=None,
                 table_col=table_col,
                 supertable=supertable_name,
                 tag_cols=["my_int"],
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -12,6 +12,8 @@ user = os.getenv("TDENGINE_USER")
 password = os.getenv("TDENGINE_PASSWORD")
 has_tdengine_credentials = all([url, user, password]) or (url and url.startswith("taosws://"))
 
+pytestmark = pytest.mark.skipif(not has_tdengine_credentials, reason="Missing TDEngine URL, user, and/or password")
+
 
 @pytest.fixture()
 def tdengine():
@@ -49,7 +51,6 @@ def tdengine():
 
 
 @pytest.mark.parametrize("table_col", [None, "$key", "table"])
-@pytest.mark.skipif(not has_tdengine_credentials, reason="Missing TDEngine URL, user, and/or password")
 def test_tdengine_target(tdengine, table_col):
     connection, url, user, password, db_name, supertable_name = tdengine
     time_format = "%d/%m/%y %H:%M:%S UTC%z"
--- a/integration/test_tdengine.py
+++ b/integration/test_tdengine.py
@@ -19,8 +19,8 @@ pytestmark = pytest.mark.skipif(not has_tdengine_credentials, reason="Missing TD
 TDEngineData = tuple[taosws.Connection, str, Optional[str], Optional[str], str, str]
 
 
-@pytest.fixture()
-def tdengine() -> Iterator[TDEngineData]:
+@pytest.fixture(params=[10])
+def tdengine(request: pytest.FixtureRequest) -> Iterator[TDEngineData]:
     db_name = "storey"
     supertable_name = "test_supertable"
 
@@ -44,7 +44,9 @@ def tdengine() -> Iterator[TDEngineData]:
         if "STable not exist" not in str(err):
             raise err
 
-    connection.execute(f"CREATE STABLE {supertable_name} (time TIMESTAMP, my_string NCHAR(10)) TAGS (my_int INT);")
+    connection.execute(
+        f"CREATE STABLE {supertable_name} (time TIMESTAMP, my_string NCHAR({request.param})) TAGS (my_int INT);"
+    )
 
     # Test runs
     yield connection, url, user, password, db_name, supertable_name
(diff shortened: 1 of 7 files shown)


storey/targets.py
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+            for val_ind, val_name in enumerate(val_names):
+                raw_value = event.get(val_name)
+                if event_ind == 0:
+                    to_column_funs.append(self._raw_value_to_column_fun(raw_value))
+                values[val_ind].append(self._raw_value_to_value(raw_value))
+        return [fun(vals) for fun, vals in zip(to_column_funs, values)]
+
+    async def _emit(self, batch: list[dict], batch_key: str, batch_time, batch_events, last_event_time=None):
+        stmt = self._connection.statement()
+        stmt.prepare(self._sql_template)
```
gtopper, line 1022: Please move prepared statement to self for better performance.

integration/test_tdengine.py
```diff
@@ -133,3 +134,49 @@ def test_tdengine_target(tdengine, table_col):
...
+        [
+            SyncEmitSource(),
+            TDEngineTarget(
+                url=url,
+                time_col="time",
+                columns=["my_string"],
+                user=user,
+                password=password,
+                database=db_name,
+                table=None,
```
gtopper:
```suggestion
```

storey/targets.py
```diff
@@ -805,6 +805,9 @@ class TDEngineTarget(_Batching, _Writer):
     :type flush_after_seconds: int
     """
 
+    class UnsupportedTDEngineTypeError(TypeError):
+        pass
```
gtopper: I don't think we need a custom error for that. If we do, a better place for it is probably in `dtypes.py` with the other custom errors. There's no need for it to be nested (which is something I don't think we do anywhere in storey or even in mlrun).

storey/targets.py
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+        raise cls._get_unsupported_error(value)
+
+    def _get_tags_from_event(self, event: dict) -> list:
+        tags = []
+        for tag_name in self._columns[: self._number_of_tags]:
+            tags.append(self._value_to_tag(event.get(tag_name)))
+        return tags
+
+    @classmethod
+    def _raw_value_to_column_fun(cls, value) -> Callable:
```
```suggestion
    def _raw_value_to_column_func(cls, value) -> Callable:
```
(or function)

Applies elsewhere as well.

storey/targets.py
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+            return int(value.timestamp() * 1000)
+        return value
+
+    def _get_batch_values(self, batch: list[dict]) -> list:
+        values = [[] for _ in range(self._number_of_values)]
+        to_column_funs = []
+        val_names = self._columns[self._number_of_tags :]
+        for event_ind, event in enumerate(batch):
+            for val_ind, val_name in enumerate(val_names):
+                raw_value = event.get(val_name)
```
gtopper: Need to raise an error if missing (or is that ok in some cases – i.e. nullable columns?). As it's written, it will raise a
```UnsupportedTDEngineTypeError("Unsupported value type NoneType in TDEngineTarget")```

File: storey/targets.py [hunk 6]
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+    @classmethod
+    def _raw_value_to_value(cls, value):
+        if isinstance(value, datetime.datetime):
+            return int(value.timestamp() * 1000)
+        return value
+
+    def _get_batch_values(self, batch: list[dict]) -> list:
+        values = [[] for _ in range(self._number_of_values)]
+        to_column_funs = []
+        val_names = self._columns[self._number_of_tags :]
```
gtopper: I don't think "value names" is correct nomenclature.
```suggestion
        col_names = self._columns[self._number_of_tags :]
```
or `column_names`. Or `regular_column_names` I suppose.

storey/targets.py
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+        for event_ind, event in enumerate(batch):
+            for val_ind, val_name in enumerate(val_names):
+                raw_value = event.get(val_name)
+                if event_ind == 0:
+                    to_column_funs.append(self._raw_value_to_column_fun(raw_value))
+                values[val_ind].append(self._raw_value_to_value(raw_value))
+        return [fun(vals) for fun, vals in zip(to_column_funs, values)]
```
gtopper: Seems unnecessarily functional to me. I'd do something like
```python
    def _get_batch_values(self, batch: list[dict]) -> list:
        result = []
        val_names = self._columns[self._number_of_tags :]
        for event in batch:
            values = []
            for val_name in val_names:
                raw_value = event.get(val_name)
                values.append(self._raw_value_to_value(raw_value))
            to_column_func = self._raw_value_to_column_fun(event.get(val_names[0]))
            result.append(to_column_func(values))
        return result
```
(potentially dealing with `event.get(val_names[0])` differently, and do note that I didn't test this snippet)

storey/targets.py
```diff
@@ -888,48 +891,100 @@ def _init(self):
...
+        if isinstance(value, float):
+            return taosws.float_to_tag(value)
+        if isinstance(value, str):
+            return taosws.varchar_to_tag(value)
+        raise cls._get_unsupported_error(value)
+
+    def _get_tags_from_event(self, event: dict) -> list:
+        tags = []
+        for tag_name in self._columns[: self._number_of_tags]:
+            tags.append(self._value_to_tag(event.get(tag_name)))
```
gtopper: Here too, we need to handle a missing value better than raising an `Unsupported value type NoneType` error.

gtopper (commented):
Nice. Some comments but looks good overall.

For reference, this is the PR that added parameter binding in mlrun – https://github.com/mlrun/mlrun/pull/6195.

storey/targets.py
jond01: I tried reusing a statement, but it cannot "clean" previous bindings - it only aggregates the new parameters.
So I think we should instantiate a new statement for each emission.

storey/targets.py
jond01: I'll handle the type resolution with the table's schema - obtained through a `DESCRIBE` query.

storey/targets.py
gtopper: Ok, maybe there's a way to get it working, but we can put it aside for now.

storey/targets.py
jond01: Done

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+        self, table_name: str
+    ) -> tuple[
+        list[tuple[str, Callable[[Any], "taosws.PyTagView"]]], list[tuple[str, Callable[[list], "taosws.PyColumnView"]]]
+    ]:
+        fields_data = [_TDEngineFieldData(*raw) for raw in self._connection.query(f"DESCRIBE {table_name};")]
+        tags_schema = []
+        reg_cols_schema = []
+        for field_data in fields_data:
+            name = field_data.field
+            type_ = field_data.type
```
gtopper: Maybe call it `field_type`.

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+            try:
+                if field_data.note == "TAG":
+                    tags_schema.append((name, self._tdengine_type_to_tag_func[type_]))
+                else:
+                    reg_cols_schema.append((name, self._tdengine_type_to_column_func[type_]))
+            except KeyError:
```
gtopper: It's better check using a condition rather than causing a `KeyError` and catching it, from a performance standpoint.

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+                raise TDEngineTypeError(
+                    f"Unsupported value type '{type_}' of field '{name}' in {self.__class__.__name__}"
```
```suggestion
                    f"Unsupported value type '{type_}' of field '{name}' in {type(self).__name__}"
```

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
-        return str(value)
+    def _get_params_template(num_param: int) -> str:
+        return f"({','.join(num_param * ['?'])})"
+
+    def _get_table_schema(
+        self, table_name: str
+    ) -> tuple[
+        list[tuple[str, Callable[[Any], "taosws.PyTagView"]]], list[tuple[str, Callable[[list], "taosws.PyColumnView"]]]
+    ]:
+        fields_data = [_TDEngineFieldData(*raw) for raw in self._connection.query(f"DESCRIBE {table_name};")]
```
gtopper: I guess we should validate the table name in the constructor then?

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+        tags_schema = []
+        reg_cols_schema = []
+        for field_data in fields_data:
```
gtopper: Maybe just call it
```suggestion
        for field in fields:
```
because `fields_data` isn't correct grammar, which makes it harder to understand.

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+    @staticmethod
+    def _get_tags_from_event(
+        tags_schema: list[tuple[str, Callable[[Any], "taosws.PyTagView"]]], event: dict
+    ) -> list["taosws.PyTagView"]:
+        return [tag_func(event.get(tag_name)) for tag_name, tag_func in tags_schema]
+
+    @staticmethod
+    def _raw_value_to_value(value):
+        if isinstance(value, datetime.datetime):
+            return int(value.timestamp() * 1000)  # use millisecond precision
```
gtopper: I'd clarify to avoid implying that we have any other choice.
```suggestion
            return int(value.timestamp() * 1000)  # convert to milliseconds
```

tests/test_types.py
```diff
@@ -76,3 +77,21 @@ def test_emit_policy_period():
     policy = _dict_to_emit_policy(policy_dict)
     assert type(policy) == EmitAfterPeriod
     assert policy.delay_in_seconds == 8
+
+
+@pytest.mark.parametrize(
+    "description",
+    [
+        # taosBenchmark
+        # list(conn.query("describe test.d6241;"))
```
gtopper: What does it mean?

tests/test_types.py
```diff
@@ -76,3 +77,21 @@ def test_emit_policy_period():
...
+        ("ts", "TIMESTAMP", 8, "", "delta-i", "lz4", "medium"),
+        ("current", "FLOAT", 4, "", "delta-d", "lz4", "medium"),
+        ("voltage", "INT", 4, "", "simple8b", "lz4", "medium"),
+        ("phase", "FLOAT", 4, "", "delta-d", "lz4", "medium"),
+        ("groupid", "INT", 4, "TAG", "disabled", "disabled", "disabled"),
+        ("location", "VARCHAR", 24, "TAG", "disabled", "disabled", "disabled"),
+    ],
+)
+def test_description_parsing_to_tdengine_field(description: tuple) -> None:
+    """Test that the parsing works"""
```
gtopper: Not sure I understand what parsing this is testing. If it does test parsing, where are the error cases?

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+    def _get_batch_values(
+        cls, reg_cols_schema: list[tuple[str, Callable[[list], "taosws.PyColumnView"]]], batch: list[dict]
+    ) -> list["taosws.PyColumnView"]:
+        return [
+            col_func([cls._raw_value_to_value(event.get(col_name)) for event in batch])
+            for col_name, col_func in reg_cols_schema
+        ]
+
+    async def _emit(self, batch: list[dict], batch_key: str, batch_time, batch_events, last_event_time=None):
+        tags_schema, reg_cols_schema = self._get_table_schema(self._table or self._supertable)
```
gtopper: We shouldn't have to describe the table before each statement, doubling the number of IO operations.

storey/targets.py
```diff
@@ -888,48 +922,87 @@ def _init(self):
...
+        if isinstance(value, datetime.datetime):
+            return int(value.timestamp() * 1000)  # use millisecond precision
+        return value
+
+    @classmethod
+    def _get_batch_values(
+        cls, reg_cols_schema: list[tuple[str, Callable[[list], "taosws.PyColumnView"]]], batch: list[dict]
+    ) -> list["taosws.PyColumnView"]:
+        return [
+            col_func([cls._raw_value_to_value(event.get(col_name)) for event in batch])
```
gtopper, line 1016: It's a little hard to follow, but what happens if `event.get(col_name)` resolves to `None`?

storey/targets.py
jond01: But we do - we could use micro-sec or nano-sec:
https://docs.tdengine.com/reference/taos-sql/data-type/#timestamp

https://docs.tdengine.com/cloud/taos-sql/data-type/#data-types:
> Default precision is millisecond, microsecond and nanosecond are also supported.

tests/test_types.py
jond01: It comes from the benchmark DB - specifically this table. I think it can be useful as a reference

storey/targets.py
jond01: Then it is converted via `col_func`, and inserted to the table as `NULL`.

tests/test_types.py
jond01: Only happy flows are tested here.
But for example - more/fewer fields than expected will result in an error.

storey/targets.py
gtopper: I meant we don't have a choice at this point in the code :)

tests/test_types.py
gtopper: That can be said for any constructor or function that doesn't take `*args` though. It doesn't really count as parsing. And doesn't really justify a test if I'm not misunderstanding something here.

storey/targets.py
jond01: This part does not create the table so it does not decide on precision.
However, the table's creator has the choice, but we only support ms at the moment.

tests/test_types.py
jond01: This test is beneficial for the following reasons:
1. Increases the coverage in the non-integration tests that run in the CI. TDEngine integration tests are skipped.
2. Insures a regression in the class - e.g. removing/adding a field - will be caught.

tests/test_types.py
jond01: https://docs.tdengine.com/get-started/docker/#test-data-insert-performance

tests/test_types.py
gtopper: Would be nice to turn it into a test of actual functionality by testing the logic in `_get_table_schema` (preferably by moving everything other than `self._connection.query(...)` to another method and testing that).

Then it can test error cases as well.

gtopper (approved):

PR #537, opened 2024-09-15
Rewrite Choice step to make it usable from mlrun

gtopper:
[ML-7818](https://iguazio.atlassian.net/browse/ML-7818)

[ML-7818]: https://iguazio.atlassian.net/browse/ML-7818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -252,15 +252,16 @@ class Flow:
     def _should_terminate(self):
         return self._termination_received == len(self._inlets)
 
-    async def _do_downstream(self, event):
-        if not self._outlets:
+    async def _do_downstream(self, event, outlets=None):
+        outlets = self._outlets if outlets is None else outlets
+        if not outlets:
             return
         if event is _termination_obj:
             # Only propagate the termination object once we received one per inlet
-            self._outlets[0]._termination_received += 1
-            if self._outlets[0]._should_terminate():
-                self._termination_result = await self._outlets[0]._do(_termination_obj)
-            for outlet in self._outlets[1:] + self._get_recovery_steps():
+            outlets[0]._termination_received += 1
+            if outlets[0]._should_terminate():
+                self._termination_result = await outlets[0]._do(_termination_obj)
+            for outlet in outlets[1:] + self._get_recovery_steps():
                 outlet._termination_received += 1
                 if outlet._should_terminate():
                     self._termination_result = self._termination_result_fn(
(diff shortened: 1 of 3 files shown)


storey/flow.py
```diff
@@ -347,46 +348,48 @@ def _get_uuid(self):
...
-
-        self._choice_array = choice_array
-        for outlet, _ in choice_array:
-            self.to(outlet)
-
-        if default:
-            self.to(default)
-        self._default = default
+    def _init(self, **kwargs):
+        super()._init()
```
guy1992l, line 357: Just to make sure it is not an error: before the change the `kwargs` were sent at the `Flow` init so I would expect to see `super().__init__(**kwargs)` unless it is redundant. I think the `outlets` param should be passed here?

storey/flow.py
```diff
@@ -347,46 +348,48 @@ def _get_uuid(self):
...
+        self._name_to_outlet = {}
+        for outlet in self._outlets:
+            if outlet.name in self._name_to_outlet:
+                raise ValueError(f"Ambiguous outlet name '{outlet.name}' in Choice step")
+            self._name_to_outlet[outlet.name] = outlet
+        # TODO: hacky way of supporting mlrun preview, which replaces targets with a DFTarget
+        self._passthrough_for_preview = list(self._name_to_outlet) == ["dataframe"]
+
+    def select_outlets(self, event):
```
guy1992l: I would add a type hint and also in the doc string mention that the selected outlets should be returned in a list of strings - the outlet names.

```suggestion
    def select_outlets(self, event) -> List[str]:
```

storey/flow.py
```diff
@@ -347,46 +348,48 @@ def _get_uuid(self):
...
-            await self._default._do(event)
+        if event is _termination_obj:
+            return await self._do_downstream(_termination_obj)
+        else:
+            event_body = event if self._full_event else event.body
+            outlet_names = self.select_outlets(event_body)
+            outlets = []
+            if self._passthrough_for_preview:
+                outlet = self._name_to_outlet["dataframe"]
+                outlets.append(outlet)
```
guy1992l, line 382: add a TODO here as well to remember to delete this "hack"

guy1992l (commented):
I'm not the person who should review it, but if I may, it looks pretty much straight forward. I had a comment and few suggestions but besides these, A+ :)

storey/flow.py
gtopper: I don't really want to duplicate the TODO that's already on the attribute definition.

storey/flow.py
gtopper: Thanks, the `**kwargs` need to be removed. There's no kwargs in `_init()`. Note that `_init` is unrelated to `__init__`.

guy1992l (approved):

PR #543, opened 2024-11-05
`ParallelExecution` step

gtopper:
[ML-7689](https://iguazio.atlassian.net/browse/ML-7689)

[ML-7689]: https://iguazio.atlassian.net/browse/ML-7689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1435,6 +1435,9 @@ class Context:
 class ParallelExecutionRunnable:
     execution_mechanism = "multiprocessing"
 
+    def __init__(self, name):
+        self.name = name
+
     def init(self):
         pass
 
@@ -1451,7 +1454,11 @@ class ParallelExecution(Flow):
         super()._init()
         num_process = 0
         num_thread = 0
+        names = set()
         for runnable in self._runnables:
+            if runnable.name in names:
+                raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
+            names.add(runnable.name)
             runnable.init()
             if runnable.execution_mechanism == "multiprocessing":
                 num_process += 1
@@ -1477,5 +1484,8 @@ class ParallelExecution(Flow):
                     executor = self._executors[runnable.execution_mechanism]
                     task = asyncio.get_running_loop().run_in_executor(executor, runnable.run, event)
                 tasks.append(task)
-            event.body = await asyncio.gather(*tasks)
+            results = await asyncio.gather(*tasks)
+            event.body = {"inputs": event.body, "outputs": {}}
+            for index, result in enumerate(results):
+                event.body["outputs"][self._runnables[index].name] = result
             return await self._do_downstream(event)
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1476,18 +1476,18 @@ class ParallelExecutionRunnable:
     def init(self):
         pass
 
-    def run(self, event):
+    def run(self, event, path: str):
         return event
 
-    def _run(self, event):
+    def _run(self, event, path: str):
         start = time.monotonic()
-        data = self.run(event)
+        data = self.run(event, path)
         end = time.monotonic()
         return _ParallelExecutionRunnableResult(data, end - start)
 
-    async def _async_run(self, event):
+    async def _async_run(self, event, path: str):
         start = time.monotonic()
-        data = await self.run(event)
+        data = await self.run(event, path)
         end = time.monotonic()
         return _ParallelExecutionRunnableResult(data, end - start)
 
@@ -1563,15 +1563,21 @@ class ParallelExecution(Flow):
                     runnable = self._runnable_by_name[runnable]
                 if id(runnable) in runnables_encountered:
                     raise ValueError(f"select_runnables() returned more than one outlet named '{runnable.name}'")
+                input = event.body if runnable.execution_mechanism == "multiprocessing" else copy.deepcopy(event.body)
                 runnables_encountered.add(id(runnable))
                 if runnable.execution_mechanism == "asyncio":
-                    future = asyncio.get_running_loop().create_task(runnable._async_run(event))
+                    future = asyncio.get_running_loop().create_task(runnable._async_run(input, event.path))
                 elif runnable.execution_mechanism == "naive":
                     future = asyncio.get_running_loop().create_future()
-                    future.set_result(runnable._run(event))
+                    future.set_result(runnable._run(input, event.path))
                 else:
                     executor = self._executors[runnable.execution_mechanism]
-                    future = asyncio.get_running_loop().run_in_executor(executor, runnable._run, event)
+                    future = asyncio.get_running_loop().run_in_executor(
+                        executor,
+                        runnable._run,
+                        input,
+                        event.path,
+                    )
                 futures.append(future)
             results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
             event.body = {"input": event.body, "results": {}}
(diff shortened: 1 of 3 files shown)


ZeevRispler (approved):
LGTM

storey/flow.py
```diff
@@ -1435,3 +1436,182 @@ def get_table(self, key):
...
+        if not runnables:
+            raise ValueError("ParallelExecution cannot be instantiated without at least one runnable")
+
+        self.runnables = runnables
+        self._runnable_by_name = {}
+
+        self.max_processes = max_processes
+        self.max_threads = max_threads
+
+    def select_runnables(self, event):
```
davesh0812: I think typing for the return value can be very helpful here

storey/flow.py
```diff
@@ -1435,3 +1436,182 @@ def get_table(self, key):
...
+    :param runnables: A list of ParallelExecutionRunnable instances.
+    :param max_processes: Maximum number of processes to spawn.
+    :param max_threads: Maximum number of threads to start.
+    """
+
+    def __init__(
+        self,
+        runnables: list[ParallelExecutionRunnable],
+        max_processes: Optional[int] = None,
+        max_threads: Optional[int] = None,
```
davesh0812, line 1535: maybe we can add some default

storey/flow.py
```diff
@@ -1435,3 +1436,182 @@ def get_table(self, key):
...
+    Subclasses must also override the run() method with user code that handles the event and returns a result.
+
+    Subclasses may optionally override the init() method if the user's implementation of run() requires prior
+    initialization.
+
+    :param name: Runnable name
+    """
+
+    execution_mechanism = None
+    supported_mechanisms = ("multiprocessing", "threading", "asyncio", "naive")
```
davesh0812: I think it would be better to move this definition out of the class.

storey/flow.py
```diff
@@ -1435,3 +1436,182 @@ def get_table(self, key):
...
+                        executor,
+                        runnable._run,
+                        input,
+                        event.path,
+                    )
+                futures.append(future)
+            results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
+            event.body = {"input": event.body, "results": {}}
+            for result in results:
+                event.body["results"][result.runnable_name] = {"runtime": result.runtime, "output": result.data}
```
davesh0812, line 1618: I think we should avoid modifying the original event directly and instead add all relevant information under a new key (self.name represent the name of the current step(ParallelExecution)):
```suggestion
            event.body["result"][self.name] = {"input": input}
            for result in results:
                event.body["results"][self.name][result.runnable_name] = {"runtime": result.runtime, "output": result.data}
```

storey/flow.py
```diff
@@ -1435,3 +1436,182 @@ def get_table(self, key):
...
+            if runnables is None:
+                runnables = self.runnables
+            futures = []
+            runnables_encountered = set()
+            for runnable in runnables:
+                if isinstance(runnable, str):
+                    runnable = self._runnable_by_name[runnable]
+                if id(runnable) in runnables_encountered:
+                    raise ValueError(f"select_runnables() returned more than one outlet named '{runnable.name}'")
+                input = event.body if runnable.execution_mechanism == "multiprocessing" else copy.deepcopy(event.body)
```
davesh0812, line 1599: What about the input mechanism, the user can choose some path in the  `event.body`?

davesh0812 (commented):
Looks great... few comments :)

storey/flow.py
gtopper: Already done :)

storey/flow.py
gtopper: Not sure how helpful it would be... It could also be confusing if we want to keep supporting `None` as a way to set no limit.

gtopper: We discussed and decided to add some reasonable limit, but maybe in mlrun and not here (TBD).

storey/flow.py
gtopper: The current output is in line with the planning we've done so far, but we've discussed the fact that we should run this by DS and consider different common scenarios, and maybe need to change this format accordingly.

Therefore, leaving as is in this PR as discussed.

storey/flow.py
gtopper: As discussed, this will be looked at together with the output format.

storey/flow.py
gtopper: Actually, there is already a default in `ProcessPoolExecutor` which is [os.cpu_count() or 1](https://github.com/python/cpython/blob/v3.9.21/Lib/concurrent/futures/process.py#L592), and in `ThreadPoolExecutor` which is [min(32, (os.cpu_count() or 1) + 4)](https://github.com/python/cpython/blob/v3.9.21/Lib/concurrent/futures/thread.py#L142).

Although python 3.13 changes these defaults as per https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor and https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor.

gtopper: I'm adding explicit defaults in anyway.

storey/flow.py
```diff
@@ -1435,3 +1437,185 @@ def get_table(self, key):
...
+            elif runnable.execution_mechanism == "threading":
+                num_threads += 1
+            elif runnable.execution_mechanism not in ("asyncio", "naive"):
+                raise ValueError(f"Unsupported execution mechanism: {runnable.execution_mechanism}")
+
+        # enforce max
+        if self.max_processes:
+            num_processes = min(num_processes, self.max_processes)
+        if self.max_threads:
+            num_threads = min(num_threads, self.max_threads)
```
davesh0812: The if statement can be remove - due to the new defaults.  
```suggestion
        num_processes = min(num_processes, self.max_processes)
        num_threads = min(num_threads, self.max_threads)
```

davesh0812 (approved):
LGTM

storey/flow.py
```diff
@@ -1435,3 +1437,183 @@ def get_table(self, key):
...
+    * "multiprocessing" – To run in a separate process. This is appropriate for CPU or GPU intensive tasks as they
+        would otherwise block the main process by holding Python's Global Interpreter Lock (GIL).
+    * "threading" – To run in a separate thread. This is appropriate for blocking I/O tasks, as they would otherwise
+        block the main event loop thread.
+    * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
+        loop to continue running while waiting for a response.
+    * "naive" – To run in the main event loop. This is appropriate only for trivial computation and/or file I/O. It
+        means that the runnable will not actually be run in parallel to anything else.
+
+    Subclasses must also override the run() method with user code that handles the event and returns a result.
```
assaf758: also need to mention the run vs run_async thing

storey/flow.py
```diff
@@ -1435,3 +1437,183 @@ def get_table(self, key):
...
+
+    def _init(self):
+        super()._init()
+        num_processes = 0
+        num_threads = 0
+        for runnable in self.runnables:
+            if runnable.name in self._runnable_by_name:
+                raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
+            self._runnable_by_name[runnable.name] = runnable
+            print(f"initializing runnable {runnable.name}")
```
assaf758: is this print needed?

assaf758 (commented):
Looks great!
minor comments

storey/flow.py
gtopper: Not really.

assaf758 (approved):

PR #549, opened 2024-12-16
Add API to terminate and await termination

gtopper:
To make it easier for the caller to do both regardless of whether source is sync or async.

[ML-8738](https://iguazio.atlassian.net/browse/ML-8738)

[ML-8738]: https://iguazio.atlassian.net/browse/ML-8738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_flow.py
+++ b/tests/test_flow.py
@@ -185,8 +185,7 @@ def test_offset_commit():
             event.shard_id = shard
             event.offset = offset
             controller.emit(event)
-    controller.terminate()
-    termination_result = controller.await_termination()
+    termination_result = controller.terminate(wait=True)
     assert termination_result == 330
 
     offsets = copy.copy(platform.offsets)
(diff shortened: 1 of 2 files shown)


tests/test_flow.py
```diff
@@ -225,9 +224,8 @@ async def async_offset_commit():
     try:
         assert offsets == {("/", i): num_records_per_shard for i in range(num_shards)}
     finally:
-        await controller.terminate()
+        termination_result = await controller.terminate(wait=True)
 
-    termination_result = await controller.await_termination()
```
assaf758, line 230: why this line is removed?

gtopper: Replaced by `termination_result = await controller.terminate(wait=True)` which terminates and awaits termination in one.

assaf758 (approved):

PR #552, opened 2024-12-23
Add dedicated process mode to `ParallelExecutionRunnable`

gtopper:
To avoid serialization of the runnable per event, and to allow the user to allocate resources specific to that runnable (e.g. GPUs).

### DIFF
--- a/tests/test_flow.py
+++ b/tests/test_flow.py
@@ -4782,7 +4782,7 @@ def test_filters_type():
 
 
 class RunnableBusyWait(ParallelExecutionRunnable):
-    execution_mechanism = "multiprocessing"
+    execution_mechanism = "process_pool"
     _result = 0
 
     def init(self):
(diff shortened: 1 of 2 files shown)


gtopper:
After talking to @GiladShapira94, the plan is to keep the original process pool mode and add an addition sticky process mode (one-to-one for now, perhaps one-to-many in the future).

tests/test_flow.py
```diff
@@ -4915,31 +4919,48 @@ def test_invalid_runnable():
...
+class RunnableMultiprocessingWithLargeData(ParallelExecutionRunnable):
+    execution_mechanism = "dedicated_process"
+
+    def __init__(self, data_size, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.data = None
+        self.data_size = data_size
+
+    def init(self):
+        self.data = list(range(self.data_size))
```
assaf758, line 4937: can you demo how would the runnable allocate a gpu? For example, by mapping from runnable-number to gpu 1-3

assaf758 (commented):
Looks good. One comment inline.

assaf758 (approved):

PR #558, opened 2025-04-08
Run CI on python 3.9 & 3.11

davesh0812:
https://iguazio.atlassian.net/browse/ML-9774


.github/workflows/pr.yml
```diff
@@ -14,48 +14,67 @@
 #
 name: CI
 
-on: pull_request_target
+on:
+  pull_request:
+    branches:
+      - main
```
gtopper: Doesn't even exist.
```suggestion
```

.github/workflows/push-dev.yml
```diff
@@ -4,17 +4,21 @@ on:
   push:
     branches:
       - development
+      - 1.8.x
```
gtopper: I think we can just delete this file. No one uses its outputs.

gtopper (approved):

PR #563, opened 2025-05-12
[ParallelExecution] Adding metadata field to event

royischoss:
https://iguazio.atlassian.net/browse/ML-9892

Includes:
1. Allow `ParallelExecutionRunnable` raise_error behavior allowing catching errors in run
2. Adding field to `_ParallelExecutionRunnableResult` for timestamp of the run.
3. Adding field `monitoring_data` to the event output from `ParallelExecution:_do`

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1665,14 +1665,12 @@ class ParallelExecution(Flow):
                     )
                 futures.append(future)
             results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(
-                *futures, return_exceptions=self._return_exceptions
+                *futures
             )
             if len(self.runnables) == 1:
                 event.body = results[0].data if results else None
                 if self.monitored:
-                    setattr(
-                        event,
-                        "monitoring_data",
+                    event.monitoring_data = (
                         {
                             "microsec": results[0].runtime,
                             "when": results[0].timestamp.isoformat(sep=" ", timespec="microseconds"),
@@ -1694,5 +1692,5 @@ class ParallelExecution(Flow):
                         }
                         for result in results
                     }
-                    setattr(event, "monitoring_data", monitoring_data)
+                    event.monitoring_data = monitoring_data
             return await self._do_downstream(event)
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1664,9 +1664,7 @@ class ParallelExecution(Flow):
                         event.path,
                     )
                 futures.append(future)
-            results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(
-                *futures
-            )
+            results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
             if len(self.runnables) == 1:
                 event.body = results[0].data if results else None
                 if self.monitored:
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1518,10 +1518,10 @@ class ParallelExecutionRunnable:
         try:
             body = self.run(body, path)
         except Exception as e:
-            if not self._raise_exception:
-                body = {"error": f"{type(e)}: {e}"}
-            else:
+            if self._raise_exception:
                 raise e
+            else:
+                body = {"error": f"{type(e)}: {e}"}
         end = time.monotonic()
         return _ParallelExecutionRunnableResult(self.name, body, end - start, timestamp)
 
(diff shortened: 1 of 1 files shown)


storey/flow.py
```diff
@@ -1649,6 +1667,28 @@ async def _do(self, event):
...
+                            "when": result.timestamp.isoformat(sep=" ", timespec="microseconds"),
+                            "error": (
+                                event.body.get(result.runnable_name).get("error")
+                                if isinstance(event.body.get(result.runnable_name), dict)
+                                else None
+                            ),
+                        }
+                        for result in results
+                    }
+                    event.monitoring_data = monitoring_data
```
gtopper: I guess there's no reason to assign indirectly?
```suggestion
                if self.monitored:
                    event.monitoring_data = {
                        result.runnable_name: {
                            "microsec": result.runtime,
                            "when": result.timestamp.isoformat(sep=" ", timespec="microseconds"),
                            "error": (
                                event.body.get(result.runnable_name).get("error")
                                if isinstance(event.body.get(result.runnable_name), dict)
                                else None
                            ),
                        }
                        for result in results
                    }
```

gtopper: Also wondering if it should be `event._monitoring_data` since we don't want the user rummaging through it.

storey/flow.py
```diff
@@ -1485,6 +1486,7 @@ def __init__(self, name: str, **kwargs):
                 '"process_pool", "dedicated_process", "thread_pool", "asyncio", "naive"'
             )
         self.name = name
+        self._raise_exception = kwargs.get("raise_exception", True)
```
gtopper: I think you can add it to the parameter list explicitly.

storey/flow.py
```diff
@@ -1566,6 +1582,8 @@ def __init__(
 
         self._process_executor_by_runnable_name = {}
 
+        self.monitored = kwargs.get("track", False)
```
gtopper: I think you can add it to the parameter list explicitly. And maybe decide on one name vis-a-vis `monitored` vs `tracked`?

storey/flow.py
```diff
@@ -1511,16 +1513,30 @@ async def run_async(self, body: Any, path: str) -> Any:
         return body
 
     def _run(self, body: Any, path: str) -> Any:
+        timestamp = datetime.datetime.now(tz=datetime.timezone.utc)
```
gtopper, line 1516: I wonder if we can't just use start for this? If we need a timestamp object, we can always build it from the numerical representation.

storey/flow.py
```diff
@@ -1511,16 +1513,30 @@ async def run_async(self, body: Any, path: str) -> Any:
...
         start = time.monotonic()
-        body = self.run(body, path)
+        try:
+            body = self.run(body, path)
+        except Exception as e:
+            if not self._raise_exception:
+                body = {"error": str(e)}
```
gtopper: I think `str(e)` gives just the message, without the type. So something like `NotImplementedError` without a message would appear as `{"error": ""}`. That is to say, we should probably make it something like
```suggestion
                body = {"error": f"{type(e)}: {e}"}
```

storey/flow.py
royischoss: I think monotonic does not represent a real clock

gtopper: Ah, true.

storey/flow.py
```diff
@@ -1649,6 +1668,27 @@ async def _do(self, event):
             results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
             if len(self.runnables) == 1:
                 event.body = results[0].data if results else None
+                if self.monitored:
+                    event.monitoring_data = (
```
gtopper: I guess you missed this comment. Wdyt about it?
> Also wondering if it should be event._monitoring_data since we don't want the user rummaging through it.

storey/flow.py
```diff
@@ -1511,16 +1513,30 @@ async def run_async(self, body: Any, path: str) -> Any:
...
+        timestamp = datetime.datetime.now(tz=datetime.timezone.utc)
         start = time.monotonic()
-        body = await self.run_async(body, path)
+        try:
+            body = await self.run_async(body, path)
+        except Exception as e:
+            if not self._raise_exception:
+                body = {"error": f"{type(e)}: {e}"}
+            else:
+                raise e
```
gtopper: Very minor thing but I'd reverse the condition (elsewhere as well).
```suggestion
            if self._raise_exception:
                raise e
            else:
                body = {"error": f"{type(e)}: {e}"}
```

storey/flow.py
royischoss: sounds about right thanks

gtopper (approved):

davesh0812 (approved):

PR #568, opened 2025-05-26
[Flow] Implement shared executor for `ParallelExecution` steps

davesh0812:
Part A  of https://iguazio.atlassian.net/browse/ML-9967

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -26,11 +26,9 @@ import uuid
 from asyncio import Task
 from collections import defaultdict
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
-from lib2to3.fixes.fix_input import context
 from typing import Any, Callable, Collection, Dict, Iterable, List, Optional, Set, Union
 
 import aiohttp
-from blackd import executor
 
 from .dtypes import Event, FlowError, V3ioError, _termination_obj, known_driver_schemes
 from .queue import AsyncQueue
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1799,7 +1799,9 @@ class ParallelExecution(Flow):
             runnables_encountered = set()
             for runnable in runnables:
                 runnable: ParallelExecutionRunnable = (
-                    runnable if isinstance(runnable, ParallelExecutionRunnable) else self._runnable_by_name[runnable]
+                    runnable
+                    if isinstance(runnable, ParallelExecutionRunnable)
+                    else self.runnable_executor._runnable_by_name[runnable]
                 )
                 if runnable.execution_mechanism == ParallelExecutionMechanisms.shared_executor:
                     future = self.context.executor.run_executor(
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -22,13 +22,13 @@ import pickle
 import time
 import traceback
 import uuid
-from asyncio import Task
+import enum
+from asyncio import Task, Future
 from collections import defaultdict
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 from typing import Any, Callable, Collection, Dict, Iterable, List, Optional, Set, Union
 
 import aiohttp
-
 from .dtypes import Event, FlowError, V3ioError, _termination_obj, known_driver_schemes
 from .queue import AsyncQueue
 from .table import Table
@@ -1450,7 +1450,47 @@ class _ParallelExecutionRunnableResult:
         self.timestamp = timestamp
 
 
-parallel_execution_mechanisms = ("process_pool", "dedicated_process", "thread_pool", "asyncio", "naive")
+parallel_execution_mechanisms = ("process_pool", "dedicated_process", "thread_pool", "asyncio", "shared_proxy", "naive")
+
+
+class ParallelExecutionMechanisms(str, enum.Enum):
+    process_pool = "process_pool"
+    dedicated_process = "dedicated_process"
+    thread_pool = "thread_pool"
+    asyncio = "asyncio"
+    shared_proxy = "shared_proxy"
+    naive = "naive"
+
+    def __str__(self):
+        return self.value
+
+    def __repr__(self):
+        return self.value
+
+    @classmethod
+    def all(cls) -> list[str]:
+        return [
+            ParallelExecutionMechanisms.process_pool,
+            ParallelExecutionMechanisms.dedicated_process,
+            ParallelExecutionMechanisms.thread_pool,
+            ParallelExecutionMechanisms.asyncio,
+            ParallelExecutionMechanisms.shared_proxy,
+            ParallelExecutionMechanisms.naive,
+        ]
+
+    @classmethod
+    def single_thread(cls) -> list[str]:
+        return [
+            ParallelExecutionMechanisms.asyncio,
+            ParallelExecutionMechanisms.naive,
+        ]
+
+    @classmethod
+    def process(cls) -> list[str]:
+        return [
+            ParallelExecutionMechanisms.process_pool,
+            ParallelExecutionMechanisms.dedicated_process,
+        ]
 
 
 class ParallelExecutionRunnable:
@@ -1464,6 +1504,7 @@ class ParallelExecutionRunnable:
         block the main event loop thread.
     * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
         loop to continue running while waiting for a response.
+    * "shared_proxy" - To run the shared model of the graph.
     * "naive" – To run in the main event loop. This is appropriate only for trivial computation and/or file I/O. It
         means that the runnable will not actually be run in parallel to anything else.
 
@@ -1480,7 +1521,7 @@ class ParallelExecutionRunnable:
 
     # ignore unused keyword arguments such as context which may be passed in by mlrun
     def __init__(self, name: str, raise_exception: bool = True, **kwargs):
-        if self.execution_mechanism not in parallel_execution_mechanisms:
+        if self.execution_mechanism not in ParallelExecutionMechanisms.all():
             raise ValueError(
                 "ParallelExecutionRunnable's execution_mechanism attribute must be overridden with one of: "
                 '"process_pool", "dedicated_process", "thread_pool", "asyncio", "naive"'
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -1450,7 +1451,47 @@ def __init__(self, runnable_name: str, data: Any, runtime: float, timestamp: dat
         self.timestamp = timestamp
 
 
-parallel_execution_mechanisms = ("process_pool", "dedicated_process", "thread_pool", "asyncio", "naive")
+parallel_execution_mechanisms = ("process_pool", "dedicated_process", "thread_pool", "asyncio", "shared_proxy", "naive")
```
gtopper: Do we need this if we have `ParallelExecutionMechanisms.all()`?

storey/flow.py
```diff
@@ -1464,6 +1505,7 @@ class ParallelExecutionRunnable:
         block the main event loop thread.
     * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
         loop to continue running while waiting for a response.
+    * "shared_proxy" - To run the shared model of the graph.
```
gtopper: Should probably elaborate further and avoid mention of model here (since this code is generic and not specific to model serving).

storey/flow.py
```diff
@@ -1552,7 +1594,124 @@ def _static_run(*args, **kwargs):
     return _sval._run(*args, **kwargs)
 
 
-class ParallelExecution(Flow):
+class Executor:
```
gtopper: Name might be a bit too generic? Maybe `RunnableExecutor`?

storey/flow.py
```diff
@@ -1552,7 +1594,124 @@ def _static_run(*args, **kwargs):
...
+        self.num_threads = 0
+
+        self._process_executor_by_runnable_name = {}
+        self._mp_context = None
+
+    def add_runnable(self, runnable: ParallelExecutionRunnable) -> None:
+        if runnable.name not in self.runnables:
+            self.runnables[runnable.name] = runnable
+        else:
+            raise ValueError(f"Model name '{runnable.name}' is not unique")
```
```suggestion
            raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
```

storey/flow.py
```diff
@@ -1552,7 +1594,124 @@ def _static_run(*args, **kwargs):
...
+
+    def init_runnable(self, runnable: Union[str, ParallelExecutionRunnable]) -> None:
+        if isinstance(runnable, ParallelExecutionRunnable) and runnable.name in self._runnable_by_name:
+            raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
+        if isinstance(runnable, str) and runnable not in self.runnables:
+            raise ValueError(f"{runnable} is not exists")
```
```suggestion
            raise ValueError(f"{runnable} does not exist")
```

storey/flow.py
gtopper: Also, don't we have the exact same check in `init_runnable`? Do we really need to check it in both the constructor and `_init()`?

storey/flow.py
```diff
@@ -1552,7 +1594,124 @@ def _static_run(*args, **kwargs):
...
+    def add_runnable(self, runnable: ParallelExecutionRunnable) -> None:
+        if runnable.name not in self.runnables:
+            self.runnables[runnable.name] = runnable
+        else:
+            raise ValueError(f"Model name '{runnable.name}' is not unique")
+
+    def init_runnable(self, runnable: Union[str, ParallelExecutionRunnable]) -> None:
+        if isinstance(runnable, ParallelExecutionRunnable) and runnable.name in self._runnable_by_name:
+            raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
+        if isinstance(runnable, str) and runnable not in self.runnables:
```
gtopper: I think it makes sense to first have an `if isinstance(runnable, str):` and the three cases under it.

storey/flow.py
```diff
@@ -1552,7 +1594,124 @@ def _static_run(*args, **kwargs):
...
+
+        self.max_processes = max_processes or os.cpu_count() or 16
+        self.max_threads = max_threads or 32
+        self.num_processes = 0
+        self.num_threads = 0
+
+        self._process_executor_by_runnable_name = {}
+        self._mp_context = None
+
+    def add_runnable(self, runnable: ParallelExecutionRunnable) -> None:
```
gtopper, line 1628: Do we need to make it mutable? It can be mutable in mlrun and immutable at runtime (this is a runtime object).

tests/test_flow.py
```diff
@@ -4969,3 +4970,50 @@ def test_parallel_execution_with_large_data():
         else:
             for expected_gpu, runnable_result in enumerate(result.values()):
                 assert runnable_result == {"data_size": data_size, "n": n, "gpu": expected_gpu}
+
+
+class RunnableShared(ParallelExecutionRunnable):
+    execution_mechanism = "shared_proxy"
```
gtopper: Maybe
```suggestion
    execution_mechanism = "shared_executor"
```
?

storey/flow.py
```diff
@@ -1464,6 +1502,7 @@ class ParallelExecutionRunnable:
         block the main event loop thread.
     * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
         loop to continue running while waiting for a response.
+    * "shared_proxy" - To run the shared runnable of the flow graph.
```
gtopper: Should probably elaborate further and consider renaming to something like shared_executor.

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+        else:
+            raise ValueError("Unexpected runnable type")
+        if id(runnable) in runnables_encountered:
+            raise ValueError(f"select_runnables() returned more than one outlet named '{runnable.name}'")
+        input = (
+            event.body
+            if runnable.execution_mechanism in ParallelExecutionMechanisms.process()
+            else copy.deepcopy(event.body)
+        )
+        runnables_encountered.add(id(runnable))
```
gtopper: `id` returns an int, and `runnables_encountered` is declared as a set of strings.

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+        elif runnable.execution_mechanism == ParallelExecutionMechanisms.dedicated_process:
+            executor = self._process_executor_by_runnable_name[runnable.name]
+            future = asyncio.get_running_loop().run_in_executor(
+                executor,
+                _static_run,
+                input,
+                event.path,
+            )
+        elif runnable.execution_mechanism == ParallelExecutionMechanisms.shared_proxy and isinstance(
+            self, ParallelExecution
```
gtopper: Maybe we can avoid this awkwardness by using composition over inheritance.

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+        self._executors = {}
+        if num_processes:
+            self._executors[ParallelExecutionMechanisms.process_pool] = ProcessPoolExecutor(
+                max_workers=num_processes, mp_context=self._mp_context
+            )
+        if num_threads:
+            self._executors[ParallelExecutionMechanisms.thread_pool] = ThreadPoolExecutor(max_workers=num_threads)
+
+    def run_executor(
+        self, runnable_name: ParallelExecutionRunnable, runnables_encountered: set[str], event
```
```suggestion
        self, runnable_name: str, runnables_encountered: set[str], event
```

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+    ) -> asyncio.Future:
+        if isinstance(runnable_name, str):
+            runnable = self._runnable_by_name[runnable_name]
+        elif isinstance(runnable_name, ParallelExecutionRunnable):
+            runnable = runnable_name
+        else:
+            raise ValueError("Unexpected runnable type")
```
gtopper: Maybe we can tell the user what the type was. Also it should probably be a `TypeError` rather than a `ValueError`.

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+
+    def init_runnable(self, runnable: Union[str, ParallelExecutionRunnable]) -> None:
+        if isinstance(runnable, ParallelExecutionRunnable) and runnable.name in self._runnable_by_name:
+            raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
+        if isinstance(runnable, str) and runnable not in self.runnables:
+            raise ValueError(f"{runnable} is not exists")
+        elif isinstance(runnable, str) and runnable in self._runnable_by_name:
+            return
+        elif isinstance(runnable, str):
+            runnable = self.runnables[runnable]
```
gtopper: I'd refactor it a bit:
```suggestion
        if isinstance(runnable, str):
            if runnable not in self.runnables:
                raise ValueError(f"ParallelExecutionRunnable '{runnable}' does not exist")
            if runnable in self._runnable_by_name:
                return
            runnable = self.runnables[runnable]
        elif runnable.name in self._runnable_by_name:
            raise ValueError(f"ParallelExecutionRunnable name '{runnable.name}' is not unique")
```

storey/flow.py
gtopper: I also think it probably should have a docstring.

storey/flow.py
```diff
@@ -1552,7 +1591,124 @@ def _static_run(*args, **kwargs):
...
+            self, ParallelExecution
+        ):
+            self.context.executor.init_runnable(runnable=runnable.name)
+        elif runnable.execution_mechanism not in ParallelExecutionMechanisms.single_thread():
+            raise ValueError(f"Unsupported execution mechanism: {runnable.execution_mechanism}")
+
+    def init_executors(self):
+        # enforce max
+        num_threads = min(self.num_threads * 2, self.max_threads)
+        num_processes = min(self.num_processes * 2, self.max_processes)
```
gtopper: Why times 2?

storey/flow.py
gtopper: Or rather make it a union and change the name to `runnable`.

storey/flow.py
```diff
@@ -1565,8 +1721,6 @@ class ParallelExecution(Flow):
     def __init__(
         self,
         runnables: list[ParallelExecutionRunnable],
-        max_processes: Optional[int] = None,
-        max_threads: Optional[int] = None,
         **kwargs,
     ):
         super().__init__(**kwargs)
```
gtopper, line 1763: When introducing multiple inheritance, a single super init is no longer sufficient. But as mentioned elsewhere, I'm in favour of using composition instead.

gtopper (commented):
Looks right overall, but could use a little refactoring IMO. Particularly with respect to composition of the executor rather than (multiple) inheritance.

storey/flow.py
davesh0812: I added a `super().__init__(**kwargs)` in the `Executor` so it is OK.

storey/flow.py
davesh0812: I think that in MLRUN I will add the runnable it only at runtime

storey/flow.py
davesh0812: I treat it differently please have a look

storey/flow.py
davesh0812: changed this logic a bit

storey/flow.py
davesh0812: to be able run more when 1 event in the same time

storey/flow.py
```diff
@@ -1464,6 +1502,13 @@ class ParallelExecutionRunnable:
         block the main event loop thread.
     * "asyncio" – To run in an asyncio task. This is appropriate for I/O tasks that use asyncio, allowing the event
         loop to continue running while waiting for a response.
+    * "shared_executor" – Reuses an external executor (typically managed by the flow or context)
```
gtopper: I think it's important to explain that this is specifically for sharing the model between different steps in the graph. I.e. there is never a reason to use it if there is only one model runner in the graph.

storey/flow.py
gtopper: I'm pretty sure you don't want to add it at runtime (i.e. while already handling events). And `RunnableExecutor` should only exist at runtime, not when deploying the function SDK-side or server-side.

storey/flow.py
```diff
@@ -1634,33 +1798,20 @@ async def _do(self, event):
...
-                        input,
-                        event.path,
+                runnable: ParallelExecutionRunnable = (
+                    runnable
+                    if isinstance(runnable, ParallelExecutionRunnable)
+                    else self.runnable_executor._runnable_by_name[runnable]
+                )
+                if runnable.execution_mechanism == ParallelExecutionMechanisms.shared_executor:
+                    future = self.context.executor.run_executor(
+                        runnable=runnable.name, runnables_encountered=runnables_encountered, event=event
```
gtopper, line 1812: Only the executor differs:
```suggestion
                executor = (
                    self.context.executor
                    if runnable.execution_mechanism == ParallelExecutionMechanisms.shared_executor
                    else self.runnable_executor
                )
                future = executor.run_executor(
                    runnable=runnable, runnables_encountered=runnables_encountered, event=event
                )
```

storey/flow.py
```diff
@@ -1575,12 +1765,9 @@ def __init__(
...
         self.runnables = runnables
-        self._runnable_by_name = {}
-
-        self.max_processes = max_processes or os.cpu_count() or 16
-        self.max_threads = max_threads or 32
-
-        self._process_executor_by_runnable_name = {}
+        self.runnable_executor = RunnableExecutor(
+            max_processes=max_processes, max_threads=max_threads, pool_factor=pool_factor
+        )
```
gtopper: Actually, we need to move it to `_init`, because the graph is reusable. Otherwise, this will fail on graph termination and restart, which is what we do on rebalance.

storey/flow.py
```diff
@@ -1593,36 +1780,13 @@ def select_runnables(self, event) -> Optional[Union[list[str], list[ParallelExec
...
-
-        self._executors = {}
-        if num_processes:
-            self._executors["process_pool"] = ProcessPoolExecutor(max_workers=num_processes, mp_context=mp_context)
-        if num_threads:
-            self._executors["thread_pool"] = ThreadPoolExecutor(max_workers=num_threads)
+            if runnable.execution_mechanism == ParallelExecutionMechanisms.shared_executor:
+                self.context.executor.init_runnable(runnable=runnable.name)
+            else:
+                self.runnable_executor.add_runnable(runnable=runnable)
```
gtopper, line 1791: Won't this fail with `ParallelExecutionRunnable name '{runnable.name}' is not unique` if the same runnable is used from different runners (which is what we're trying to support)? Or am I missing something?

tests/test_flow.py
```diff
@@ -4969,3 +4969,49 @@ def test_parallel_execution_with_large_data():
...
+    source.to(parallel_execution).to(reduce)
+
+    start = time.monotonic()
+    controller = source.run()
+    controller.emit(0)
+    controller.terminate()
+    termination_result = controller.await_termination()
+    end = time.monotonic()
+
+    assert end - start < 6
```
gtopper: Pretty sure in this test we want to make sure it's less than 3. Because they could be running in serial and still be done in less than 6 seconds.

gtopper (commented):
Nice. Some comments.

gtopper (approved):

PR #569, opened 2025-05-28
[ParallelExecution] Update `event._metadata` dictionary instead of overwriting it

royischoss:
Relates to https://iguazio.atlassian.net/browse/ML-9892
As part of connecting to monitoring there is a need to allow metadata creation in mlrun side before calling the _do method

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1782,7 +1782,7 @@ class ParallelExecution(Flow):
     def preprocess_event(self, event):
         """
         Given an event, preprocess it with user code.
-        runs before the runnable selector.
+        Runs before the runnable selector.
         Should return the new enriched event.
         :param event: Event object
         """
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1591,6 +1591,14 @@ class ParallelExecution(Flow):
         """
         pass
 
+    def enrich_event(self, event):
+        """
+        Given an event, enriches it with user code.
+        Should return the new enriched event.
+        :param event: Event object
+        """
+        return event
+
     def _init(self):
         super()._init()
         num_processes = 0
@@ -1628,6 +1636,7 @@ class ParallelExecution(Flow):
         if event is _termination_obj:
             return await self._do_downstream(_termination_obj)
         else:
+            event = self.enrich_event(event)
             runnables = self.select_runnables(event)
             if runnables is None:
                 runnables = self.runnables
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1665,19 +1665,23 @@ class ParallelExecution(Flow):
             results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
             if len(self.runnables) == 1:
                 event.body = results[0].data if results else None
-                event._metadata = (
-                    {
-                        "microsec": results[0].runtime,
-                        "when": results[0].timestamp.isoformat(sep=" ", timespec="microseconds"),
-                    },
-                )
+
+                metadata = {
+                    "microsec": results[0].runtime,
+                    "when": results[0].timestamp.isoformat(sep=" ", timespec="microseconds"),
+                }
             else:
                 event.body = {result.runnable_name: result.data for result in results}
-                event._metadata = {
+                metadata = {
                     result.runnable_name: {
                         "microsec": result.runtime,
                         "when": result.timestamp.isoformat(sep=" ", timespec="microseconds"),
                     }
                     for result in results
                 }
+
+            if hasattr(event, "_metadata") and isinstance(event._metadata, dict):
+                event._metadata.update(metadata)
+            else:
+                event._metadata = metadata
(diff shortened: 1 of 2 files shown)


gtopper (commented):
Only comment is we should probably have a unit test that covers this.

tests/test_flow.py
```diff
@@ -4969,3 +4969,62 @@ def test_parallel_execution_with_large_data():
...
+
+    runnables = [
+        RunnableWithError("error"),
+        busy_wait_pool,
+        busy_wait_dedicated,
+        RunnableSleep("sleep1"),
+        RunnableSleep("sleep2"),
+        RunnableAsyncSleep("asleep1"),
+        RunnableAsyncSleep("asleep2"),
+        RunnableAsyncSleep("naive"),
```
gtopper: Probably we don't need quite so many for this test which is focused on the metadata attribute.

storey/flow.py
```diff
@@ -1665,19 +1674,23 @@ async def _do(self, event):
...
+                metadata = {
                     result.runnable_name: {
                         "microsec": result.runtime,
                         "when": result.timestamp.isoformat(sep=" ", timespec="microseconds"),
                     }
                     for result in results
                 }
+
+            if hasattr(event, "_metadata") and isinstance(event._metadata, dict):
+                event._metadata.update(metadata)
```
gtopper, line 1849: I don't think the test is validating this case though?

gtopper (commented):
Seems good other than a couple of comments about the test.

storey/flow.py
royischoss: We do cause in enrich_event I created this with name, might need to check it without creating it

tests/test_flow.py
```diff
@@ -5015,3 +5015,92 @@ def __init__(self, executor: RunnableExecutor):
...
+    ]
+
+    class MyParallelExecution(ParallelExecution):
+
+        def enrich_event(self, event):
+            event._metadata = {"name": self.name}
+            return event
+
+        def select_runnables(self, event):
+            return [runnable.name for runnable in runnables if runnable.name != "error"]
```
gtopper: Hm, this doesn't seem relevant here.

tests/test_flow.py
```diff
@@ -5015,3 +5015,92 @@ def __init__(self, executor: RunnableExecutor):
...
+
+    runnables = [
+        busy_wait_pool,
+        busy_wait_dedicated,
+    ]
+
+    class MyParallelExecution(ParallelExecution):
+
+        def select_runnables(self, event):
+            return [runnable.name for runnable in runnables if runnable.name != "error"]
```
gtopper: Hm, this doesn't seem relevant here.

storey/flow.py
```diff
@@ -1779,6 +1779,14 @@ def select_runnables(self, event) -> Optional[Union[list[str], list[ParallelExec
         """
         pass
 
+    def enrich_event(self, event):
```
gtopper: Maybe call it preprocess and explain in the docstring that this happens before the selector and the runnables. From the current name and docstring, one could just as well assume that this runs after the runnables or between the selector and the runnables.

tests/test_flow.py
```diff
@@ -5015,3 +5015,92 @@ def __init__(self, executor: RunnableExecutor):
...
+    class MyParallelExecution(ParallelExecution):
+
+        def enrich_event(self, event):
+            event._metadata = {"name": self.name}
+            return event
+
+        def select_runnables(self, event):
+            return [runnable.name for runnable in runnables if runnable.name != "error"]
+
+    parallel_execution = MyParallelExecution(runnables, full_event=True)
```
gtopper: I don't think `full_event` is observed by `ParallelExecution`. It always receives the full event and hands off just the body and path to the runnables.

tests/test_flow.py
```diff
@@ -5015,3 +5015,92 @@ def __init__(self, executor: RunnableExecutor):
...
+        busy_wait_pool,
+        busy_wait_dedicated,
+    ]
+
+    class MyParallelExecution(ParallelExecution):
+
+        def select_runnables(self, event):
+            return [runnable.name for runnable in runnables if runnable.name != "error"]
+
+    parallel_execution = MyParallelExecution(runnables, full_event=True)
```
gtopper: I don't think full_event is observed by ParallelExecution. It always receives the full event and hands off just the body and path to the runnables.

storey/flow.py
```diff
@@ -1779,6 +1779,15 @@ def select_runnables(self, event) -> Optional[Union[list[str], list[ParallelExec
         """
         pass
 
+    def preprocess_event(self, event):
+        """
+        Given an event, preprocess it with user code.
+        runs before the runnable selector.
```
```suggestion
        Runs before the runnable selector.
```

gtopper (approved):
LGTM otherwise 👍

gtopper (approved):

PR #581, opened 2025-07-14
Recover from error raised by the offset committer callback

gtopper:
[ML-10538](https://iguazio.atlassian.net/browse/ML-10538)

[ML-10538]: https://iguazio.atlassian.net/browse/ML-10538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/sources.py
+++ b/storey/sources.py
@@ -289,7 +289,9 @@ class SyncEmitSource(Flow):
                     or num_offsets_not_committed > 1
                     and time.monotonic() >= last_commit_time + self._max_time_before_commit
                 ):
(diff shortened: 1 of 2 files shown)


alxtkr77 (approved):
Looking good. One question - shouldn't we handle _do_downstream() similar way ? I am not that familiar with the logic, but it looks like it also has a potential to break the loop

gtopper:
@alxtkr77, we already wrap `_do_downstream()` calls in a try-except block, with some special handling because errors thrown there are presumed to be applicative errors.

PR #582, opened 2025-07-17
Fix error message format in `ParallelExecutionRunnable`

gtopper:
From something line
```
<class 'MyError'>: my error message
```
to
```
MyError: my error message
```

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1555,7 +1555,7 @@ class ParallelExecutionRunnable:
             if self._raise_exception:
                 raise e
             else:
-                body = {"error": f"{type(e)}: {e}"}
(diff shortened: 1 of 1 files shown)


royischoss (approved):
🔥  When we bump this in mlrun some tests are counting on this ugly format so let notice it.

gtopper:
@royischoss, which tests? Any system tests? So long as they're not system tests, CI can be our guide.

royischoss:
One system test the others are part of ci

gtopper:
Would be nice if you could tell me which one though @royischoss :)

royischoss:
`test_deploy_model_runner_error_handler` but it still not [merged](https://github.com/mlrun/mlrun/pull/8331) so I can react to it when you bump storey

PR #583, opened 2025-08-06
[ParallelExecution] Fix key error for invalid runnable

royischoss:
Fixes: https://iguazio.atlassian.net/browse/ML-10371
Include verification on runnable name and type returns from select_runnables

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1795,6 +1795,7 @@ class ParallelExecution(Flow):
             ParallelExecutionMechanisms.validate(execution_mechanism)
 
         self.runnables = runnables
+        self.register_runnables = set(r.name if isinstance(r, ParallelExecutionRunnable) else r for r in self.runnables)
         self.execution_mechanism_by_runnable_name = execution_mechanism_by_runnable_name
         self.max_processes = max_processes or os.cpu_count() or 16
         self.max_threads = max_threads or 32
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -1887,3 +1888,15 @@ async def _do(self, event):
...
             return await self._do_downstream(event)
+
+    def _verify_runnables(self, runnables: List[Union[str, ParallelExecutionRunnable]]):
+        """Verifies that the provided runnables are valid and registered."""
+        runnables_names = [r.name if isinstance(r, ParallelExecutionRunnable) else r for r in runnables]
+        register_runnables = [r.name if isinstance(r, ParallelExecutionRunnable) else r for r in self.runnables]
+        if not set(runnables_names).issubset(set(register_runnables)):
+            raise ValueError(
+                f"Runnables {set(runnables_names) - set(register_runnables)} are not part of the registered runnables"
+            )
```
gtopper: Suggested refactoring and error message improvement:
```suggestion
        runnable_names = {r.name if isinstance(r, ParallelExecutionRunnable) else r for r in runnables}
        registered_runnables = {r.name if isinstance(r, ParallelExecutionRunnable) else r for r in self.runnables}
        if not runnable_names.issubset(registered_runnables):
            unregistered_runnables_string = ",".join(runnable_names - registered_runnables)
            raise ValueError(f"The following selected Runnables are not registered: {unregistered_runnables_string}")
```

storey/flow.py
```diff
@@ -1887,3 +1888,15 @@ async def _do(self, event):
...
+        for runabble in runnables:
```
gtopper: Typo:
```suggestion
        for runnable in runnables:
```

storey/flow.py
```diff
@@ -1887,3 +1888,15 @@ async def _do(self, event):
...
+            if not isinstance(runabble, str) and not isinstance(runabble, ParallelExecutionRunnable):
```
```suggestion
            if not isinstance(runnable, (str, ParallelExecutionRunnable)):
```

storey/flow.py
```diff
@@ -1887,3 +1888,15 @@ async def _do(self, event):
...
+                raise TypeError(f"Expected a ParallelExecutionRunnable or str, but got: {type(runabble).__name__}")
```
gtopper: Either this check is redundant, or the type annotation is incorrect. For this reason, PyCharm identifies this code as unreachable.

storey/flow.py
gtopper: We should also probably put `registered_runnables` in self, to avoid recalculating it. Number of runnables can be large, and this is run per event.

storey/flow.py
royischoss: Weird I will test it

storey/flow.py
gtopper: @royischoss, my snippet above makes multiple changes, some of which you did not apply. Please update the code or let me know why you prefer not to make the suggested changes. 🙏

storey/flow.py
gtopper: Why is it weird though? Either the annotation is wrong, or the check is unnecesasry.

storey/flow.py
royischoss: I preferred caching in the init, and the unreachable code handling made me change the way I create the sets, because I am already iterating over the runnables

storey/flow.py
royischoss: It was unreachable because if one of the runnable type is other than string or ParallelExecutionRunnable the code should raise exception above, changed it checking the type before to allow a better error type

gtopper (approved):
LGTM ✅

PR #584, opened 2025-08-14
Timescaledb: handle schemas

alxtkr77:


storey/timescaledb_target.py
```diff
@@ -112,6 +112,9 @@ def __init__(
         self._dsn = dsn
         self._pool = None  # Connection pool will be created lazily during first use
         self._column_names = self._get_column_names()
+        self._schema = None
+        if "." in self._table:
+            self._schema, self._table = self._table.split(".", 1)
```
gtopper, line 117: Should probably edit the docstring accordingly. Is the term schema correct here? I couldn't find anything about a `<schema>.<table>` form in TimescaleDB.

alxtkr77: Schema is a feature in PostgreSQL (https://magicstack.github.io/asyncpg/current/_modules/asyncpg/connection.html#Connection.copy_records_to_table). Since the storey interface has no way to pass it, I concatenate it to the table name with '.'.

gtopper (approved):

PR #587, opened 2025-09-04
Discard events once persistence is given up

gtopper:
[ML-10957](https://iguazio.atlassian.net/browse/ML-10957)

While we may want to change our policy to not giving up in the first place, we should not prevent offset commit indefinitely for events that we never intend to process.

[ML-10957]: https://iguazio.atlassian.net/browse/ML-10957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


royischoss (approved):
LGTM

PR #588, opened 2025-09-09
Timescaledb: add column nullability check on insert and switch to psycopg3

alxtkr77:


integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+        # Missing required_col (non-nullable)
+    }
+
+    # Should raise ValueError for missing required column
+    with pytest.raises(ValueError, match=r"Missing required non-nullable column 'required_col'"):
+        controller.emit(invalid_data, None)
+        controller.terminate()
+        controller.await_termination()
+
+    # Cleanup
```
gtopper: The cleanup block will not run in case of a test failure (when an error is not raised).

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    # Test data missing the required column
+    invalid_data = {
+        "time": "18/09/19 01:55:10 UTC+0000",
+        "optional_col": "test_value",
+        # Missing required_col (non-nullable)
+    }
+
+    # Should raise ValueError for missing required column
+    with pytest.raises(ValueError, match=r"Missing required non-nullable column 'required_col'"):
+        controller.emit(invalid_data, None)
```
gtopper: No reason to pass `None`, especially by position.
```suggestion
        controller.emit(invalid_data)
```

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+        controller.terminate()
+        controller.await_termination()
+
+    # Cleanup
+    cursor = connection.cursor()
+    cursor.execute(f"DROP TABLE IF EXISTS {validation_table} CASCADE;")
+    cursor.close()
+
+
+def test_timescaledb_schema_validation_missing_nullable_column(timescaledb):
```
gtopper: Might be worth using parameterization, i.e. `@pytest.mark.parametrize("nullable", [True, False])`?

integration/test_timescaledb.py
gtopper: Applies elsewhere as well.

integration/test_timescaledb.py
gtopper: Applies elsewhere as well.

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+
+    # Verify data was inserted correctly for configured columns
+    cursor = connection.cursor()
+    cursor.execute(f"SELECT binary_col, int_col FROM {table_name} ORDER BY int_col DESC LIMIT 1")
+    result = cursor.fetchone()
+    assert result[1] == 42  # int_col
+    # PostgreSQL returns memoryview for binary data, convert to bytes for comparison
+    binary_result = bytes(result[0]) if isinstance(result[0], memoryview) else result[0]
+    assert binary_result == b"test_binary"  # binary_col
+    cursor.close()
```
gtopper: This won't run in case of error.

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    # Should raise TypeError for non-dictionary data (the error occurs in the parent Writer class)
+    with pytest.raises(TypeError, match=r"string indices must be integers"):
+        controller.emit(invalid_data, None)
+        controller.terminate()
+        controller.await_termination()
+
+
+@pytest.mark.asyncio
+async def test_timescaledb_validation_direct_emit_non_dict_error(timescaledb):
+    """Test that TimescaleDBTarget._emit properly validates dictionary data types"""
```
gtopper: It would be a little better to test it within a graph, which should be easy by setting `max_events` to 1.

storey/timescaledb_target.py
```diff
@@ -137,9 +140,55 @@ async def _async_init(self):
         while ensuring the pool is available when needed for data operations.
         """
         if self._pool is None:
-            import asyncpg
+            self._pool = AsyncConnectionPool(self._dsn, min_size=1, max_size=1, open=False)
+            await self._pool.open()
+
+    async def _get_table_schema(self) -> dict:
```
gtopper, line 216: This could actually end up initializing `self._table_schema` more than once under certain circumstances.

storey/timescaledb_target.py
```diff
@@ -179,26 +228,60 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
+                if col in schema:
+                    col_info = schema[col]
+                    if col in item:
+                        # Column present in data
+                        record.append(item[col])
+                    elif not col_info["nullable"]:
+                        # Column missing but required (not nullable)
+                        raise ValueError(
+                            f"Missing required non-nullable column '{col}' in event. "
+                            f"Available columns: {list(item.keys())}"
```
gtopper:
```suggestion
                            f"Available columns: {', '.join(item.keys())}"
```

storey/timescaledb_target.py
```diff
@@ -156,7 +205,7 @@ def _event_to_batch_entry(self, event):
         """
         return self._event_to_writer_entry(event)
 
-    async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_time=None):
+    async def _emit(self, batch, _batch_key, _batch_time, _batch_events, _last_event_time=None):
```
gtopper: Why is an underscore added to the parameters here? This is an overridden method, so its parameter names should not be changed.

storey/timescaledb_target.py
```diff
@@ -179,26 +228,60 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
+                table_identifier = sql.Identifier(self._table)
+                if self._schema:
+                    table_identifier = sql.Identifier(self._schema, self._table)
+
+                column_identifiers = [sql.Identifier(col) for col in self._column_names]
+                copy_query = sql.SQL("COPY {} ({}) FROM STDIN").format(
+                    table_identifier, sql.SQL(", ").join(column_identifiers)
+                )
+
+                async with cur.copy(copy_query) as copy:
```
gtopper: There is a built-in python module named `copy` (the one holding the `copy()` and `deepcopy()` functions), so the name, while not reserved, should probably be avoided to avoid confusion and future conflict.

storey/timescaledb_target.py
```diff
@@ -179,26 +228,60 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
-            # Use PostgreSQL's COPY protocol for optimal performance
-            # This is significantly faster than individual INSERT statements
-            await conn.copy_records_to_table(
-                self._table, schema_name=self._schema, records=records, columns=self._column_names
-            )
+        async with self._pool.connection() as conn:
+            async with conn.cursor() as cur:
+                # Use PostgreSQL's COPY protocol for optimal performance
+                # This is significantly faster than individual INSERT statements
+                import psycopg.sql as sql
```
gtopper: If we can import from `psycopg_pool` at the top, we can import from `psycopg.sql` there as well.

storey/timescaledb_target.py
```diff
@@ -137,9 +140,55 @@ async def _async_init(self):
...
+                WHERE table_name = %s
+                AND table_schema = %s
+                ORDER BY ordinal_position
+                """
+                schema_name = self._schema or "public"
+                await cur.execute(query, (self._table, schema_name))
+                rows = await cur.fetchall()
+
+            if not rows:
+                raise ValueError(f"Table '{schema_name}.{self._table}' not found or has no columns")
```
gtopper: Might be nice to tell the user which one it is.

storey/timescaledb_target.py
```diff
@@ -179,26 +228,60 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
+
         # Convert dictionaries to tuples for copy_records_to_table
         # PostgreSQL's COPY protocol requires data in tuple format with consistent column ordering
+        # Validate against schema to catch missing required columns early
 
         records = []
         for item in batch:
             if not isinstance(item, dict):
                 # Only dictionaries are supported as input
                 raise TypeError(f"TimescaleDBTarget only supports dictionary data, got {type(item)}")
```
gtopper:
```suggestion
                raise TypeError(f"TimescaleDBTarget only supports dictionary data, got {type(item).__name__}")
```

storey/timescaledb_target.py
gtopper: Also, any value in doing these in parallel? Together with increasing the pool size, of course.

storey/timescaledb_target.py
```diff
@@ -179,26 +228,60 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
         if not batch:
             return
 
+        # Get table schema for validation on first use
+        schema = await self._get_table_schema()
```
gtopper, line 316: I wonder if we shouldn't just do it in `_init()`, so that there's no need to cache it, just initialize it on startup.

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+                max_events=10,
+            ),
+        ]
+    ).run()
+
+    # Test with non-dictionary data
+    invalid_data = "this_is_not_a_dictionary"
+
+    # Should raise TypeError for non-dictionary data (the error occurs in the parent Writer class)
+    with pytest.raises(TypeError, match=r"string indices must be integers"):
```
gtopper: If we're already testing it, maybe we should raise a better error?

integration/test_timescaledb.py
```diff
@@ -306,3 +303,356 @@ async def test_timescaledb_async_emit(timescaledb):
...
+
+    # Create a schema and table
+    cursor = connection.cursor()
+    schema_name = "test_schema"
+    with contextlib.suppress(PsycopgError):
+        cursor.execute(f"CREATE SCHEMA IF NOT EXISTS {schema_name};")
+
+    schema_table = f"{schema_name}.{table_name}_schema"
+    with contextlib.suppress(PsycopgError):
+        cursor.execute(f"DROP TABLE IF EXISTS {schema_table};")
```
gtopper: Why would these raise an error though? If it's not expected, we probably shouldn't suppress it.

gtopper (commented):
Looks great, but I did have a few comments :)

integration/test_timescaledb.py
alxtkr77: Will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will change

storey/timescaledb_target.py
alxtkr77: The worst-case scenario is a few extra database queries during initialization, which is an acceptable trade-off for simpler, more maintainable code. The schema caching still provides the intended performance benefit for all subsequent calls.

storey/timescaledb_target.py
alxtkr77: ok

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: Will fix. Regarding the parallelization: 
Connection pool: Single connection is sufficient for most use cases since:
 1. COPY operations are already bulk-optimized and very fast
 2. Multiple concurrent COPY operations may cause lock contention
 3. Most data flows process batches sequentially, not concurrently

(added comment as well)

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: The issue is that the current architecture deliberately keeps _init() synchronous and defers async operations. If we wanted to do schema initialization in _init(), we'd need to use synchronous database operations.

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
```diff
@@ -98,69 +95,67 @@ def timescaledb(request: "pytest.FixtureRequest") -> Iterator[TimescaleDBData]:
...
-    test_connection.close()
+    finally:
+        # Cleanup - guaranteed to run even if test fails
+        try:
+            with test_connection.cursor() as cleanup_cursor:
+                cleanup_cursor.execute(f"DROP TABLE IF EXISTS {table_name} CASCADE;")
+            # Close connection after cursor operations are complete
+            test_connection.close()
+        except Exception as e:
+            print(f"Cleanup error: {e}")
```
gtopper: Just let it be raised normally.

integration/test_timescaledb.py
```diff
@@ -294,15 +286,375 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    else:
+        # Should raise ValueError for missing required (non-nullable) column
+        with pytest.raises(ValueError, match=r"Missing required non-nullable column 'test_col'"):
+            controller.emit(test_data)
+            controller.terminate()
+            controller.await_termination()
+
+    # Cleanup
+    with connection.cursor() as cursor:
+        cursor.execute(f"DROP TABLE IF EXISTS {validation_table} CASCADE;")
```
gtopper: The cleanup block will not run in case of a test failure.

integration/test_timescaledb.py
```diff
@@ -294,15 +286,375 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    # Should raise ValueError for missing required column, showing correct schema.table format
+    with pytest.raises(ValueError, match=r"Missing required non-nullable column 'required_col'"):
+        controller.emit(invalid_data)
+        controller.terminate()
+        controller.await_termination()
+
+    # Cleanup
+    with connection.cursor() as cursor:
+        cursor.execute(f"DROP TABLE IF EXISTS {schema_table} CASCADE;")
+        cursor.execute(f"DROP SCHEMA IF EXISTS {schema_name} CASCADE;")
```
gtopper: The cleanup block will not run in case of a test failure.

integration/test_timescaledb.py
```diff
@@ -294,15 +286,375 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    assert schema1 is schema2  # Should be the same object (cached)
+
+    # Verify schema contains expected information
+    assert "time" in schema1
+    assert "binary_col" in schema1
+    assert schema1["time"]["nullable"] is False  # time column is NOT NULL
+    assert schema1["binary_col"]["nullable"] is True  # binary_col allows NULL
+
+    # Cleanup
+    await target._terminate()
```
gtopper: The cleanup block will not run in case of a test failure.

integration/test_timescaledb.py
```diff
@@ -294,15 +286,375 @@ async def test_timescaledb_async_emit(timescaledb):
...
+        f"""
+        CREATE TABLE {validation_table} (
+            time TIMESTAMPTZ NOT NULL,
+            required_col INTEGER NOT NULL,
+            test_col VARCHAR(50) {test_col_nullable}
+        );
+    """
+    )
+    cursor.execute(f"SELECT create_hypertable('{validation_table}', 'time');")
+    cursor.close()
```
gtopper: This won't run in case of error.

storey/timescaledb_target.py
gtopper: I meant telling the user that the table doesn't exist, or that it has no columns. Pretty sure we can tell from whether rows is `None` or an empty list. It's nice to avoid ambiguity in the error message.

storey/timescaledb_target.py
gtopper: Right, the init sequence is synchronous. It would be good to do it there, but it's not very important.

integration/test_timescaledb.py
gtopper: I think you might have missed this one though since this line wasn't updated.

integration/test_timescaledb.py
```diff
@@ -294,15 +286,375 @@ async def test_timescaledb_async_emit(timescaledb):
...
+        dsn=dsn_url,
+        table=table_name,
+        time_col="time",
+        columns=columns_config,
+        time_format=time_format,
+        max_events=10,
+    )
+
+    # Initialize the target
+    await target._async_init()
```
gtopper: Here too, I think it would be better to test within a graph rather than by calling the target's private methods.

gtopper (commented):
Good stuff. A couple more things to consider.

integration/test_timescaledb.py
alxtkr77: Actually the error is raise by Writer::_event_to_writer_entry(), so not sure what to test there

integration/test_timescaledb.py
alxtkr77: ok

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: ok

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
gtopper: What I see in `Writer::_event_to_writer_entry()` is that it's supposed to raise `TypeError("Writer supports only events of type dict or list.")`, so why are we getting `string indices must be integers`?

integration/test_timescaledb.py
```diff
@@ -81,53 +61,56 @@ def _generate_test_data(precision: str, row_count: int = 5):
...
-    test_cursor.execute(f"SELECT create_hypertable('{table_name}', 'time');")
-    test_cursor.close()
+    try:
+        yield cleanup
+    finally:
+        if tables:
+            with psycopg.connect(dsn, autocommit=True) as connection:
+                with connection.cursor() as cursor:
+                    for table_name in tables:
+                        cursor.execute(f"DROP TABLE IF EXISTS {table_name} CASCADE;")
```
gtopper, line 86: I think you can drop them all together like this?
```suggestion
                    cursor.execute(f"DROP TABLE IF EXISTS {', '.join(tables)} CASCADE;")
```

integration/test_timescaledb.py
```diff
@@ -294,15 +276,563 @@ async def test_timescaledb_async_emit(timescaledb):
...
+    schema_name = "test_schema"
+    schema_table = f"{schema_name}.{table_name}_schema"
+
+    # Register table for cleanup (schema is persistent)
+    table_cleanup.add_table(schema_table)
+
+    with psycopg.connect(dsn_url, autocommit=True) as connection:
+        with connection.cursor() as cursor:
+            # Create TimescaleDB extension if needed
+            cursor.execute(f"CREATE SCHEMA IF NOT EXISTS {schema_name};")
```
gtopper, line 411: We don't need to drop the schema?

integration/test_timescaledb.py
```diff
@@ -138,36 +121,38 @@ def timescaledb(request: "pytest.FixtureRequest") -> Iterator[TimescaleDBData]:
...
-    # Cleanup
-    try:
-        cleanup_cursor = test_connection.cursor()
-        cleanup_cursor.execute(f"DROP TABLE IF EXISTS {table_name} CASCADE;")
-        cleanup_cursor.close()
-    except Exception as e:
-        print(f"Cleanup error: {e}")
-    test_connection.close()
 
+@pytest.fixture(params=[("milliseconds"), ("microseconds")])
```
gtopper: No-op parentheses. Either this should be:
```suggestion
@pytest.fixture(params=["milliseconds", "microseconds"])
```
or there is a bug here, and it should be
```suggestion
@pytest.fixture(params=[("milliseconds",), ("microseconds",)])
```

integration/test_timescaledb.py
```diff
@@ -138,36 +121,38 @@ def timescaledb(request: "pytest.FixtureRequest") -> Iterator[TimescaleDBData]:
...
+    table_name = f"test_table_extended_{timestamp_precision}"
+
+    yield _create_timescaledb_table_and_data(table_name, timestamp_precision, table_cleanup)
+
+
+@pytest.fixture(scope="function")
+def timescaledb(table_cleanup) -> TimescaleDBData:
+    """TimescaleDB fixture for tests that don't need timestamp precision variations."""
+    table_name = f"test_table_{uuid4().hex[:8]}"
+    timestamp_precision = "milliseconds"  # Default precision for single tests
```
gtopper: Not sure what "single tests" means in this context.

integration/test_timescaledb.py
```diff
@@ -191,23 +176,23 @@ def test_timescaledb_all_types_and_precision(timescaledb):
...
+        with connection.cursor() as cursor:
+            query = f"""
+                SELECT time, binary_col, bool_col, double_col, float_col,
+                       int_col, timestamp_col, nchar_col, varchar_col
+                FROM {table_name} ORDER BY int_col;
+            """
+            cursor.execute(query)
+
+            result_list = []
+            result_list.extend(list(row) for row in cursor.fetchall())
```
gtopper:
```suggestion
            result_list  = [list(row) for row in cursor.fetchall()]
```

integration/test_timescaledb.py
```diff
@@ -230,48 +215,45 @@ def test_timescaledb_all_types_and_precision(timescaledb):
...
+                    INSERT INTO {table_name} (time, binary_col, bool_col, double_col, float_col,
+                                            int_col, timestamp_col, nchar_col, varchar_col)
+                    VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
+                """,
+                    (test_time, b"test", True, 1.23, 4.56, 999, test_time, "MS", "millisecond_test"),
+                )
+
+                cursor.execute(f"SELECT EXTRACT(MILLISECONDS FROM time) FROM {table_name} WHERE int_col = 999")
+                result = cursor.fetchone()
+                assert result[0] >= 10000  # At least 10 seconds worth of milliseconds
```
gtopper: This helps readability a bit
```suggestion
                assert result[0] >= 10_000  # At least 10 seconds worth of milliseconds
```

integration/test_timescaledb.py
```diff
@@ -230,48 +215,45 @@ def test_timescaledb_all_types_and_precision(timescaledb):
...
+                    INSERT INTO {table_name} (time, binary_col, bool_col, double_col, float_col,
+                                            int_col, timestamp_col, nchar_col, varchar_col)
+                    VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s)
+                """,
+                    (test_time, b"test", True, 1.23, 4.56, 999, test_time, "US", "microsecond_test"),
+                )
+
+                cursor.execute(f"SELECT EXTRACT(MICROSECONDS FROM time) FROM {table_name} WHERE int_col = 999")
+                result = cursor.fetchone()
+                assert result[0] >= 10000000  # At least 10 seconds worth of microseconds
```
gtopper: This helps readability a bit
```suggestion
                assert result[0] >= 10_000_000  # At least 10 seconds worth of microseconds
```

integration/test_timescaledb.py
```diff
@@ -294,15 +276,563 @@ async def test_timescaledb_async_emit(timescaledb):
...
+
+    assert result == "success"
+    assert mock_operation.call_count == 4
+
+    # Verify sleep calls with expected timing: 0.1 + 0.025, 0.2 + 0.025, 0.4 + 0.025
+    expected_delays = [0.125, 0.225, 0.425]
+    actual_delays = [call[0][0] for call in mock_sleep.call_args_list]
+    # Use approximate comparison for floating point precision
+    for actual, expected in zip(actual_delays, expected_delays):
+        assert abs(actual - expected) < 1e-10
```
gtopper: Looks like something that could cause transient failures.

storey/timescaledb_target.py
```diff
@@ -137,9 +165,114 @@ async def _async_init(self):
...
+                if isinstance(e, psycopg.errors.DeadlockDetected):
+                    if deadlock_attempts >= self.MAX_DEADLOCK_RETRIES:
+                        raise ValueError(f"Deadlock persisted after {self.MAX_DEADLOCK_RETRIES} retries: {e}") from e
+                    # Fast retry for deadlocks: ~0.1s, ~0.2s, ~0.4s with jitter
+                    delay = (2**deadlock_attempts) * 0.1 + random.uniform(0, 0.05)
+                    deadlock_attempts += 1
+                else:
+                    if connection_attempts >= self._max_retries:
+                        raise ValueError(f"Connection failed after {self._max_retries} retries: {e}") from e
+                    # Slower retry for connection issues: 1s, 2s, 4s
```
gtopper: Comment is only correct if delay is not overridden by the caller. Even for the default value, it's better not to repeat it here, as this comment is likely to be overlooked if the default is changed.

storey/timescaledb_target.py
```diff
@@ -137,9 +165,114 @@ async def _async_init(self):
...
+        Note: Unhandled errors are passed through without wrapping to preserve
+        original exception types and stack traces for proper debugging.
+
+        :param operation_callable: Async function that executes the database operation
+        :return: Result of operation_callable()
+        """
+        deadlock_attempts = 0
+        connection_attempts = 0
+
+        for _ in range(self.MAX_DEADLOCK_RETRIES + self._max_retries + 1):
```
gtopper: Since it should never exit normally, it's probably clearer for it to just be
```suggestion
        while True:
```
Otherwise, it raises the question of whether this method can return `None`.

storey/timescaledb_target.py
```diff
@@ -179,26 +312,62 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
+                        # Column missing but required (not nullable)
+                        raise ValueError(
+                            f"Missing required non-nullable column '{col}' in event. "
+                            f"Available columns: {', '.join(item.keys())}"
+                        )
+                    else:
+                        # Column missing but nullable - use None
+                        record.append(None)
+                else:
+                    # Column not in schema - fallback to original behavior for compatibility
```
gtopper: Original behaviour?

gtopper (commented):
Looks good. Some comments.

integration/test_timescaledb.py
alxtkr77: The source of the "string indices must be integers" error is in storey/targets.py:204 in the _get_event_time() method:

  event_time = event.body[time_field]

  Where:
  - event.body = "this_is_not_a_dictionary" (string)
  - time_field = "time" (string)
  - Attempting: "this_is_not_a_dictionary"["time"] → TypeError: string indices must be integers

  Complete Call Stack Analysis

  Direct method call path:
  1. target._event_to_batch_entry(event)
  2. self._event_to_writer_entry(event)
  3. Validates isinstance(data, dict) → False
  4. Checks elif self._columns: → True
  5. Raises: TypeError: Writer supports only events of type dict or list.

  Flow processing path:
  1. Flow batching framework calls _get_event_time() first
  2. _get_event_time() tries event.body[time_field]
  3. "string_data"["time"] → TypeError: string indices must be integers
  4. Never reaches the proper validation!

integration/test_timescaledb.py
alxtkr77: TimescaleDB hypertables cannot be dropped together in a single statement due to a PostgreSQL/TimescaleDB limitation

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
alxtkr77: will fix

integration/test_timescaledb.py
gtopper: Cool... Can you fix it? 😄

Even if it's pre-existing code, we shouldn't be adding tests that expect the wrong thing to happen.

integration/test_timescaledb.py
gtopper: Oh well.

integration/test_timescaledb.py
```diff
@@ -294,15 +281,564 @@ async def test_timescaledb_async_emit(timescaledb):
...
+
+    assert result == "success"
+    assert mock_operation.call_count == 4
+
+    # Verify sleep calls with expected timing: 0.1 + 0.025, 0.2 + 0.025, 0.4 + 0.025
+    expected_delays = [0.125, 0.225, 0.425]
+    actual_delays = [call[0][0] for call in mock_sleep.call_args_list]
+    # Use approximate comparison for floating point precision
+    for actual, expected in zip(actual_delays, expected_delays):
+        assert abs(actual - expected) < 0.001
```
gtopper: I still think this risks transient failures, and we'd probably see this in CI if we had this test running there.

storey/timescaledb_target.py
```diff
@@ -179,26 +312,62 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
...
+                        raise ValueError(
+                            f"Missing required non-nullable column '{col}' in event. "
+                            f"Available columns: {', '.join(item.keys())}"
+                        )
+                    else:
+                        # Column missing but nullable - use None
+                        record.append(None)
+                else:
+                    # Column not in schema - use item.get()
+                    record.append(item.get(col))
```
gtopper: This is the same as `record.append(None)`. If this is really the behaviour we want (is it though?), then this code can be simplified quite a bit (always take `item.get(col)`, and raise an error if that's `None`, and the column is non-nullable in the schema).

integration/test_timescaledb.py
alxtkr77: will convert to integers

integration/test_timescaledb.py
alxtkr77: will fix

storey/timescaledb_target.py
alxtkr77: ok - maybe later

storey/timescaledb_target.py
alxtkr77: will fix

gtopper (approved):

PR #589, opened 2025-09-21
[Parquet Target] Add single parquet option when using partitioning

davesh0812:
In this case, we save the most recent parquet file in each partitioned directory.
Which parquet file is considered “last” depends on parameters such as `max_event` or `flush_after_seconds`.

### DIFF
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -561,7 +561,7 @@ class ParquetTarget(_Batching, _Writer):
             else:
                 partition_cols = [("$key", 256), "$year", "$month", "$day", "$hour"]
         else:
-            self._single_file_mode = single_file
+            self._single_file_mode = single_file or False
             kwargs["partition_cols"] = partition_cols
 
         if self._single_file_mode and not partition_cols:
--- a/storey/targets.py
+++ b/storey/targets.py
@@ -533,7 +533,7 @@ class ParquetTarget(_Batching, _Writer):
     :param storage_options: Extra options that make sense for a particular storage connection, e.g. host, port,
         username, password, etc., if using a URL that will be parsed by fsspec, e.g., starting
         "s3://”, "gcs://”. Optional.
-    :param single_file: If True, all the partitioned data will be written to a single file named {name}.parquet in
+    :param single_file: If True, all the partitioned data will be written to a single file named target.parquet in
         the specified path. If False (the default), each batch will be written to a separate file with a random uuid.
     :type storage_options: dict
     """
@@ -549,7 +549,7 @@ class ParquetTarget(_Batching, _Writer):
         infer_columns_from_data: Optional[bool] = None,
         max_events: Optional[int] = None,
         flush_after_seconds: Union[int, float, None] = None,
-        single_file: bool = False,
+        single_file: Optional[bool] = None,
         **kwargs,
     ):
         self._single_file_mode = False
@@ -633,12 +633,10 @@ class ParquetTarget(_Batching, _Writer):
             dir_path += "/"
         if dir_path and not self._file_system.exists(dir_path):
             self._file_system.makedirs(dir_path, exist_ok=True)
-        if not self._single_file_mode:
-            file_path = f"{dir_path}{uuid.uuid4()}.parquet"
-        elif self._single_file_mode and not self._partition_cols:
-            file_path = self._path
+        if self._single_file_mode:
+            file_path = f"{dir_path}target.parquet" if self._partition_cols else self._path
         else:
-            file_path = f"{dir_path}{self.name}.parquet"
+            file_path = f"{dir_path}{uuid.uuid4()}.parquet"
         # Remove nanosecs from timestamp columns & index
         for name, _ in df.items():
             if str(df[name].dtype) == "datetime64[ns]":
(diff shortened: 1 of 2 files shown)


storey/targets.py
```diff
@@ -547,6 +549,7 @@ def __init__(
         infer_columns_from_data: Optional[bool] = None,
         max_events: Optional[int] = None,
         flush_after_seconds: Union[int, float, None] = None,
+        single_file: bool = False,
```
```suggestion
        single_file: Optional[bool] = None,
```
This needs to be done for mlrun serialization, which is the reason the rest of the parameters use this pattern.

storey/targets.py
```diff
@@ -533,6 +533,8 @@ class ParquetTarget(_Batching, _Writer):
     :param storage_options: Extra options that make sense for a particular storage connection, e.g. host, port,
         username, password, etc., if using a URL that will be parsed by fsspec, e.g., starting
         "s3://”, "gcs://”. Optional.
+    :param single_file: If True, all the partitioned data will be written to a single file named {name}.parquet in
```
gtopper: So `{name}` is the name of the target? i.e. `ParquetTarget.parquet` or `my_parquet_target.parquet`? Not sure if that's the best way to go, but anyway this needs to be clarified if we keep this behaviour.

storey/targets.py
```diff
@@ -629,7 +633,12 @@ async def _emit(self, batch, batch_key, batch_time, batch_events, last_event_tim
             dir_path += "/"
         if dir_path and not self._file_system.exists(dir_path):
             self._file_system.makedirs(dir_path, exist_ok=True)
-        file_path = self._path if self._single_file_mode else f"{dir_path}{uuid.uuid4()}.parquet"
+        if not self._single_file_mode:
+            file_path = f"{dir_path}{uuid.uuid4()}.parquet"
+        elif self._single_file_mode and not self._partition_cols:
+            file_path = self._path
+        else:
+            file_path = f"{dir_path}{self.name}.parquet"
```
gtopper: Logic could use a bit of refactoring here:
```suggestion
        if self._single_file_mode:
            file_path = f"{dir_path}{self.name}.parquet" if self._partition_cols else self._path
        else:
            file_path = f"{dir_path}{uuid.uuid4()}.parquet"
```

tests/test_flow.py
```diff
@@ -2830,6 +2830,35 @@ def test_write_to_parquet_partition_by_datetime(tmpdir):
     assert read_back_df.equals(expected_df)
 
 
+@pytest.mark.parametrize("max_events", [1, 5])
+def test_write_to_single_partition_parquet(tmpdir, max_events):
```
gtopper, line 2834: I guess this test should check the file name?

It would be good for it to also run the flow twice to make sure the second run overwrites the first run. (this should work without rebuilding the flow – just terminate and rerun)

gtopper (commented):
Nice. Some comments. Maybe give a little background in the description and link back to the relevant ticket?

storey/targets.py
davesh0812: I will go with `target.parquet `

tests/test_flow.py
davesh0812: done

storey/targets.py
```diff
@@ -558,9 +561,10 @@ def __init__(
             else:
                 partition_cols = [("$key", 256), "$year", "$month", "$day", "$hour"]
         else:
+            self._single_file_mode = single_file
```
```suggestion
            self._single_file_mode = single_file or False
```

tests/test_flow.py
```diff
@@ -2830,6 +2830,62 @@ def test_write_to_parquet_partition_by_datetime(tmpdir):
...
+    def check_target_parquets(dir, ids=(0, 1, 2)):
```
```suggestion
    def check_target_parquets(dir_path, ids=(0, 1, 2)):
```
to avoid shadowing built-in `dir()`.

tests/test_flow.py
```diff
@@ -2830,6 +2830,62 @@ def test_write_to_parquet_partition_by_datetime(tmpdir):
...
+    expected_df["id"] = expected_df["id"].astype("int32").astype("category")
+    controller.terminate()
+    controller.await_termination()
+
+    read_back_df = pd.read_parquet(out_dir, columns=columns)
+    read_back_df.sort_values("my_int", inplace=True)
+    read_back_df.reset_index(drop=True, inplace=True)
+    assert read_back_df.equals(expected_df)
+
+    check_target_parquets(out_dir, ids=(0, 1, 2))
```
gtopper: Either don't pass them in, or remove the parameter default value?

tests/test_flow.py
```diff
@@ -2830,6 +2830,62 @@ def test_write_to_parquet_partition_by_datetime(tmpdir):
...
+            expected.append([i, f"this is {i}", i % 3])
+    expected_df = pd.DataFrame(expected, columns=columns)
+    expected_df["id"] = expected_df["id"].astype("int32").astype("category")
+    controller.terminate()
+    controller.await_termination()
+
+    read_back_df = pd.read_parquet(out_dir, columns=columns)
+    read_back_df.sort_values("my_int", inplace=True)
+    read_back_df.reset_index(drop=True, inplace=True)
+    assert read_back_df.equals(expected_df)
```
```suggestion
    pd.testing.assert_frame_equal(read_back_df, expected_df)
```

tests/test_flow.py
```diff
@@ -2830,6 +2830,62 @@ def test_write_to_parquet_partition_by_datetime(tmpdir):
...
```
```suggestion
    pd.testing.assert_frame_equal(read_back_df, expected_df)
```

gtopper (commented):
Great. A couple more minor comments.

gtopper (approved):

PR #593, opened 2025-12-08
Allow cycle graph & add option to select outlets from all kind of steps

davesh0812:
https://iguazio.atlassian.net/browse/ML-11495
https://iguazio.atlassian.net/browse/ML-11499

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -505,7 +505,7 @@ class _UnaryFunctionFlow(Flow):
         if fn_select_outlets and not callable(fn_select_outlets):
             raise TypeError(f"Expected fn_select_outlets to be callable, got {type(fn)}")
         self._outlets_selector = fn_select_outlets
-        self._create_name_to_outlet = self._outlets_selector is not None or self._method_is_overridden(
+        self._create_name_to_outlet = self._outlets_selector or self._method_is_overridden(
             "select_outlets", _UnaryFunctionFlow
         )
 
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -293,10 +293,7 @@ class Flow:
                     outlet_names = self._selected_outlets
                     self._selected_outlets = None
                 else:
-                    if asyncio.iscoroutinefunction(self.select_outlets):
-                        outlet_names = await self.select_outlets(event.body)
-                    else:
-                        outlet_names = self.select_outlets(event.body)
+                    outlet_names = self.select_outlets(event.body)
                 outlets = self._check_outlets_by_names(outlet_names) if outlet_names else self._outlets
 
         if not outlets:
@@ -402,10 +399,7 @@ class Flow:
             event._cyclic_counter = {self.name: 1}
 
     def get_iteration_counter(self, event):
-        if not hasattr(event, "_cyclic_counter"):
-            return 0
-        else:
-            return event._cyclic_counter.get(self.name, 0)
+        return getattr(event, "_cyclic_counter", {}).get(self.name, 0)
 
     def select_outlets(self, event) -> Optional[Collection[str]]:
         """
@@ -511,6 +505,9 @@ class _UnaryFunctionFlow(Flow):
         if fn_select_outlets and not callable(fn_select_outlets):
             raise TypeError(f"Expected fn_select_outlets to be callable, got {type(fn)}")
         self._outlets_selector = fn_select_outlets
+        self._create_name_to_outlet = self._outlets_selector is not None or self._method_is_overridden(
+            "select_outlets", _UnaryFunctionFlow
+        )
 
     async def _call(self, element, fn, pass_kwargs=True):
         if self._long_running:
@@ -538,17 +535,11 @@ class _UnaryFunctionFlow(Flow):
             fn_result = await self._call(element, self._fn)
             await self._do_internal(event, fn_result)
 
-    async def select_outlets(self, event_body) -> Optional[Collection[str]]:
+    def select_outlets(self, event_body) -> Optional[Collection[str]]:
         if self._outlets_selector:
             return self._outlets_selector(event_body)
         else:
             return super().select_outlets(event_body)
-
-    def _init(self):
-        self._create_name_to_outlet = self._outlets_selector is not None or self._method_is_overridden(
-            "select_outlets", _UnaryFunctionFlow
-        )
-        super()._init()
 
 
 class DropColumns(Flow):
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -77,6 +77,7 @@ class Flow:
             self.name = type(self).__name__
 
         self._closeables = []
+        self._selected_outlet: Optional[list[str]] = None
 
     def _init(self):
         self._termination_received = 0
@@ -277,10 +278,14 @@ class Flow:
         if outlets:
             outlets = outlets
         elif event is not _termination_obj:
-            if asyncio.iscoroutinefunction(self.select_outlets):
-                outlet_names = await self.select_outlets(event.body)
+            if self._selected_outlet:
+                outlet_names = self._selected_outlet
+                self._selected_outlet = None
             else:
-                outlet_names = self.select_outlets(event.body)
+                if asyncio.iscoroutinefunction(self.select_outlets):
+                    outlet_names = await self.select_outlets(event.body)
+                else:
+                    outlet_names = self.select_outlets(event.body)
             outlets = self._check_outlets_by_names(outlet_names) if outlet_names else self._outlets
         else:
             outlets = self._outlets
@@ -403,6 +408,15 @@ class Flow:
         """
         return None
 
+    def set_next_outlets(self, outlet_names: Union[str, list[str]]):
+        """
+        Set the next outlets to which the event will be sent. This method can be used in conjunction with
+        select_outlets() to dynamically determine the outlets for the next event.
+
+        :param outlet_names: A collection of outlet names to which the next event should be sent.
+        """
+        self._selected_outlet = outlet_names if isinstance(outlet_names, list) else [outlet_names]
+
     def _check_outlets_by_names(self, outlet_names: Collection[str]) -> list["Flow"]:
         outlets = []
 
@@ -448,7 +462,7 @@ class Choice(Flow):
     def _init(self):
         super()._init()
         # TODO: hacky way of supporting mlrun preview, which replaces targets with a DFTarget
-        self._passthrough_for_preview = list(self._name_to_outlet) == ["dataframe"] if self._name_to_outlet else False
+        self._passthrough_for_preview = list(self._name_to_outlet) == ["dataframe"]
 
     async def _do(self, event):
         if event is _termination_obj:
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -22,7 +22,6 @@ import os
 import pickle
 import time
 import traceback
-import typing
 import uuid
 from asyncio import Task
 from collections import defaultdict
@@ -401,7 +400,7 @@ class Flow:
         else:
             return event._cyclic_counter.get(self.name, 0)
 
-    def select_outlets(self, event) -> typing.Optional[Collection[str]]:
+    def select_outlets(self, event) -> Optional[Collection[str]]:
         """
         Override this method to route events based on a customer logic. The default implementation will route all
         events to all outlets.
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -725,11 +725,7 @@ class MapClass(Flow):
             raise ValueError("long_running=True cannot be used in conjunction with a coroutine do()")
         self._long_running = long_running
         self._filter = False
-
-    def _init(self):
-        # Ensure _name_to_outlet is built to support set_next_outlets()
         self._create_name_to_outlet = True
-        super()._init()
 
     def filter(self):
         # used in the .do() code to signal filtering
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -22,6 +22,7 @@ import os
 import pickle
 import time
 import traceback
+import typing
 import uuid
 from asyncio import Task
 from collections import defaultdict
@@ -44,7 +45,6 @@ class Flow:
         recovery_step=None,
         termination_result_fn=lambda x, y: x if x is not None else y,
         context=None,
-        break_step=None,
         **kwargs,
     ):
         self._outlets = []
@@ -57,14 +57,6 @@ class Flow:
                     step._inlets.append(self)
             else:
                 recovery_step._inlets.append(self)
-
-        self._break_step = break_step
-        if break_step:
-            if isinstance(break_step, dict):
-                for step in break_step.values():
-                    step._inlets.append(self)
-            else:
-                break_step._inlets.append(self)
 
         self._termination_result_fn = termination_result_fn
         self.context = context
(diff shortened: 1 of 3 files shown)


tests/test_flow.py
```diff
@@ -3897,6 +3898,43 @@ def test_flow_reuse():
         assert result == 55
 
 
+def test_flow_reuse_with_cycle():
```
gtopper: This test doesn't seem to create a cycle in the flow, nor does it seem to reuse it?

tests/test_flow.py
```diff
@@ -5322,7 +5360,7 @@ def __init__(self, executor: RunnableExecutor):
     termination_result = controller.await_termination()
     end = time.monotonic()
 
-    assert end - start < 4
+    assert end - start < 5
```
gtopper: The point of the test is to demonstrate that running four 1-second tasks in parallel takes less than 4 seconds.

tests/test_flow.py
```diff
@@ -5464,3 +5502,175 @@ async def async_test_error_raising_batch_target():
...
+
+    if iterations == 5:
+        awaitable_result = controller.emit(1)
+        assert awaitable_result.await_result() == iterations + 1
+    else:
+        if with_recovery:
+            awaitable_result = controller.emit(1)
+            assert awaitable_result.await_result() == -1
+        else:
+            with pytest.raises(RuntimeError):
```
gtopper: Would be nice to check the error message with `match="..."`.

tests/test_flow.py
```diff
@@ -5464,3 +5502,175 @@ async def async_test_error_raising_batch_target():
...
+    my_loop._outlets.append(counter)
+    counter_2.to(my_loop_2)
+    my_loop_2.to(end)
+    my_loop_2._outlets.append(counter_2)
+    end.to(Complete())
+    controller = source.run()
+
+    awaitable_result = controller.emit(1)
+    assert awaitable_result.await_result() == 11
+    controller.terminate()
```
gtopper, line 5605:
```suggestion
    controller.terminate()
    controller.await_termination()
```

tests/test_flow.py
```diff
@@ -5464,3 +5502,175 @@ async def async_test_error_raising_batch_target():
...
+    else:
+        if with_recovery:
+            awaitable_result = controller.emit(1)
+            assert awaitable_result.await_result() == -1
+        else:
+            with pytest.raises(RuntimeError):
+                awaitable_result = controller.emit(1)
+                awaitable_result.await_result()
+
+    controller.terminate()
```
gtopper, line 5573:
```suggestion
    controller.terminate()
    controller.await_termination()
```

storey/flow.py
```diff
@@ -74,10 +76,25 @@ def __init__(
...
 
         self._closeables = []
+        self._selected_outlet: Optional[list[str]] = None
+        self._create_name_to_outlet = True
 
     def _init(self):
         self._termination_received = 0
         self._termination_result = None
+        self._name_to_outlet = {}
+        if self._method_is_overridden("select_outlets", Flow) and self._create_name_to_outlet:
```
gtopper, line 86: A bit confused about this logic, but I guess we can just always create it in `__init__`? It's just a one time creation of a small dictionary. Doesn't seem like something we need to optimize...

storey/flow.py
```diff
@@ -74,10 +76,25 @@ def __init__(
...
+            self._init_name_to_outlet()
+
+    def _init_name_to_outlet(self):
+        for outlet in self._outlets:
+            if outlet.name in self._name_to_outlet:
+                raise ValueError(f"Ambiguous outlet name '{outlet.name}' in Choice step")
```
gtopper: What Choice step? :)

```suggestion
                raise ValueError(f"Ambiguous outlet name '{outlet.name}' in step '{self.name}'")
```

storey/flow.py
davesh0812: the problem is that storey allowing the same name for diffrent outlets.. so only when needed I creating this dictionary and checking that I dont have any ambiguity

storey/flow.py
```diff
@@ -320,21 +364,97 @@ def _user_fn_output_to_event(self, event, fn_result):
...
             for step in self._recovery_step.values():
-                if step._check_step_in_flow(type_to_check):
+                if step._check_step_in_flow(type_to_check, visited):
                     return True
         return False
 
+    def check_and_update_iteration_number(self, event) -> Optional[Callable]:
+        if (
+            hasattr(event, "_cyclic_counter")
+            and isinstance(event._cyclic_counter, dict)
```
gtopper: Should be safe to assume this?

storey/flow.py
```diff
@@ -320,21 +364,97 @@ def _user_fn_output_to_event(self, event, fn_result):
...
+            if counter >= self._max_iteration:
+                raise RuntimeError(
+                    f"Event {event.id} exceeded the maximum iteration count of {self._max_iteration} in step "
+                    f"'{self.name}'."
+                )
+            event._cyclic_counter[self.name] = counter + 1
+        else:
+            event._cyclic_counter = {self.name: 1}
+
+    def _get_iteration_counter(self, event):
```
gtopper: Not used anywhere.

tests/test_flow.py
davesh0812: I dont have the event id - so I can't match the error

storey/flow.py
davesh0812: will use now

storey/flow.py
davesh0812: which one of them?
the dict? `and isinstance(event._cyclic_counter, dict)`?

storey/flow.py
davesh0812: also can usede by a user - I can change the name not to start with '_' what do you think?

storey/flow.py
```diff
@@ -320,21 +364,97 @@ def _user_fn_output_to_event(self, event, fn_result):
...
+        if not hasattr(event, "_cyclic_counter"):
+            return 0
+        else:
+            return event._cyclic_counter.get(self.name, 0)
+
+    def select_outlets(self, event) -> Optional[Collection[str]]:
+        """
+        Override this method to route events based on a customer logic. The default implementation will route all
```
```suggestion
        Override this method to route events based on a custom logic. The default implementation will route all
```

tests/test_flow.py
gtopper: It's a regex :)

storey/flow.py
gtopper: You mean it allows different steps to have the same name? I guess it's not enforced in storey itself (it's enforced in mlrun), but I wouldn't say that this is something we should expressly allow, and certainly not go out of our way to allow.

storey/flow.py
gtopper: Yes, the line I commented on :) – `and isinstance(event._cyclic_counter, dict)`
I don't see any reason that it would be anything other than a dict.

storey/flow.py
gtopper: Not sure how useful it will be to users, but I don't have a strong opinion either way.

storey/flow.py
```diff
@@ -452,9 +553,21 @@ async def _do(self, event):
             return await self._do_downstream(_termination_obj)
         else:
             element = self._get_event_or_body(event)
-            fn_result = await self._call(element)
+            fn_result = await self._call(element, self._fn)
             await self._do_internal(event, fn_result)
 
+    async def select_outlets(self, event_body):
+        if self._outlets_selector:
+            return await self._call(event_body, self._outlets_selector, pass_kwargs=False)
```
gtopper: Seems to me that we should just call `self._outlets_selector()` directly? `self._call()` is for the actual work.

storey/flow.py
davesh0812: but we have a lot of test to change if we want to restrict this.. so I think for now its better to have this logic.

gtopper (changes requested):
CI hangs

storey/flow.py
gtopper: We don't really need `self._method_is_overridden("select_outlets", Flow)` though, do we?

storey/flow.py
```diff
@@ -187,16 +204,28 @@ def _get_recovery_step(self, exception):
         else:
             return self._recovery_step
 
-    def run(self):
+    def run(self, visited=None):
```
gtopper, line 207: Method signature needs to be consistent across the class hierarchy.

storey/sources.py
```diff
@@ -388,7 +388,7 @@ def _emit(self, event):
 
     def run(self):
         """Starts the flow"""
-        self._closeables = super().run()
+        self._closeables = super().run(visited=set())
```
gtopper: `visited=set()` is the default anyway.

storey/sources.py
```diff
@@ -712,7 +712,7 @@ async def _emit(self, event):
...
```
gtopper: `visited=set()` is the default anyway.

storey/flow.py
```diff
@@ -187,16 +204,28 @@ def _get_recovery_step(self, exception):
...
         if not self._legal_first_step and not self._runnable:
             raise ValueError("Flow must start with a source")
+
+        # Initialize visited set once at the top (only for the root call)
+        if visited is None:
+            visited = set()
+
+        # Detect cycles: if we've already visited this step, don't run it again
+        if self in visited:
+            return None
```
gtopper: Should be consistent with the rest of the function:
```suggestion
            return []
```

storey/flow.py
```diff
@@ -187,16 +204,28 @@ def _get_recovery_step(self, exception):
...
+        visited.add(self)
+
         outlets = []
         outlets.extend(self._outlets)
         outlets.extend(self._get_recovery_steps())
         for outlet in outlets:
             outlet._runnable = True
-            self._closeables.extend(outlet.run())
+            outlet_closeables = outlet.run(visited)
+            if outlet_closeables:
```
gtopper: Not needed once consistent as per my previous comment.

storey/flow.py
gtopper: It's also a bit unfortunate to put this is public API. I considered maybe we could make it an attribute, but that introduces a vector for bugs on reuse.

storey/flow.py
```diff
@@ -74,10 +76,25 @@ def __init__(
             self.name = type(self).__name__
 
         self._closeables = []
+        self._selected_outlet: Optional[list[str]] = None
```
gtopper: Selected outlet or outlets?

storey/flow.py
```diff
@@ -256,7 +286,21 @@ def _should_terminate(self):
         return self._termination_received == len(self._inlets)
 
     async def _do_downstream(self, event, outlets=None):
-        outlets = self._outlets if outlets is None else outlets
+        if outlets:
+            outlets = outlets
+        elif event is not _termination_obj:
```
gtopper: Better to avoid negation where possible.

storey/flow.py
```diff
@@ -256,7 +286,21 @@ def _should_terminate(self):
         return self._termination_received == len(self._inlets)
 
     async def _do_downstream(self, event, outlets=None):
-        outlets = self._outlets if outlets is None else outlets
+        if outlets:
+            outlets = outlets
```
gtopper: This statement is a no-op.

storey/flow.py
```diff
@@ -320,21 +364,90 @@ def _user_fn_output_to_event(self, event, fn_result):
...
+    def set_next_outlets(self, outlet_names: Union[str, list[str]]):
+        """
+        Set the next outlets to which the event will be sent. This method can be used in conjunction with
+        select_outlets() to dynamically determine the outlets for the next event.
+
+        :param outlet_names: A collection of outlet names to which the next event should be sent.
+        """
+        if not self._name_to_outlet:
+            self._init_name_to_outlet()
+        self._selected_outlet = outlet_names if isinstance(outlet_names, list) else [outlet_names]
```
gtopper: Using an attribute like this isn't actually safe. It assumes that each step handles events one by one, which isn't necessarily the case. It's true of `Map`, not true of something like concurrent execution, parallel execution, batch, targets, etc... If this facility is only relevant to certain types of steps, the functionality could go into a new intermediate class in the hierarchy (if needed).

davesh0812: delete for now

storey/flow.py
```diff
@@ -256,7 +285,20 @@ def _should_terminate(self):
...
+                outlets = self._outlets
+            else:
+                if self._selected_outlets:
+                    outlet_names = self._selected_outlets
+                    self._selected_outlets = None
+                else:
+                    if asyncio.iscoroutinefunction(self.select_outlets):
+                        outlet_names = await self.select_outlets(event.body)
+                    else:
+                        outlet_names = self.select_outlets(event.body)
```
gtopper: Better to look at the result with `asyncio.iscoroutine()` than inspecting the code with `asyncio.iscoroutinefunction()`. Mainly because it's actually legal for an ordinary, non-async function to return a coroutine (basically as a passthrough).

storey/flow.py
```diff
@@ -320,21 +362,79 @@ def _user_fn_output_to_event(self, event, fn_result):
...
+                raise RuntimeError(f"Max iterations exceeded in step '{self.name}' for event {event.id}")
+            event._cyclic_counter[self.name] = counter + 1
+        else:
+            event._cyclic_counter = {self.name: 1}
+
+    def get_iteration_counter(self, event):
+        if not hasattr(event, "_cyclic_counter"):
+            return 0
+        else:
+            return event._cyclic_counter.get(self.name, 0)
```
gtopper: A more idiomatic way of doing this would be:
```suggestion
        counter = getattr(event, "_cyclic_counter", {}).get(self.name, 0)
```
which I think might as well be inlined.

storey/flow.py
```diff
@@ -452,9 +535,21 @@ async def _do(self, event):
             return await self._do_downstream(_termination_obj)
         else:
             element = self._get_event_or_body(event)
-            fn_result = await self._call(element)
+            fn_result = await self._call(element, self._fn)
             await self._do_internal(event, fn_result)
 
+    async def select_outlets(self, event_body) -> Optional[Collection[str]]:
```
```suggestion
    def select_outlets(self, event_body) -> Optional[Collection[str]]:
```

storey/flow.py
```diff
@@ -452,9 +535,21 @@ async def _do(self, event):
...
+        if self._outlets_selector:
+            return self._outlets_selector(event_body)
+        else:
+            return super().select_outlets(event_body)
+
+    def _init(self):
+        self._create_name_to_outlet = self._outlets_selector is not None or self._method_is_overridden(
+            "select_outlets", _UnaryFunctionFlow
+        )
+        super()._init()
```
gtopper: We really don't need to make allowance for the user subclassing something like Map and overriding its `select_outlets`. Better give the user one way to do things. In this case, via the `outlets_selector` parameter.
```suggestion
```

storey/flow.py
```diff
@@ -74,10 +76,25 @@ def __init__(
             self.name = type(self).__name__
 
         self._closeables = []
+        self._selected_outlets: Optional[list[str]] = None
+        self._create_name_to_outlet = True
```
gtopper, line 80: I don't think we need this one either.

storey/flow.py
gtopper:
```suggestion
        if self.select_outlets is not Flow.select_outlets:
```

storey/flow.py
```diff
@@ -640,6 +726,11 @@ def __init__(self, long_running=None, **kwargs):
         self._long_running = long_running
         self._filter = False
 
+    def _init(self):
+        # Ensure _name_to_outlet is built to support set_next_outlets()
+        self._create_name_to_outlet = True
+        super()._init()
```
```suggestion
```

storey/flow.py
```diff
@@ -421,26 +491,36 @@ async def _do(self, event):
...
-        if self._is_async and long_running:
+        if asyncio.iscoroutinefunction(fn) and long_running:
             raise ValueError("long_running=True cannot be used in conjunction with a coroutine")
         self._long_running = long_running
         self._fn = fn
         self._pass_context = pass_context
+        if fn_select_outlets and not callable(fn_select_outlets):
+            raise TypeError(f"Expected fn_select_outlets to be callable, got {type(fn)}")
+        self._outlets_selector = fn_select_outlets
+        self._create_name_to_outlet = self._outlets_selector is not None or self._method_is_overridden(
```
```suggestion
        self._create_name_to_outlet = self._outlets_selector or self._method_is_overridden(
```

gtopper (approved):
LGTM otherwise

PR #594, opened 2025-12-09
[Table] Fix memory leak when Table is reused across flow cycles

alxtkr77:
When a Table is reused across multiple flow cycles (e.g., Kafka rebalancing
in model monitoring), internal caches like _attrs_cache, _aggregates, and
_schema were not being cleared on termination, causing memory to grow
linearly with each cycle.

Changes:
- Clear all internal caches in Table._terminate() to prevent memory leak
- Initialize _worker_awaitable in __init__ to avoid AttributeError
- Call _terminate() from close() to ensure cleanup happens
- Cancel _Batching._timeout_task on flow termination to prevent task leak

The fix ensures memory stays constant when reusing Table instances, which
is critical for long-running streaming applications.

ML-11518

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1134,6 +1134,14 @@ class _Batching(Flow):
 
     async def _do(self, event):
         if event is _termination_obj:
+            # Cancel _timeout_task to prevent memory leak (ML-11518)
+            if self._timeout_task is not None:
+                self._timeout_task.cancel()
+                try:
+                    await self._timeout_task
+                except asyncio.CancelledError:
+                    pass
+                self._timeout_task = None
             await self._emit_all()
             await self._terminate()
             return await self._do_downstream(_termination_obj)
--- a/storey/table.py
+++ b/storey/table.py
@@ -68,6 +68,7 @@ class Table:
         self._aggregations_read_only = False
         self._use_windows_from_schema = False
         self._q = None
+        self._worker_awaitable = None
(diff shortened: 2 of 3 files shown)


storey/flow.py
```diff
@@ -1134,6 +1134,14 @@ async def _terminate(self):
 
     async def _do(self, event):
         if event is _termination_obj:
+            # Cancel _timeout_task to prevent memory leak (ML-11518)
```
gtopper, line 1137: It's good to cancel it, but it doesn't seem like it would be the cause of a leak, as the task would just time out. I would be more worried that it could affect the next run – i.e. cause a logical bug rather than a leak. As such, I doubt it relates to ML-11518.

storey/flow.py
```diff
@@ -1134,6 +1134,14 @@ async def _terminate(self):
...
+            if self._timeout_task is not None:
```
gtopper, line 1138:
```suggestion
            if self._timeout_task:
```

storey/table.py
```diff
@@ -179,6 +180,7 @@ def _init_flush_task(self):
             self._flush_task = asyncio.get_running_loop().create_task(self._flush_worker())
 
     async def close(self):
+        await self._terminate()
```
gtopper, line 183: If we call it here, I guess we should remove the call from `NoSqlTarget.do()`?

PR #600, opened 2025-12-22
[Utils] fix `find_partitions`

royischoss:
Avoid breaking change in storey 1.10.x by introducing a flag for changing the outputs of `find_partitions`


davesh0812 (approved):
LGTM

PR #605, opened 2026-01-20
Add support for stream responses

gtopper:
[ML-11875](https://iguazio.atlassian.net/browse/ML-11875)
[ML-11877](https://iguazio.atlassian.net/browse/ML-11877)

[ML-11875]: https://iguazio.atlassian.net/browse/ML-11875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ML-11877]: https://iguazio.atlassian.net/browse/ML-11877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #607, opened 2026-01-26
Fix `ParallelExecution` API breakage

gtopper:
#605 broke mlrun test `serving.test_async_flow.test_model_runner_with_selector`.

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -2155,7 +2155,15 @@ class ParallelExecution(Flow, _StreamingStepMixin):
                 return None
 
         # Non-streaming path
-        if len(runnables) == 1:
+        # Check if any results are generators (not allowed with multiple runnables)
+        for result in results:
+            if _is_generator(result):
+                raise StreamingError(
+                    "Streaming is not supported when multiple runnables are selected. "
+                    "Streaming runnables must be the only runnable selected for an event."
+                )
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -2155,7 +2155,15 @@ async def _do(self, event):
...
+        # Check if any results are generators (not allowed with multiple runnables)
+        for result in results:
+            if _is_generator(result):
+                raise StreamingError(
+                    "Streaming is not supported when multiple runnables are selected. "
+                    "Streaming runnables must be the only runnable selected for an event."
+                )
+        # Use self.runnables (registered) not runnables (selected) to determine wrapping
+        if len(self.runnables) == 1:
             result: _ParallelExecutionRunnableResult = results[0]
```
royischoss, line 2167: might cause index error if results is empty

royischoss (commented):
Small comment

storey/flow.py
gtopper: I think you're right, but it's been there [since at least 1.10.0](https://github.com/mlrun/storey/blob/v1.10.0/storey/flow.py#L1667), so let's leave it for another PR.

gtopper:
Merging to unbreak mlrun.

storey/flow.py
gtopper: Done in #608.

PR #610, opened 2026-01-28
Support `ParallelExecution` with batch

tomerm-iguazio:
Supports batching with `ParallelExecution`, including sub_events handling on the first level and deepcopy handling.

Ticket:
[ML-12067](https://iguazio.atlassian.net/browse/ML-12067)

[ML-12067]: https://iguazio.atlassian.net/browse/ML-12067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -332,7 +332,9 @@ class Flow:
         sub_event_copies = []
         if is_batched:
             for sub_event in event.body:
-                sub_event_copies.append(self._deepcopy_event(sub_event, sub_event, False, False))
+                sub_event_copies.append(
+                    self._deepcopy_event(sub_event, sub_event, is_stream_completion=False, is_batched=False)
+                )
                 sub_event._awaitable_result = None
                 sub_event._original_events = None
         event_copy = copy.deepcopy(event)
@@ -387,7 +389,9 @@ class Flow:
             )
 
             for i in range(1, len(outlets)):
-                event_copy = self._deepcopy_event(event, target_obj, is_stream_completion, is_batched)
+                event_copy = self._deepcopy_event(
+                    event, target_obj, is_stream_completion=is_stream_completion, is_batched=is_batched
+                )
                 tasks.append(asyncio.get_running_loop().create_task(outlets[i]._do_and_recover(event_copy)))
         if self.verbose and self.logger:
             step_name = self.name
@@ -2219,7 +2223,9 @@ class ParallelExecution(Flow, _StreamingStepMixin):
             event_bodies = []
             for sub_event in event.body:
                 # copy sub events for avoiding overriding original sub events
-                sub_event_copy = self._deepcopy_event(sub_event, sub_event, False, False)
+                sub_event_copy = self._deepcopy_event(
+                    sub_event, sub_event, is_stream_completion=False, is_batched=False
+                )
                 sub_events_to_modify.append(sub_event_copy)
                 # for the invocation, we only want to pass the body
                 event_bodies.append(copy.deepcopy(sub_event.body))
@@ -2304,7 +2310,7 @@ class ParallelExecution(Flow, _StreamingStepMixin):
 
         if is_full_event_batched:
             for sub_event in event.body:
-                self.set_event_metadata(sub_event, metadata)
+                self.set_event_metadata(sub_event, copy.deepcopy(metadata))
         self.set_event_metadata(event, metadata)
         return await self._do_downstream(event)
 
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -2193,7 +2193,7 @@ class ParallelExecution(Flow, _StreamingStepMixin):
                 )
             runnables_encountered.add(id(runnable))
             futures.append(future)
-        results: list[_ParallelExecutionRunnableResult] = await asyncio.gather(*futures)
+        results = await asyncio.gather(*futures)
         # Check for streaming response (only when a single runnable is selected)
         if len(runnables) == 1 and results:
             result = results[0]
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -319,6 +319,9 @@ class Flow:
                            (either the event itself or event.original_event for StreamCompletion).
         :param is_stream_completion: If True, copy target is event_copy.original_event,
                                      otherwise it's event_copy itself.
+        :param is_batched: If True, performs nested copying for batched events where event.body contains
+                          a list of sub-events. Only 1 layer of batching is supported. Not supported with
+                          StreamCompletion events.
 
         :returns: The deepcopied event with unpicklable attributes restored.
         """
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -347,7 +347,7 @@ class Flow:
         target_obj._original_events = original_events
         if is_batched:
             event_copy.body = sub_event_copies
-            for sub_event, sub_event_copy in zip(event, sub_event_copies):
+            for sub_event, sub_event_copy in zip(event.body, sub_event_copies):
                 sub_event._awaitable_result = sub_event_copy._awaitable_result
                 sub_event._original_events = sub_event_copy._original_events
         return event_copy
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -2208,18 +2208,18 @@ class ParallelExecution(Flow, _StreamingStepMixin):
                 await self._emit_streaming_chunks(event, result)
                 return None
 
-            # Non-streaming path
-            # Check if any results are generators (not allowed with multiple runnables)
+        # Non-streaming path
+        # Check if any results are generators (not allowed with multiple runnables)
         for result in results:
             if _is_generator(result):
                 raise StreamingError(
                     "Streaming is not supported when multiple runnables are selected. "
                     "Streaming runnables must be the only runnable selected for an event."
                 )
-            # If no runnables were selected, don't emit the event
+        # If no runnables were selected, don't emit the event
         if not results:
             return None
-            # Use self.runnables (registered) not runnables (selected) to determine wrapping
+        # Use self.runnables (registered) not runnables (selected) to determine wrapping
         if len(self.runnables) == 1:
             result: _ParallelExecutionRunnableResult = results[0]
             event.body = result.data
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -87,6 +87,15 @@ class Flow:
         if self._method_is_overridden("select_outlets", Flow) and self._create_name_to_outlet:
             self._init_name_to_outlet()
 
+    @staticmethod
+    def _prepare_event_for_deepcopy(event):
+        # Temporarily delete self-reference to avoid deepcopy getting stuck in an infinite loop
+        awaitable_result = event._awaitable_result
+        event._awaitable_result = None
+        original_events = getattr(event, "_original_events", None)
+        event._original_events = None
+        return awaitable_result, original_events
+
     def _init_name_to_outlet(self):
         for outlet in self._outlets:
             if outlet.name in self._name_to_outlet:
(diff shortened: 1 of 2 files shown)


tests/test_flow.py
```diff
@@ -2334,6 +2372,122 @@ def reduce_fn(acc, x):
...
+        execution_mechanism_by_runnable_name={
+            "multiply": "naive",
+            "add": "naive",
+            "now": "naive",
+        },
+    )
+    flat_map = FlatMap(fn=lambda x: x.body, full_event=True)
+    reducer = Reduce([], lambda acc, x: append_and_return(acc, x))
+    source.to(batch_step).to(parallel_execution)
+    parallel_execution.to(flat_map).to(reducer)
```
gtopper: Probably clearer to read in `build_flow` form.

tests/test_flow.py
```diff
@@ -2334,6 +2372,122 @@ def reduce_fn(acc, x):
...
+    number_of_events = 10
+
+    runnables = [
+        RunnableMultiplyBy2("multiply"),
+        RunnableAdd10("add"),
+        RunnableGetNow("now"),
+    ]
+
+    source = SyncEmitSource()
+    batch_step = Batch(batch_size, 100, full_event=True)
```
gtopper: Better use named parameters.

tests/test_flow.py
```diff
@@ -101,6 +103,42 @@ def raise_ex(self, element):
...
+            return [sub_value + 10 for sub_value in data]
+        return data + 10
+
+
+class RunnableGetNow(ParallelExecutionRunnable):
+    def init(self):
+        pass
+
+    def run(self, data, path, origin_name=None):
+        now = datetime.now(timezone.utc)
```
gtopper: It's better as a rule not to use now, because it changes from run to run, potentially causing nondeterministic behaviour.

storey/flow.py
```diff
@@ -2181,6 +2149,25 @@ async def _do(self, event):
             return await self._do_downstream(event)
 
         event = self.preprocess_event(event)
+        sub_events_to_modify = []
+        is_full_event_batched = (
+            not isinstance(event, StreamCompletion)
+            and isinstance(getattr(event, "body", None), list)
+            and event.body
+            and all("event" in sub_event.__class__.__name__.lower() for sub_event in event.body)
```
gtopper: No one said an event has to have the word "event" in its class name, though. Better go by whether it has a `body` attribute. Also, it's probably enough to just check the first one.

storey/flow.py
```diff
@@ -327,13 +330,26 @@ def _deepcopy_event_for_outlet(self, event, target_obj, is_stream_completion: bo
         original_events = getattr(target_obj, "_original_events", None)
         target_obj._original_events = None
 
+        sub_event_copies = []
+        if is_batched:
+            for sub_event in event.body:
+                if isinstance(sub_event, StreamCompletion):
+                    raise ValueError("batching is not supported with streaming")
```
gtopper: Seems like you already check for `StreamCompletion` earlier in the flow, so this can't actually happen.

storey/flow.py
```diff
@@ -311,14 +311,17 @@ def _event_string(event):
...
         """Deepcopy event while handling unpicklable attributes on target_obj.
 
         :param event: The event to deepcopy.
         :param target_obj: The object containing _awaitable_result and _original_events
                            (either the event itself or event.original_event for StreamCompletion).
         :param is_stream_completion: If True, copy target is event_copy.original_event,
                                      otherwise it's event_copy itself.
+        :param is_batched: If True, performs nested copying for batched events where event.body contains
+                          a list of sub-events. Only 1 layer of batching is supported. Not supported with
+                          StreamCompletion events.
```
gtopper: Probably no need to mention StreamCompletion here, just as we don't mention termination object.

tests/test_flow.py
```diff
@@ -2334,6 +2372,122 @@ def reduce_fn(acc, x):
...
+
+    source.to(batch_step).to(parallel_execution)
+    parallel_execution.to(flat_map1).to(collector)
+    parallel_execution.to(flat_map2).to(collector)
+    collector.to(reducer)
+
+    controller = source.run()
+
+    for i in range(number_of_events):
+        sleep(0.2)
```
gtopper: Why do we need to sleep here, though?

tests/test_flow.py
```diff
@@ -2334,6 +2372,122 @@ def reduce_fn(acc, x):
...
+        execution_mechanism_by_runnable_name={
+            "multiply": "naive",
+            "add": "naive",
+            "now": "naive",
+        },
+    )
+
+    flat_map1 = FlatMap(fn=lambda x: x.body, full_event=True)
+    flat_map2 = FlatMap(fn=lambda x: x.body, full_event=True)
+    collector = Collector(expected_completions=2)
```
gtopper: Why collector? This test isn't related to streaming.

gtopper (commented):
Quick preliminary review. Please add a link to the JIRA for some context.

storey/flow.py
tomerm-iguazio: I will check it all just in case, but it is ok

tests/test_flow.py
```diff
@@ -101,6 +101,42 @@ def raise_ex(self, element):
...
+            return [sub_value + 10 for sub_value in data]
+        return data + 10
+
+
+class RunnableGetRandom(ParallelExecutionRunnable):
+    def init(self):
+        pass
+
+    def run(self, data, path, origin_name=None):
+        random_uuid = str(uuid.uuid4())
```
gtopper, line 123: It's better not to use (unseeded) random in tests, as it can cause nondeterministic results.

tests/test_flow.py
```diff
@@ -101,6 +101,42 @@ def raise_ex(self, element):
...
+    def run(self, data, path, origin_name=None):
+        if isinstance(data, list):
+            return [sub_value + 10 for sub_value in data]
+        return data + 10
+
+
+class RunnableGetRandom(ParallelExecutionRunnable):
+    def init(self):
+        pass
+
```
```suggestion
```

tests/test_flow.py
```diff
@@ -2334,6 +2370,120 @@ def reduce_fn(acc, x):
...
+    number_of_events = 10
+
+    runnables = [
+        RunnableMultiplyBy2("multiply"),
+        RunnableAdd10("add"),
+        RunnableGetRandom("uuid"),
+    ]
+
+    source = SyncEmitSource()
+    batch_step = Batch(max_events=batch_size, flush_after_seconds=4, full_event=True)
```
gtopper: Do we need `flush_after_seconds` here?

storey/flow.py
gtopper: It's not very important. Just seems like unnecessary extra processing.

storey/flow.py
```diff
@@ -327,13 +329,24 @@ def _deepcopy_event_for_outlet(self, event, target_obj, is_stream_completion: bo
         original_events = getattr(target_obj, "_original_events", None)
         target_obj._original_events = None
 
+        sub_event_copies = []
+        if is_batched:
+            for sub_event in event.body:
+                sub_event_copies.append(self._deepcopy_event(sub_event, sub_event, False, False))
```
gtopper: Better use named parameters for the boolean flags.

storey/flow.py
```diff
@@ -327,13 +329,24 @@ def _deepcopy_event_for_outlet(self, event, target_obj, is_stream_completion: bo
...
+                sub_event._awaitable_result = None
+                sub_event._original_events = None
```
gtopper, line 348: Maybe I'm missing something, but aren't we going to end up with both of these fields being `None`? Messing up `Complete` and explicit ack?

storey/flow.py
```diff
@@ -366,8 +379,15 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
+            is_batched = (
+                not is_stream_completion
+                and isinstance(getattr(event, "body", None), list)
+                and event.body
+                and all(hasattr(sub_event, "body") for sub_event in event.body)
+            )
+
             for i in range(1, len(outlets)):
-                event_copy = self._deepcopy_event_for_outlet(event, target_obj, is_stream_completion)
+                event_copy = self._deepcopy_event(event, target_obj, is_stream_completion, is_batched)
```
gtopper: Better use named parameters for boolean flags.

storey/flow.py
```diff
@@ -2236,20 +2280,32 @@ async def _do(self, event):
...
+                event.body = {result.runnable_name: result.data for result in results}
 
-        if hasattr(event, "_metadata") and isinstance(event._metadata, dict):
-            event._metadata.update(metadata)
-        else:
-            event._metadata = metadata
+        if is_full_event_batched:
+            for sub_event in event.body:
+                self.set_event_metadata(sub_event, metadata)
+        self.set_event_metadata(event, metadata)
```
gtopper, line 2323: Aren't they all going to end up with references to the same metadata? Potentially causing corruption downstream?

gtopper (commented):
Please add tests for `Complete` (I suspect a bug there), and at least one for explicit ack (I suspect a bug there as well).

storey/flow.py
tomerm-iguazio: we set back these values in the end of the function

tests/test_flow.py
tomerm-iguazio: Ok on this case as we talked on private

tests/test_flow.py
gtopper: Thumbs up but still in diff :)

tests/test_flow.py
gtopper: Right, in this case it's ok because the value doesn't matter, but it's better to avoid.

tests/test_flow.py
```diff
@@ -2334,6 +2366,149 @@ def reduce_fn(acc, x):
...
+        execution_mechanism_by_runnable_name={
+            "multiply": "naive",
+            "add": "naive",
+            "uuid": "naive",
+        },
+    )
+    controller = build_flow(
+        [
+            SyncEmitSource(),
+            Batch(max_events=batch_size, full_event=True, flush_after_seconds=4),
```
gtopper: We discussed removing `flush_after_seconds` and adjusting the test, to shorten the runtime. Alternatively, consider shortening it to one second.

tests/test_flow.py
```diff
@@ -2334,6 +2366,149 @@ def reduce_fn(acc, x):
...
+            parallel_execution,
+            FlatMap(fn=lambda x: x.body, full_event=True),
+            Complete(),
+            Reduce(initial_value=[], fn=lambda acc, x: append_and_return(acc, x)),
+        ]
+    ).run()
+
+    from concurrent.futures import ThreadPoolExecutor, as_completed
```
gtopper: This can be imported at the top.

tests/test_flow.py
```diff
@@ -2334,6 +2366,149 @@ def reduce_fn(acc, x):
...
+    complete = Complete()
+    reducer = Reduce([], lambda acc, x: append_and_return(acc, x))
+
+    source.to(batch_step).to(parallel_execution)
+    parallel_execution.to(flat_map1).to(complete).to(reducer)
+    parallel_execution.to(flat_map2).to(reducer)
+
+    controller = source.run()
+
+    from concurrent.futures import ThreadPoolExecutor, as_completed
```
gtopper: This can be imported at the top.

tests/test_flow.py
```diff
@@ -2334,6 +2366,149 @@ def reduce_fn(acc, x):
...
+        assert "uuid" in result
+        assert result["add"] == 10 + i
+        assert result["multiply"] == i * 2
+        return result
+
+    # Emit events in parallel using ThreadPoolExecutor with batch_size threads
+    with ThreadPoolExecutor(max_workers=batch_size) as executor:
+        futures = [executor.submit(emit_event, i) for i in range(number_of_events)]
+        for future in as_completed(futures):
+            future.result()  # Wait for all to complete and raise any exceptions
```
gtopper: Please terminate in a finally block, then, to terminate also in case of an error.

tests/test_flow.py
```diff
@@ -2334,6 +2366,149 @@ def reduce_fn(acc, x):
...
+        # Verify each event has the expected fields after parallel execution
+        assert "add" in result
+        assert "multiply" in result
+        assert "uuid" in result
+        assert result["add"] == 10 + i
+        assert result["multiply"] == i * 2
+        return result
+
+    # Emit events in parallel using ThreadPoolExecutor with batch_size threads
+    with ThreadPoolExecutor(max_workers=batch_size) as executor:
```
gtopper: Btw, this test would be nicer with an `AsyncEmitSource`, using asyncio and not threads.

gtopper (commented):
LGTM otherwise. Just a handful of outstanding comments.

storey/flow.py
gtopper: I see that you now deepcopy the metadata. Great. Consider adding a test for this.

tests/test_flow.py
tomerm-iguazio: ok

tests/test_flow.py
```diff
@@ -3846,10 +4017,6 @@ async def async_test_async_metadata_fields():
     assert result.body == body
 
 
-def test_async_metadata_fields():
-    asyncio.run(async_test_async_metadata_fields())
```
gtopper, line 3850: Why remove this test? And also keep `async_test_async_metadata_fields` orphaned.

gtopper (approved):

PR #613, opened 2026-02-11
Add timing metadata for streaming `ParallelExecution` responses

gtopper:
[ML-11879](https://iguazio.atlassian.net/browse/ML-11879)

Streaming responses from `ParallelExecution` were missing the when and `microsec` timing metadata that non-streaming responses include. This metadata is required for model monitoring in MLRun.

Changes
* Add `_StreamingResult` class to wrap streaming generators with timing info
* Set timing metadata on events before emitting streaming chunks
* Handle both in-process streaming (`_StreamingResult`) and process-based streaming (raw generators)
* `Collector` calculates total streaming duration (`microsec`) when the stream completes

Notes
* `ParallelExecution` sets `microsec` to `None` initially since total runtime isn't known until streaming completes
* `Collector` calculates the actual `microsec` value (total elapsed time from stream start to completion)
* For process-based streaming, `when` uses the timestamp when chunks start arriving

[ML-11879]: https://iguazio.atlassian.net/browse/ML-11879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1728,6 +1728,20 @@ class _ParallelExecutionRunnableResult:
         self.timestamp = timestamp
 
 
+class _StreamingResult:
+    """Wraps a streaming generator with timing metadata for model monitoring."""
+
+    def __init__(
+        self,
+        runnable_name: str,
+        generator: Generator | AsyncGenerator,
+        timestamp: datetime.datetime,
+    ):
+        self.runnable_name = runnable_name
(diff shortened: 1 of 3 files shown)


royischoss (approved):
LGTM

PR #615, opened 2026-02-22
[Batch] Change batch step default for `full_event` to `True`

tomerm-iguazio:
This PR update Batch step default to full_event=True.
It also updates all Batch steps tests to operate also with full events throughout.

[ML-12099](https://iguazio.atlassian.net/browse/ML-12099)

[ML-12099]: https://iguazio.atlassian.net/browse/ML-12099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1492,6 +1492,12 @@ class Batch(_Batching, WithUUID):
     _do_downstream_per_event = False
 
     def __init__(self, *args, **kwargs):
+        # Set full_event to True by default if not specified
+        if "full_event" not in kwargs:
+            kwargs["full_event"] = True
+        elif kwargs["full_event"] is False:
+            raise ValueError("Batch step supports full_event=True. Setting full_event=False is not supported.")
+
         _Batching.__init__(self, *args, **kwargs)
         WithUUID.__init__(self)
 
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -23,6 +23,7 @@ import pickle
 import time
 import traceback
 import uuid
+import warnings
 from asyncio import Task
 from collections import defaultdict
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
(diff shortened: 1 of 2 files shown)


tests/test_flow.py
```diff
@@ -2279,11 +2314,19 @@ def reduce_fn(acc, event):
         acc.append(event)
         return acc
 
+    def reduce_fn_batch(acc, event):
+        if len(event) == 1 and event[0].body == 1:
+            q.put(None)
+        acc.append([sub_Event.body for sub_Event in event])
```
gtopper: Naming convention:
```suggestion
        acc.append([sub_event.body for sub_event in event])
```

storey/flow.py
```diff
@@ -1492,6 +1493,14 @@ class Batch(_Batching, WithUUID):
     _do_downstream_per_event = False
 
     def __init__(self, *args, **kwargs):
+        # Set full_event to True by default if not specified
+        if "full_event" not in kwargs:
+            warnings.warn(
```
gtopper, line 1498: Missing `FutureWarning`.

storey/flow.py
```diff
@@ -1492,6 +1493,14 @@ class Batch(_Batching, WithUUID):
...
+                "The default value of full_event in Batch changed to True."
```
```suggestion
                "The default value of full_event in Batch changed to True in mlrun 1.11.0."
```

gtopper (commented):
Looks good overall, in line with the decision as cited in https://iguazio.atlassian.net/browse/ML-12099?focusedCommentId=207050.

Please update the PR title to reflect the fact that support isn't removed, but rather deprecated, with the default value changed.

Please make sure this change gets a release note.

Would be good to add a test for the deprecation warning itself, e.g.:
```python
def test_batch_warns_when_full_event_not_specified():
    with pytest.warns(FutureWarning, match="full_event"):
        Batch(4, 100)
```

While this is a concern we can defer, there are tests in storey and mlrun that would fail once `Batch(full_event=False)` is actually blocked. See `test_to_dataframe_with_index` and `test_batch_by_user_key` in storey, and `test_batch` in mlrun for examples.

storey/flow.py
gtopper: Never mind, I missed the fact that we only change the default, not deprecating anything actually in the end.

gtopper (approved):

PR #616, opened 2026-02-23
Propagate streaming errors through `Collector` for monitoring

gtopper:
This fix is needed to complete https://github.com/mlrun/mlrun/pull/9343 ([ML-12182](https://iguazio.atlassian.net/browse/ML-12182)).

When a streaming generator raises an exception, the error was not propagated through the normal graph pipeline — the `Collector` never emitted an event, causing a memory leak in `_collected_streams` and preventing downstream consumers (e.g. model monitoring) from tracking the error.

### Changes
- **`StreamCompletion`**: Add optional `error: str` field to carry the error type and message (e.g. `"ValueError: something failed"`)
- **`_emit_streaming_chunks`**: Use explicit `__anext__()` iteration to separate generator errors from downstream errors. Always send `StreamCompletion` (even on error) so `Collector` can emit and clean up. The generator error is **not** re-raised — the flow stays alive and the error propagates only via `StreamCompletion`.
- **`Collector`**: When `StreamCompletion.error` is set, emit `body={"error": error_string}` instead of collected chunks, matching the format used by `ParallelExecution` for non-streaming errors.
- **`AwaitableResult` / `AsyncAwaitableResult`**: Check `StreamCompletion.error` in the stream generators and raise `StreamingError` wrapping the error string, so consumers are notified of the failure.
- Remove unrelated but erroneous `@pytest.mark.asyncio` annotation

### Tests
- `StreamCompletion` with error field (construction + repr)
- `Collector` error propagation: error dict format, `stream_collected` attribute, cleanup of `_collected_streams`
- Streaming generator errors deliver `StreamingError` to consumers via `AwaitableResult`; `await_termination()` completes normally
- `ParallelExecution` streaming error propagation across execution mechanisms

[ML-12182]: https://iguazio.atlassian.net/browse/ML-12182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_streaming.py
+++ b/tests/test_streaming.py
@@ -92,6 +92,23 @@ class TestStreamingPrimitives:
         assert "StreamCompletion" in repr(completion)
         assert "my_step" in repr(completion)
 
+    def test_stream_completion_with_error(self):
+        """Test StreamCompletion with error string."""
+        event = Event(body="test", id="123")
+        completion = StreamCompletion("test_step", event, error="ValueError: test error")
+        assert completion.streaming_step == "test_step"
+        assert completion.original_event is event
+        assert completion.error == "ValueError: test error"
+
+    def test_stream_completion_with_error_repr(self):
+        """Test StreamCompletion repr includes error string."""
+        event = Event(body="test", id="abc")
+        completion = StreamCompletion("my_step", event, error="RuntimeError: something went wrong")
+        repr_str = repr(completion)
+        assert "StreamCompletion" in repr_str
+        assert "my_step" in repr_str
+        assert "RuntimeError" in repr_str
+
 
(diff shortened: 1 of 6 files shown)


tests/test_streaming.py
tomerm-iguazio, line 1: Can you test that the error is raised on the awaited result, in addition to these tests that check that the error is raised on the termination result?

tomerm-iguazio (changes requested):
Looks great:) small comment

tomerm-iguazio (approved):
Nice work!

PR #617, opened 2026-02-24
Avoid recursion error when running cyclic graph

davesh0812:
https://iguazio.atlassian.net/browse/ML-12017

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -507,7 +507,8 @@ class Flow:
                 raise RuntimeError(
                     f"Step '{self.name}' exceeded the default cycle limit of {effective_max_iterations} iterations "
                     f"for event {event.id}. This typically indicates an infinite cycle without an exit condition. "
-                    f"To fix: 1) Add an exit condition to your cycle, OR 2) Set max_iterations explicitly if this is intentional."
+                    f"To fix: 1) Add an exit condition to your cycle, OR 2) Set max_iterations explicitly if "
+                    f"this is intentional."
                 )
             else:
                 # User explicitly set max_iterations
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -56,10 +56,6 @@ from .queue import AsyncQueue
 from .table import Table
 from .utils import _split_path, get_in, stringify_key, update_in
 
-# Default maximum iterations for cyclic graphs to prevent accidental infinite loops.
-# This provides a safety net when max_iterations is not explicitly set.
-# Users can override by setting max_iterations explicitly (or set to a large number).
-DEFAULT_MAX_ITERATIONS_FOR_CYCLES = 10_000
 
 
 def _is_generator(obj) -> bool:
@@ -118,6 +114,11 @@ class Flow:
         self._closeables = []
         self._selected_outlets: Optional[list[str]] = None
         self._create_name_to_outlet = True
+
+        # Default maximum iterations for cyclic graphs to prevent accidental infinite loops.
+        # This provides a safety net when max_iterations is not explicitly set.
+        # Users can override by setting max_iterations explicitly (or set to a large number).
+        self._default_max_iterations_for_cycle = int(os.environ.get("DEFAULT_MAX_ITERATIONS_FOR_CYCLES") or 10_000)
 
     def _init(self):
         self._closeables = []
@@ -492,7 +493,7 @@ class Flow:
             # Users experiencing this can either:
             # 1. Fix their cycle to have an exit condition, OR
             # 2. Set max_iterations explicitly to a higher value (or very high for "unlimited")
-            effective_max_iterations = DEFAULT_MAX_ITERATIONS_FOR_CYCLES
+            effective_max_iterations = self._default_max_iterations_for_cycle
 
         if counter >= effective_max_iterations:
             if self._max_iterations is None:
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -57,7 +57,6 @@ from .table import Table
 from .utils import _split_path, get_in, stringify_key, update_in
 
 
-
 def _is_generator(obj) -> bool:
     """Check if an object is a sync or async generator."""
     return inspect.isgenerator(obj) or inspect.isasyncgen(obj)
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -400,11 +405,27 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
+
+        # Prevent deep recursion in cyclic graphs
+        iteration_count = getattr(event, "_cyclic_counter", {}).get(self.name, 0)
+        has_multiple_outlets = len(outlets) > 1
+        use_task_to_prevent_recursion = iteration_count > 2 and not has_multiple_outlets
+
+        if use_task_to_prevent_recursion:
+            # Create a task to avoid building a deep await chain
+            task = asyncio.get_running_loop().create_task(outlets[0]._do_and_recover(event))
+            tasks.insert(0, task)  # Add to front so we await it first
```
gtopper: There's no need to insert it at the front if it's the only one.

storey/flow.py
```diff
@@ -400,11 +405,27 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
+        else:
+            # Direct await - ensures errors propagate before any parallel outlets can return
+            await outlets[0]._do_and_recover(event)
+
+        # Await all tasks and collect any exceptions
+        # This ensures errors from background tasks are properly propagated
+        for i, task in enumerate(tasks):
+            if self.verbose and self.logger and (i > 0 or use_task_to_prevent_recursion):
+                if i < len(outlets):
```
gtopper: How can we end up with more tasks than outlets?

storey/flow.py
```diff
@@ -400,11 +405,27 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
+            # Create a task to avoid building a deep await chain
+            task = asyncio.get_running_loop().create_task(outlets[0]._do_and_recover(event))
+            tasks.insert(0, task)  # Add to front so we await it first
+        else:
+            # Direct await - ensures errors propagate before any parallel outlets can return
+            await outlets[0]._do_and_recover(event)
+
+        # Await all tasks and collect any exceptions
+        # This ensures errors from background tasks are properly propagated
+        for i, task in enumerate(tasks):
```
gtopper: I think removing the start=1 will break the log, causing the routing information to be erroneous.

storey/flow.py
```diff
@@ -400,11 +405,27 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
-        await outlets[0]._do_and_recover(event)  # Optimization - avoids creating a task for the first outlet.
-        for i, task in enumerate(tasks, start=1):
-            if self.verbose and self.logger:
-                self.logger.debug(f"{step_name} -> {outlets[i].name} | {event_string}")
-            await task
+
+        # Prevent deep recursion in cyclic graphs
+        iteration_count = getattr(event, "_cyclic_counter", {}).get(self.name, 0)
+        has_multiple_outlets = len(outlets) > 1
+        use_task_to_prevent_recursion = iteration_count > 2 and not has_multiple_outlets
```
gtopper: Have we communicated the fact that this fix is limited to a single outlet?

tests/test_flow.py
```diff
@@ -5816,6 +5816,66 @@ def close(self):
...
+
+        if loop_count >= 100000:
+            return ["end"]
+        if self.name == "step_6":
+            return ["step_1"]
+        return None
+
+
+def test_maximum_recursion():
+    print("Creating cyclic graph without max_iterations...")
```
gtopper: Normally we don't print what the test does, but rather put it in a comment and/or the test name itself.

tests/test_flow.py
```diff
@@ -5816,6 +5816,66 @@ def close(self):
...
+        awaitable_result.await_result()
+    except RecursionError as e:
+        raise e
+    except RuntimeError as e:
+        if "exceeded the default cycle" in str(e):
+            pass  # Expected error due to cycle without max_iterations
+        else:
+            raise e
+    finally:
+        controller.terminate()
```
gtopper, line 5872: Should add `controller.await_termination()`, as in other tests, to make sure no error is raised.

storey/flow.py
```diff
@@ -400,11 +405,27 @@ async def _do_downstream(self, event, outlets=None, select_outlets: bool = True)
...
+
+        if use_task_to_prevent_recursion:
+            # Create a task to avoid building a deep await chain
+            task = asyncio.get_running_loop().create_task(outlets[0]._do_and_recover(event))
+            tasks.insert(0, task)  # Add to front so we await it first
+        else:
+            # Direct await - ensures errors propagate before any parallel outlets can return
+            await outlets[0]._do_and_recover(event)
```
gtopper: Can be simplified a lot, and save us the need to touch the below code:

```suggestion
        coro = outlets[0]._do_and_recover(event)
        if use_task_to_prevent_recursion:
            coro = asyncio.get_running_loop().create_task(coro)
        await coro
```

gtopper (commented):
Good overall, but can be simplified a bit to reduce complexity and risk (which includes logging bug I identify below).

tests/test_flow.py
```diff
@@ -5816,6 +5816,61 @@ def close(self):
...
+
+    # Create 6 steps WITHOUT max_iterations
+    step1 = Tracer(name="step_1", full_event=False)
+    step2 = Tracer(name="step_2", full_event=False)
+    step3 = Tracer(name="step_3", full_event=False)
+    step4 = Tracer(name="step_4", full_event=False)
+    step5 = Tracer(name="step_5", full_event=False)
+    step6 = Tracer(name="step_6", full_event=False)
+
+    # NO Complete() - just create a pure cycle
```
gtopper, line 5853: There are a lot of tests without `Complete()`, so it's probably not worth mentioning.

tests/test_flow.py
```diff
@@ -5816,6 +5816,61 @@ def close(self):
...
+    source.to(step1)
+    step1.to(step2)
```
gtopper, line 5855: Do we need to test the multiple outlets case?

gtopper (commented):
LGTM otherwise. Two minor comments for consideration.

gtopper (approved):

PR #620, opened 2026-02-26
Replace COPY with INSERT ON CONFLICT for dedup support

alxtkr77:
## Summary

- Replace PostgreSQL COPY protocol with `executemany` + `INSERT ... ON CONFLICT DO NOTHING` in `TimescaleDBTarget`
- When the target table has a UNIQUE constraint, duplicate rows from Kafka rebalance re-delivery are silently dropped at write time
- Tables without UNIQUE constraints are unaffected — `ON CONFLICT DO NOTHING` with no matching constraint behaves as a regular INSERT

## Changes Made

- **`storey/timescaledb_target.py`**: Replace `cur.copy(COPY ... FROM STDIN)` with `cur.executemany(INSERT ... ON CONFLICT DO NOTHING, records)`
- **`integration/test_timescaledb.py`**: Add `test_timescaledb_dedup_with_unique_constraint` — creates table with UNIQUE(endpoint_id, end_infer_time), emits 3 events + 3 duplicates, verifies only 3 rows stored

## Testing

- All 24 integration tests pass (including new dedup test)
- Tested against TimescaleDB on vmdev61

## Reference

- Jira: ML-11979

### DIFF
--- a/storey/timescaledb_target.py
+++ b/storey/timescaledb_target.py
@@ -77,7 +77,7 @@ class TimescaleDBTarget(_Batching, _Writer):
         - The target table must be created as a TimescaleDB hypertable before use
         - The time column should be a timestamp type, preferably TIMESTAMPTZ for timezone awareness
         - Events are written using INSERT ... ON CONFLICT DO NOTHING via executemany,
-          which silently deduplicates rows when the table has a UNIQUE constraint
+          which silently deduplicates rows when the table has a UNIQUE constraint (with some performance penalty)
         - Connection pooling is handled automatically with proper cleanup on termination
         - Built-in retry logic handles deadlocks (fast retry) and connection issues (exponential backoff)
         - Deadlock retries: 3 attempts with 0.1s, 0.2s, 0.4s delays (with jitter)
@@ -133,8 +133,8 @@ class TimescaleDBTarget(_Batching, _Writer):
         # Database connection configuration
         self._dsn = dsn
         # Connection pool: Single connection is sufficient for most use cases since:
-        # 1. COPY operations are already bulk-optimized and very fast
-        # 2. Multiple concurrent COPY operations may cause lock contention
+        # 1. executemany operations are already bulk-optimized and very fast
+        # 2. Multiple concurrent write operations may cause lock contention
         # 3. Most data flows process batches sequentially, not concurrently
         # For high-throughput scenarios with concurrent batches, consider increasing pool size
         self._pool: Optional[AsyncConnectionPool] = None  # Connection pool will be created lazily during first use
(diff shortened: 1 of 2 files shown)


storey/timescaledb_target.py
assaf758: should be updated to executemany

storey/timescaledb_target.py
```diff
@@ -76,7 +76,8 @@ class TimescaleDBTarget(_Batching, _Writer):
     Note:
         - The target table must be created as a TimescaleDB hypertable before use
         - The time column should be a timestamp type, preferably TIMESTAMPTZ for timezone awareness
-        - Events are written using PostgreSQL's COPY protocol for optimal performance
+        - Events are written using INSERT ... ON CONFLICT DO NOTHING via executemany,
+          which silently deduplicates rows when the table has a UNIQUE constraint
```
assaf758:
```suggestion
          which silently deduplicates rows when the table has a UNIQUE constraint (with some performance penalty)
```

storey/timescaledb_target.py
```diff
@@ -354,20 +356,20 @@ async def batch_write_operation():
...
                     table_identifier = sql.Identifier(self._table)
                     if self._schema:
                         table_identifier = sql.Identifier(self._schema, self._table)
 
                     column_identifiers = [sql.Identifier(col) for col in self._column_names]
-                    copy_query = sql.SQL("COPY {} ({}) FROM STDIN").format(
-                        table_identifier, sql.SQL(", ").join(column_identifiers)
+                    placeholders = sql.SQL(", ").join([sql.Placeholder()] * len(self._column_names))
+
+                    insert_query = sql.SQL("INSERT INTO {} ({}) VALUES ({}) ON CONFLICT DO NOTHING").format(
```
assaf758, line 366: Why not specific to ON CONFLICT ON CONSTRAINT UNIQUE so not to silently ignore other errors?

assaf758 (commented):
Nice.
Couple of comments/questions.

storey/timescaledb_target.py
alxtkr77: You're right, will update the comment to reference executemany instead of COPY.

storey/timescaledb_target.py
alxtkr77: Good point, will add.

storey/timescaledb_target.py
alxtkr77: `ON CONFLICT DO NOTHING` only suppresses unique/exclusion constraint violations — other errors (check constraints, not-null, type errors) still raise exceptions. To use `ON CONFLICT ON CONSTRAINT <name>`, we'd need to discover the constraint name at runtime via `pg_constraint`, which adds complexity. Since our use case targets tables with a single unique constraint for dedup, the generic form should be safe. Happy to add constraint-specific handling if you think it's worth the extra query.

assaf758: I see your point, ok - let's start with this.

assaf758 (approved):

PR #621, opened 2026-03-03
[Streaming] Add `StreamCompletion` id field

royischoss:
Small fix for a logging problem. when the event is from the kind of `StreamCompletion` it's missing id field


gtopper (commented):
Great catch! Do you think we can add a regression test for the logging issue? We have some unit tests that test logs, for reference.

gtopper:
Merging so we can include it in the release I'm triggering. Feel free to PR a test if you can find the time for it.

PR #622, opened 2026-03-03
[Streaming] Add _event_string unit tests for `StreamCompletion` event

royischoss:
Adding unit tests for _event_string with `StreamCompletion`

### DIFF
--- a/storey/dtypes.py
+++ b/storey/dtypes.py
@@ -156,6 +156,10 @@ class StreamCompletion:
         error_info = f", error={self.error!r}" if self.error else ""
         return f"StreamCompletion(streaming_step={self.streaming_step!r}, event_id={event_id!r}{error_info})"
 
+    @property
+    def id(self):
+        return self.original_event.id if self.original_event else None
+
 
 class WindowBase:
     def __init__(self, window, period, window_str):
--- a/storey/dtypes.py
+++ b/storey/dtypes.py
@@ -160,10 +160,6 @@ class StreamCompletion:
     def id(self):
         return self.original_event.id if self.original_event else None
 
-    @property
-    def body(self):
-        return self.original_event.body if self.original_event else None
-
 
 class WindowBase:
     def __init__(self, window, period, window_str):
--- a/storey/dtypes.py
+++ b/storey/dtypes.py
@@ -160,6 +160,10 @@ class StreamCompletion:
     def id(self):
         return self.original_event.id if self.original_event else None
 
+    @property
+    def body(self):
+        return self.original_event.body if self.original_event else None
+
 
 class WindowBase:
     def __init__(self, window, period, window_str):
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -320,7 +320,8 @@ class Flow:
             result += f"processing_time={event.processing_time}, "
         if getattr(event, "path", None):
             result += f"path={event.path}, "
-        result += f"body={event.body})" if hasattr(event, "body") else ""
+        if getattr(event, "body", None):
+            result += f"body={event.body})"
         return result
 
     def _should_terminate(self):
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -320,7 +320,7 @@ class Flow:
             result += f"processing_time={event.processing_time}, "
         if getattr(event, "path", None):
             result += f"path={event.path}, "
-        result += f"body={event.body})"
+        result += f"body={event.body})" if hasattr(event, "body") else ""
         return result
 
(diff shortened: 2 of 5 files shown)


tests/test_streaming.py
```diff
@@ -1892,3 +1892,131 @@ def select_outlets(self, event):
...
+        self.logs.append(("warn", args, kwargs))
+
+    def info(self, *args, **kwargs):
+        self.logs.append(("info", args, kwargs))
+
+    def debug(self, *args, **kwargs):
+        self.logs.append(("debug", args, kwargs))
+
+
+class MockContext:
```
gtopper: We already have these in `test_flow.py`, though. Can extract and use to avoid duplication?

tests/test_streaming.py
```diff
@@ -1892,3 +1892,131 @@ def select_outlets(self, event):
...
+        # The logs should contain references to the step names showing flow progression
+        assert "StreamingMap" in all_log_messages or "Collector" in all_log_messages
+
+    def test_event_string_with_stream_completion_no_original_event(self):
+        """Test that _event_string handles StreamCompletion with None original_event.
+
+        Edge case: StreamCompletion.body returns None when original_event is None.
+        The _event_string method should handle this gracefully.
+        """
+        from storey.flow import Flow
```
gtopper: Imports should be at the top unless there's a good reason to put them in a local scope, which isn't the case here. Applies elsewhere as well.

tests/test_streaming.py
```diff
@@ -1892,3 +1892,131 @@ def select_outlets(self, event):
...
+
+        completion = StreamCompletion("streaming_step", None)  # type: ignore[arg-type]
```
gtopper: Should probably not test it with `None` if it's not a valid parameter.

tests/test_streaming.py
```diff
@@ -1892,3 +1892,131 @@ def select_outlets(self, event):
...
+
+        # _event_string should handle StreamCompletion with None original_event without error
+        result = Flow._event_string(completion)
+
+        # Should produce a valid string with body=None
+        assert isinstance(result, str)
+        assert "body=None" in result
+
+    def test_event_string_with_stream_completion_with_error(self):
```
gtopper: Seems like a redundant test, given that the error doesn't seem relevant here, and we already test the same thing in `test_event_string_with_stream_completion`?

storey/dtypes.py
```diff
@@ -160,6 +160,10 @@ def __repr__(self):
     def id(self):
         return self.original_event.id if self.original_event else None
 
+    @property
+    def body(self):
+        return self.original_event.body if self.original_event else None
```
gtopper: This one actually doesn't seem to make much sense... Why should we want `StreamCompletion` to have the original event's body?

gtopper:
Oh, also please fix the title :)

storey/dtypes.py
royischoss: we need to have a body or change the _event_string

tests/test_streaming.py
```diff
@@ -36,6 +37,7 @@
 )
 from storey.dtypes import Event, StreamChunk, StreamCompletion
 from storey.flow import _is_generator
+from tests.test_flow import MockContext, MockLogger
```
gtopper: I think we should avoid importing the massive test_flow, and should instead extract these classes to a helper file.

storey/flow.py
```diff
@@ -320,7 +320,7 @@ def _event_string(event):
             result += f"processing_time={event.processing_time}, "
         if getattr(event, "path", None):
             result += f"path={event.path}, "
-        result += f"body={event.body})"
+        result += f"body={event.body})" if hasattr(event, "body") else ""
```
gtopper: The paren needs to be present either way, though. How about:
```suggestion
    if hasattr(event, "body"):
        result += f"body={event.body}" 
    result += ")"
```

gtopper (commented):
Small bug, minor test housekeeping. LGTM otherwise  🚀

storey/flow.py
```diff
@@ -320,7 +320,8 @@ def _event_string(event):
             result += f"processing_time={event.processing_time}, "
         if getattr(event, "path", None):
             result += f"path={event.path}, "
-        result += f"body={event.body})"
+        if hasattr(event, "body"):
+            result += f"body={event.body})"
```
gtopper: The paren needs to be taken out of the conditional as per my earlier code suggestion.

gtopper (approved):
LGTM ✅

PR #630, opened 2026-03-16
[Batch] Fix race conditions in _Batching async flush/termination lifecycle

alxtkr77:
## Summary
Fix 6 race conditions in `_Batching._emit_batch` and `_emit_all` that cause data loss, `KeyError` crashes, and writes to closed targets when the timer task (`_sleep_and_emit`) and the run loop (`_do`) interleave at `await` points during slow target writes (S3/TSDB).

## Changes Made
- Pop `_batch_events` **before** `await _emit()` so concurrent `_do()` creates a fresh list instead of appending to the in-flight one (fixes KeyError and silent event loss)
- Cancel `_timeout_task` at the start of termination before `_emit_all` (prevents writes to closed targets)
- Replace snapshot-based key iteration in `_emit_all` with `while self._batch` loop (catches keys added during flush)
- Re-insert failed batch data on `_emit` error for retry/Kafka redelivery (prevents permanent data loss)
- Add error handling in `_emit_all` so drain path logs and continues instead of crashing

## Testing
- Added `tests/test_batching_races.py` — 6 deterministic tests reproducing each race condition using `asyncio.Event` gating
- All 6 tests pass

## Reference
- Jira: [ML-12304](https://iguazio.atlassian.net/browse/ML-12304)
- Related: ML-12286 (silent data loss), IG4-1713 (KeyError crash)

[ML-12304]: https://iguazio.atlassian.net/browse/ML-12304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1560,55 +1560,18 @@ class _Batching(Flow):
             return
         batch_time = self._batch_first_event_time.pop(batch_key)
         last_event_time = self._batch_last_event_time.pop(batch_key)
-        batch_start_time = self._batch_start_time.pop(batch_key)
+        self._batch_start_time.pop(batch_key, None)
         # Pop batch_events BEFORE the await so concurrent _do() calls create
         # a fresh list instead of appending to the one we're processing.
         batch_events = self._batch_events.pop(batch_key, [])
-        try:
-            await self._emit(batch_to_emit, batch_key, batch_time, batch_events, last_event_time)
-        except Exception:
-            # Re-insert the failed batch so it can be retried by the next timer
-            # cycle or redelivered by Kafka.  Prepend to any new events that
-            # arrived during the failed _emit.
-            if batch_key in self._batch:
-                self._batch[batch_key] = batch_to_emit + self._batch[batch_key]
-            else:
-                self._batch[batch_key] = batch_to_emit
-            if batch_key in self._batch_events:
-                self._batch_events[batch_key] = batch_events + self._batch_events[batch_key]
-            else:
-                self._batch_events[batch_key] = batch_events
-            if batch_key in self._batch_first_event_time:
-                self._batch_first_event_time[batch_key] = min(batch_time, self._batch_first_event_time[batch_key])
-            else:
-                self._batch_first_event_time[batch_key] = batch_time
-            if batch_key in self._batch_last_event_time:
-                self._batch_last_event_time[batch_key] = max(last_event_time, self._batch_last_event_time[batch_key])
-            else:
-                self._batch_last_event_time[batch_key] = last_event_time
-            self._batch_start_time.setdefault(batch_key, batch_start_time)
-            raise
+        await self._emit(batch_to_emit, batch_key, batch_time, batch_events, last_event_time)
 
     async def _emit_all(self):
         # Loop until empty instead of snapshot iteration, so keys added
         # during a yielding _emit are not missed.
         while self._batch:
             key = next(iter(self._batch.keys()))
-            try:
-                await self._emit_batch(key)
-            except Exception:
-                if self.logger:
-                    self.logger.error(
-                        f"Failed to flush batch for key '{key}' in step '{self.name}' "
-                        f"during termination:\n{traceback.format_exc()}"
-                    )
-                # _emit_batch re-inserted the failed batch.  Remove the batch
-                # data to avoid infinite retry, but keep _batch_events so event
-                # references stay alive and Kafka offsets stay uncommitted.
-                self._batch.pop(key, None)
-                self._batch_first_event_time.pop(key, None)
-                self._batch_last_event_time.pop(key, None)
-                self._batch_start_time.pop(key, None)
+            await self._emit_batch(key)
 
 
 class Batch(_Batching, WithUUID):
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import asyncio
+import contextlib
 import copy
 import datetime
 import enum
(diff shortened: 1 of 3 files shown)


storey/flow.py
```diff
@@ -1540,16 +1548,55 @@ async def _emit_batch(self, batch_key: Optional[str] = None):
...
+        # a fresh list instead of appending to the one we're processing.
+        batch_events = self._batch_events.pop(batch_key, [])
         try:
-            await self._emit(batch_to_emit, batch_key, batch_time, self._batch_events[batch_key], last_event_time)
-        finally:
-            # whether we succeeded or failed, we are done with these events
-            del self._batch_events[batch_key]
+            await self._emit(batch_to_emit, batch_key, batch_time, batch_events, last_event_time)
+        except Exception:
+            # Re-insert the failed batch so it can be retried by the next timer
```
gtopper: This change introduces unlimited retries, which is a major behavior change that would cause infinite retries if the issue is not transient. Some concerns:
* It is probably only appropriate for a stream trigger
* It may be appropriate to make it configurable, as it is in `_ConcurrentJobExecution`
* It doesn't seem to me that `_Batching` as it is will exert backpressure

tests/test_batching_races.py
```diff
@@ -0,0 +1,532 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Tests demonstrating race conditions in storey _Batching.
+
+All code runs on a single asyncio event loop.  Races happen at `await` points
+where the event loop switches to another ready coroutine.  We use an
+asyncio.Event gate inside _emit to deterministically yield control at the
+exact moment needed to trigger each race.
```
gtopper: This description becomes outdated once the fixes are in.

tests/test_batching_races.py
```diff
@@ -0,0 +1,532 @@
...
+
+Tests operate directly on _Batching subclasses, calling _do() to inject events.
+This bypasses the AsyncEmitSource run loop (which processes events one at a time),
```
gtopper: There is no need to bypass `AsyncEmitSource`. It would be better if the tests built a graph instead of exercising the target directly.

storey/flow.py
```diff
@@ -1540,16 +1548,55 @@ async def _emit_batch(self, batch_key: Optional[str] = None):
...
+            # cycle or redelivered by Kafka.  Prepend to any new events that
```
gtopper: This code isn't Kafka-specific, so comments referencing it are probably not good practice.

storey/flow.py
```diff
@@ -1540,16 +1548,55 @@ async def _emit_batch(self, batch_key: Optional[str] = None):
...
+            if batch_key in self._batch_first_event_time:
+                self._batch_first_event_time[batch_key] = min(batch_time, self._batch_first_event_time[batch_key])
+            else:
+                self._batch_first_event_time[batch_key] = batch_time
+            if batch_key in self._batch_last_event_time:
+                self._batch_last_event_time[batch_key] = max(last_event_time, self._batch_last_event_time[batch_key])
+            else:
+                self._batch_last_event_time[batch_key] = last_event_time
+            self._batch_start_time.setdefault(batch_key, batch_start_time)
+            raise
```
gtopper: The batch is reinserted, but it doesn't look like it's being reprocessed anywhere other than in `_emit_all` (on termination)? So it seems like there isn't actually a retry mechanism in place.

tests/test_batching_races.py
```diff
@@ -0,0 +1,532 @@
...
+            await asyncio.sleep(0.05)
+            assert target.emit_count == 1, "Timer should have started _emit"
+
+            # t4-t5: Events 2,3 arrive while _emit is blocked.
+            # Event 2 appends to existing _batch_events[None].
+            await target._do(_ev(2))
+            # t6-t8: Event 3 hits max_events=2 -> _do calls _emit_batch(None).
+            # This also blocks on gate.  Use ensure_future to avoid deadlock.
+            do_task = asyncio.ensure_future(target._do(_ev(3)))
+            await asyncio.sleep(0.01)
```
gtopper: I wonder what this tiny sleep is for. It doesn't seem to be necessary.

tests/test_batching_races.py
```diff
@@ -0,0 +1,532 @@
...
+        async def _test():
+            # DRAIN PATH: error should not crash the flow — it should be
+            # handled gracefully (e.g. logged, batch preserved for redelivery).
+            target1 = FailOnceTarget(fail_on=1, flush_after_seconds=999)
+            target1._init()
+            await target1._do(_ev(1))
+
+            drain_crashed = False
+            try:
+                await target1._do(_termination_obj)
```
gtopper: Please take #604 (discussed in https://iguazio.atlassian.net/browse/ML-12316?focusedCommentId=200229) into account. A crash on termination (e.g. on drain) will not prevent the offsets from being committed.

gtopper (commented):
Important aspect of resilience, but the current changes seem to only help with a limited scenario where a small number of batches failed and we go into termination (e.g. on drain) shortly after.

storey/flow.py
gtopper: * as noted below, this change doesn't _actually_ introduce unlimited retries, but that seems to be the intention.

tests/test_batching_races.py
gtopper: This one is still outstanding.

tests/test_batching_races.py
```diff
@@ -0,0 +1,379 @@
...
+            # With the fix: events should be in SEPARATE lists.
+            # The timer popped _batch_events[None] before await, so event 2
+            # went into a fresh defaultdict list, not the timer's list.
+            assert (
+                len(target._batch_events[None]) == 1
+            ), "Event 2 should be in its own _batch_events[None], not shared with event 1"
+
+            # t9-t10: Release gate -> timer's _emit completes
+            gate.set()
+            await asyncio.sleep(0.01)
```
gtopper: I guess this one can be a yield as well?

gtopper (commented):
Production side LGTM. Would be good to see the tests build a flow as we normally do in unit tests. I don't see a reason that wouldn't work here as well, and it will make the tests less artificial.

assaf758 (approved):

PR #638, opened 2026-03-31
Fix concurrent streaming in `ConcurrentExecution`

gtopper:
[ML-12378](https://iguazio.atlassian.net/browse/ML-12378)

## Problem

When `ConcurrentExecution` processes streaming (generator) event processors with `max_in_flight > 1`, the generators are iterated serially in the FIFO worker coroutine. This blocks the worker from picking up other events until the current generator is fully consumed, negating the concurrency benefit of `max_in_flight`.

## Root cause

`_handle_completed` called `_emit_streaming_chunks` (inherited from `_StreamingStepMixin`), which iterates the generator inline — holding the worker for the entire duration of each generator. With 4 events and a 0.5s generator, total time is ~2.0s instead of ~0.5s.

## Solution

Offload generator iteration to background `asyncio.Task`s that feed an `asyncio.Queue`, freeing the FIFO worker to process other events concurrently.

**New internal types:**
- `_GeneratorDone(error)` — sentinel placed on the queue when the generator is exhausted (or errors).
- `_StreamingQueue(queue, task)` — wraps the queue and its background task for lifecycle management.

**New methods on `ConcurrentExecution`:**
- `_iterate_generator` — background task that consumes a generator (sync or async) into an `asyncio.Queue`. Sync generators use `run_in_executor` per `next()` call to avoid blocking the event loop.
- `_emit_streaming_from_queue` — mirrors `_emit_streaming_chunks` but reads from the queue, emitting `StreamChunk` events and a final `StreamCompletion` (with error if the generator raised).

**Process-based mechanisms** (`process_pool`, `dedicated_process`) use the existing IPC pattern: `_concurrent_streaming_run_in_subprocess` runs the generator in the subprocess and streams chunks via `multiprocessing.Queue`, which `_async_read_streaming_queue` consumes as an async generator on the parent side.

**Task lifecycle:** `_StreamingQueue` holds the `asyncio.Task` reference, and `_handle_completed` awaits it after draining the queue. This prevents GC of fire-and-forget tasks and surfaces unexpected errors.

## Additional changes

- **Migrated `ConcurrentExecution` to `ParallelExecutionMechanisms`**: replaces the old `_supported_concurrency_mechanisms` list with the shared enum. Legacy names (`"threading"`, `"multiprocessing"`) are mapped automatically for backward compatibility.
- **Detects generator functions early** via `inspect.isgeneratorfunction` in `__init__`, used to route process-based mechanisms to the IPC streaming path in `_process_event`.

## Tests

- **Concurrency tests** (`test_concurrent_streaming_asyncio_async_gen`, `test_concurrent_streaming_sync_gen` parametrized over `thread_pool`/`process_pool`/`dedicated_process`/`naive`): verify generators run concurrently across events.
- **Error tests** (`test_concurrent_streaming_error_asyncio_async_gen`, `test_concurrent_streaming_error_sync_gen` parametrized over all mechanisms, `test_concurrent_streaming_error_mixed_with_healthy`): verify mid-stream errors propagate correctly and don't poison healthy concurrent events.

[ML-12378]: https://iguazio.atlassian.net/browse/ML-12378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -76,6 +76,25 @@ def _is_awaitable_coroutine(obj) -> bool:
 _sync_gen_sentinel = object()
 
 
+class _GeneratorDone:
+    """Sentinel placed on a streaming queue when the generator is exhausted."""
+
+    __slots__ = ("error",)
+
+    def __init__(self, error=None):
+        self.error = error
+
+
+class _StreamingQueue:
+    """Wrapper around an asyncio.Queue fed by a background generator consumer."""
+
+    __slots__ = ("queue", "task")
+
+    def __init__(self, queue, task):
+        self.queue = queue
+        self.task = task
+
+
 async def _gen_to_async_gen(sync_gen):
     """Wrap a synchronous generator as an async generator, offloading each
     next() call to a thread so it doesn't block the event loop."""
@@ -1297,10 +1316,15 @@ class ConcurrentExecution(_ConcurrentJobExecution, _StreamingStepMixin):
 
     :param process_event: Function that will be run on each event
 
-    :param concurrency_mechanism: One of:
-      * "asyncio" (default) – for I/O implemented using asyncio
-      * "threading" – for blocking I/O
-      * "multiprocessing" – for processing-intensive tasks
+    :param concurrency_mechanism: Execution mechanism. One of the ``ParallelExecutionMechanisms``
+      values, or a legacy name for backward compatibility:
+
+      * ``"asyncio"`` (default) – for I/O implemented using asyncio
+      * ``"thread_pool"`` (or legacy ``"threading"``) – for blocking I/O
+      * ``"process_pool"`` (or legacy ``"multiprocessing"``) – for processing-intensive tasks
+      * ``"dedicated_process"`` – like process_pool with a single dedicated worker
+      * ``"shared_executor"`` – use an externally supplied executor
+      * ``"naive"`` – run synchronously on the event loop (for trivial work)
 
     :param max_in_flight: Maximum number of events to be processed at a time (default 8)
     :param retries: Maximum number of retries per event (default 0)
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -1323,40 +1350,105 @@ def __init__(
...
+        if self._is_generator_fn and self._concurrency_mechanism in ParallelExecutionMechanisms.process():
+            mp_queue = self._get_mp_manager().Queue()
+            loop.run_in_executor(
+                self._executor,
+                _concurrent_streaming_run_in_subprocess,
+                self._event_processor,
+                args,
+                mp_queue,
+            )
+            ipc_gen = _async_read_streaming_queue(mp_queue)
```
royischoss, line 1444: what is ipc stand for?

tests/test_streaming.py
```diff
@@ -2085,6 +2100,270 @@ def second_stream(collected):
 
         assert result == ["re_test_0", "re_test_1"]
 
+    # -- ML-12378 concurrency tests ----------------------------------------
```
royischoss, line 2103: do we need the task reference in docs and variables name?

royischoss (commented):
Hey LGTM minor comments

storey/flow.py
gtopper: Inter-process communication.

royischoss: maybe it can have better naming ?

tests/test_streaming.py
gtopper: AI slop 🤦‍♂️

tests/test_streaming.py
```diff
@@ -2085,6 +2100,254 @@ def second_stream(collected):
...
```
royischoss, line 2103: still here

storey/flow.py
gtopper: Renamed.

tests/test_streaming.py
gtopper: Right, I left only this reference to the ticket, removed the rest.

royischoss (approved):
LGTM

PR #640, opened 2026-04-19
Fix `SimpleAsyncQueue` missing capacity enforcement

gtopper:
`SimpleAsyncQueue.put()` accepted a `capacity` parameter but never enforced it — items were appended unconditionally, making the queue effectively unbounded and defeating backpressure.

Added `_not_full_futures` to block `put()` when the deque reaches capacity, mirroring the existing `_not_empty_futures` pattern in `get()`. Updated docstring to document why this class exists (Python < 3.12 `asyncio.wait_for` item-swallowing bug, https://github.com/python/cpython/pull/98518).

Updated `test_queue.py` with assertions that `put()` blocks at capacity and resumes after `get()`.

### DIFF
--- a/storey/queue.py
+++ b/storey/queue.py
@@ -76,7 +76,7 @@ class SimpleAsyncQueue:
 
     async def get(self, timeout=None):
         if not self._deque:
-            not_empty_future = asyncio.get_running_loop().create_future()
+            not_empty_future = self._loop.create_future()
             self._not_empty_futures.append(not_empty_future)
             if timeout is None:
                 await not_empty_future
@@ -97,8 +97,9 @@ class SimpleAsyncQueue:
         return result
 
     async def put(self, item):
-        while len(self._deque) >= self._capacity:
-            not_full_future = asyncio.get_running_loop().create_future()
+        assert len(self._deque) <= self._capacity
+        while len(self._deque) == self._capacity:
+            not_full_future = self._loop.create_future()
             self._not_full_futures.append(not_full_future)
             await not_full_future
 
--- a/storey/queue.py
+++ b/storey/queue.py
@@ -60,14 +60,18 @@ def _release_waiter(waiter):
 
 
 class SimpleAsyncQueue:
-    """
-    A simple async queue with built-in timeout.
+    """A bounded async queue with built-in timeout on get().
+
+    Replaces asyncio.Queue + asyncio.wait_for, which can silently swallow
+    items on timeout in Python < 3.12. See
+    https://github.com/python/cpython/pull/98518
     """
 
     def __init__(self, capacity):
         self._capacity = capacity
         self._deque = collections.deque()
         self._not_empty_futures = collections.deque()
+        self._not_full_futures = collections.deque()
         self._loop = asyncio.get_running_loop()
 
     async def get(self, timeout=None):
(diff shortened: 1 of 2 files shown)


storey/queue.py
```diff
@@ -83,16 +87,28 @@ async def get(self, timeout=None):
...
+            not_full_future = self._not_full_futures.popleft()
+            if not not_full_future.done():
+                not_full_future.set_result(True)
+                break
+
         return result
 
     async def put(self, item):
+        while len(self._deque) >= self._capacity:
+            not_full_future = asyncio.get_running_loop().create_future()
```
assaf758: why use asyncio.get_running_loop() when you have self._loop?

storey/queue.py
```diff
@@ -83,16 +87,28 @@ async def get(self, timeout=None):
                     raise TimeoutError(f"Queue get() timed out after {timeout} seconds")
 
         result = self._deque.popleft()
+
+        while self._not_full_futures:
+            not_full_future = self._not_full_futures.popleft()
+            if not not_full_future.done():
+                not_full_future.set_result(True)
```
assaf758: Since we just popped one item from the deque there should be one empty slot in the deque, right? If that's correct i guess we can break the loop after setting one not_full_future.

storey/queue.py
```diff
@@ -83,16 +87,28 @@ async def get(self, timeout=None):
...
+                break
+
         return result
 
     async def put(self, item):
+        while len(self._deque) >= self._capacity:
```
assaf758: 1. why is it a while loop? Shouldn't it be an "if" clause?
2. how can self.deque become > self._capacity? isn't it a bug that we should assert on?

storey/queue.py
gtopper: Can do. This is an existing inconsistency in the code (`asyncio.get_running_loop()` is called in `get()`). I'll update it to use `self._loop` everywhere.

storey/queue.py
gtopper: That's what we do one line below:
```python
if not not_full_future.done():
    not_full_future.set_result(True)
    break
```

storey/queue.py
gtopper: 1. It can be added while we `await` three lines down.
2. It shouldn't. Sure, we can assert on it instead.

gtopper: We discussed this and realized that this could cause unfair scheduling and even starvation. The fix is to introduce direct tracking of the size and only decrement it once `await putter` returns.

storey/queue.py
```diff
@@ -60,39 +60,58 @@ def _release_waiter(waiter):
...
+    https://github.com/python/cpython/pull/98518
     """
 
     def __init__(self, capacity):
         self._capacity = capacity
         self._deque = collections.deque()
-        self._not_empty_futures = collections.deque()
+        self._getters = collections.deque()
+        self._putters = collections.deque()
+        self._size = 0
```
assaf758, line 79: please add comment describing the motiveation for _size on top of len(self._deque)

storey/queue.py
```diff
@@ -60,39 +60,58 @@ def _release_waiter(waiter):
...
+
         return result
 
     async def put(self, item):
-        while self._not_empty_futures:
-            not_empty_future = self._not_empty_futures.popleft()
-            if not not_empty_future.done():
-                not_empty_future.set_result(True)
-                break
+        if self._size >= self._capacity:
```
assaf758:
```suggestion
        assert 0 <= self._size <= self._capacity
        if self._size == self._capacity:
```

assaf758 (commented):
looks good - minor comments

storey/queue.py
gtopper: Done.

storey/queue.py
gtopper: Done.

assaf758 (approved):

PR #641, opened 2026-05-05
Add OTelMetricsExporter step for OTLP metrics export

davesh0812:
## Summary

- Adds `OTelMetricsExporter`, a pass-through pipeline step that exports user-defined domain metrics via OTLP gRPC as a side-effect, forwarding events downstream unchanged
- Supports all 4 OTel synchronous instrument types: `gauge`, `counter`, `updown_counter`, `histogram`; both `periodic` and `immediate` flush modes; single and multi-metric events; configurable field mapping
- Install with `pip install storey[otel]`

## Design notes

**Not a target — pass-through only.**
`OTelMetricsExporter` is a mid-pipeline step, not a terminal target. Every event is forwarded downstream unchanged after the metric is recorded. This allows metrics export to be composed with any downstream step (e.g. `ParquetTarget`, `Map`, etc.).

**No ACK / delivery guarantee — by design.**
Export is best-effort, matching the standard OTel contract. The OTel SDK handles batching, retries, and delivery to the collector internally. A failed export cycle does not block, error, or retry the event in the Storey pipeline. This is consistent with how every other OTel SDK integration works — the pipeline should not be coupled to metrics delivery.

## Test plan

- [ ] Unit tests: `pytest tests/test_otel_metrics_exporter.py` (39 tests, mock OTLP exporter)
- [ ] Integration tests: `pytest integration/test_otel_metrics_exporter.py` (15 tests, embedded in-process gRPC receiver, no Docker required)
- [ ] All 4 instrument types × both flush modes covered
- [ ] Mixed-type multi-metric events covered

🤖 Generated with [Claude Code](https://claude.com/claude-code)


integration/test_otel_metrics_exporter.py
```diff
@@ -0,0 +1,323 @@
...
+
+pytest.importorskip("opentelemetry.sdk.metrics")
+
+from opentelemetry.proto.collector.metrics.v1 import (  # noqa: E402
+    metrics_service_pb2,
+    metrics_service_pb2_grpc,
+)
+
+from storey import AsyncEmitSource, Event, build_flow  # noqa: E402
+from storey.otel_metrics_exporter import OTelMetricsExporter  # noqa: E402
```
royischoss, line 33: please check if all # noqa are needed

storey/otel_metrics_exporter.py
```diff
@@ -0,0 +1,242 @@
...
+        else:  # histogram
+            inst = self._meter.create_histogram(name=name)
+        self._instruments[name] = (inst, instrument_type)
+
+    async def _lazy_init(self):
+        if self._provider is not None:
+            return
+        if OTLPMetricExporter is None:
+            raise ImportError("Install with: pip install storey[otel]")
+        interval = sys.maxsize if self._flush_mode == "immediate" else self._export_interval_millis
```
royischoss: sys.maxsize as interval for immediate mode — using an astronomically large interval to suppress periodic exports is a hack. The OTel SDK provides ManualReader precisely for this case and makes the intent explicit

royischoss: from opentelemetry.sdk.metrics.export import ManualMetricReader                                                                                                                                                                       
                                                                                                                                                                                                                                      
  if self._flush_mode == "immediate":                                                                                                                                                                                                   
      reader = ManualMetricReader(exporter)
  else:                                                                                                                                                                                                                                 
      reader = PeriodicExportingMetricReader(exporter, export_interval_millis=self._export_interval_millis)                                                                                                                           
                                                                                                                                                                                                                                        
  - immediate mode: force_flush per event is expensive — calling force_flush(timeout_millis=5_000) on every single event will be a throughput killer. The docstring should warn that immediate mode is intended for low-volume /        
  debugging scenarios, and periodic is the recommended default for production.                                                                                                                                                          
  - Missing required field raises KeyError, not ValueError — if metric_name_field or value_field is absent from the event body, the user gets a raw KeyError. A ValueError with a clear message about which field is missing would be   
  more helpful.

storey/otel_metrics_exporter.py
```diff
@@ -0,0 +1,242 @@
...
+    PeriodicExportingMetricReader = None
+
+
+def _flush_and_shutdown(provider) -> None:
+    """Blocking flush + shutdown — always run via run_in_executor."""
+    try:
+        provider.force_flush(timeout_millis=10_000)
+        provider.shutdown(timeout_millis=30_000)
+    except Exception:
+        pass
```
royischoss: maybe it better add a log trace here

royischoss (commented):
hey small comments

integration/test_otel_metrics_exporter.py
davesh0812: They're all needed. The pytest.importorskip("opentelemetry.sdk.metrics") on line 25 forces the OTel and storey imports below it, which trips flake8 E402 (module-level import not at top of file). Without importorskip first, the tests would error at collection time when OTel isn't installed instead of skipping cleanly.

storey/otel_metrics_exporter.py
davesh0812: There is no ManualMetricReader I used math.inf like they metntion on :
  - OpenTelemetry Python SDK metrics.export docs (https://opentelemetry-python.readthedocs.io/en/latest/sdk/metrics.export.html)
  - opentelemetry-python v1.41.1 source (https://github.com/open-telemetry/opentelemetry-python/blob/v1.41.1/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/export/__init__.py)

royischoss (approved):
LGTM 🔥

PR #642, opened 2026-06-18
ML-12776: Route failing recovery steps to the error stream instead of poisoning the source

davesh0812:
## What

When a step has a recovery step (error handler) and that error handler **itself raises**, `Flow._do_and_recover` invoked the recovery step via `recovery_step._do(event)` and let the exception propagate raw. On an explicit-ack source this latched `self._ex`, **permanently poisoning the flow**: only the failing event's offset was committed, and every subsequent event was rejected at the emit gate and never processed or committed (the stream stalls).

## Fix

Wrap the recovery-step call so a failing error handler is routed through the same error-stream-or-raise handling as any other unrecovered error, extracted into a shared `_handle_unrecovered_error` helper (no duplication).

The recovery step is still invoked via `_do` (not `_do_and_recover`) and its failure is handled inline rather than re-entering recovery — so a recovery cycle (e.g. a graph-wide handler that is its own recovery step) **cannot loop infinitely**.

Note: failed events are committed only when the context exposes `push_error` (the mlrun/nuclio setup).

## Tests

Added `tests/test_flow.py::test_async_offset_commit_with_failing_step`, parametrized over a plain failing step and a failing recovery step, asserting all events are dead-lettered and committed.

## Jira

ML-12776 (fix version 1.12.0)

### DIFF
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -341,7 +341,7 @@ class Flow:
             event.error = ex
             try:
                 return await recovery_step._do(event)
-            except BaseException as recovery_ex:
+            except Exception as recovery_ex:
                 if getattr(recovery_ex, "_raised_by_storey_step", None) is None:
                     recovery_ex._raised_by_storey_step = recovery_step
                 return await self._handle_unrecovered_error(event, recovery_ex)
@@ -1220,11 +1220,22 @@ class _ConcurrentJobExecution(Flow):
                     ex._raised_by_storey_step = self
                     recovery_step = self._get_recovery_step(ex)
                     try:
+                        recovered = False
                         if recovery_step is not None:
                             event.origin_state = self.name
                             event.error = ex
-                            await recovery_step._do(event)
-                        else:
+                            try:
+                                await recovery_step._do(event)
+                                recovered = True
+                            except Exception as recovery_ex:
+                                # The recovery step (error handler) itself failed. Fall through to the
+                                # error-stream/raise handling below rather than propagating raw, which
+                                # would poison an explicit-ack source. recovery_step._do (not
+                                # _do_and_recover) is used to avoid infinite recovery loops.
+                                if getattr(recovery_ex, "_raised_by_storey_step", None) is None:
+                                    recovery_ex._raised_by_storey_step = recovery_step
+                                ex = recovery_ex
+                        if not recovered:
                             if event._awaitable_result:
                                 none_or_coroutine = event._awaitable_result._set_error(ex)
                                 if none_or_coroutine:
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -1221,6 +1221,7 @@ class _ConcurrentJobExecution(Flow):
                     recovery_step = self._get_recovery_step(ex)
                     try:
                         recovered = False
+                        recovery_message = None
                         if recovery_step is not None:
                             event.origin_state = self.name
                             event.error = ex
@@ -1235,13 +1236,17 @@ class _ConcurrentJobExecution(Flow):
                                 if getattr(recovery_ex, "_raised_by_storey_step", None) is None:
                                     recovery_ex._raised_by_storey_step = recovery_step
                                 ex = recovery_ex
+                                # Capture while recovery_ex is still the active exception; once this
+                                # except block exits, sys.exc_info() reverts to the outer ex, so a
+                                # later traceback.format_exc() would format the wrong traceback.
+                                recovery_message = traceback.format_exc()
                         if not recovered:
                             if event._awaitable_result:
                                 none_or_coroutine = event._awaitable_result._set_error(ex)
                                 if none_or_coroutine:
                                     await none_or_coroutine
                             if self.context and hasattr(self.context, "push_error"):
-                                message = traceback.format_exc()
+                                message = recovery_message if recovery_message is not None else traceback.format_exc()
                                 if self.logger:
                                     self.logger.error(f"Pushing error to error stream: {ex}\n{message}")
                                 self.context.push_error(event, f"{ex}\n{message}", source=self.name)
--- a/storey/flow.py
+++ b/storey/flow.py
@@ -336,21 +336,28 @@ class Flow:
             ex._raised_by_storey_step = self
             recovery_step = self._get_recovery_step(ex)
             if recovery_step is None:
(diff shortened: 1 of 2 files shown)


storey/flow.py
```diff
@@ -336,21 +336,28 @@ async def _do_and_recover(self, event):
...
-                    return
-                else:
-                    raise ex
+                return await self._handle_unrecovered_error(event, ex)
             event.origin_state = self.name
             event.error = ex
-            return await recovery_step._do(event)
+            try:
+                return await recovery_step._do(event)
+            except BaseException as recovery_ex:
```
royischoss: Minor — BaseException catch can swallow CancelledError (storey/flow.py:354)                                                                                                                                    
  The new except BaseException as recovery_ex will catch asyncio.CancelledError, KeyboardInterrupt, SystemExit. The existing outer handler does the same, so this is consistent with surrounding code rather than
   a new defect, but worth being aware of — if a recovery step is mid-await when the task is cancelled, the cancellation gets dead-lettered to the error stream instead of propagating. If you want to be strict,
   narrow to Exception here. Not blocking.

tests/test_flow.py
```diff
@@ -628,6 +628,63 @@ def test_async_offset_commit_error_on_termination():
...
+        event.offset = offset
+        await controller.emit(event)
+    del event
+
+    await asyncio.sleep(2)
+
+    try:
+        await controller.terminate(wait=True)
+    except Exception:
+        pass  # flow is poisoned today; the contract we're asserting is about offsets
```
royischoss: After this fix lands, terminate should succeed cleanly. The comment ("poisoned today") reads as if the bug were still present. Either drop the try/except (and let a regression surface as a failure) or update
   the comment to "defensive — flow should no longer be poisoned, but we don't make terminate behavior part of this test's contract."

royischoss (commented):
Looks good two small comments

storey/flow.py
davesh0812: Done in 210c266 — narrowed the catch to `except Exception`, so `CancelledError`/`KeyboardInterrupt`/`SystemExit` now propagate instead of being dead-lettered. Also applied the same try/except + narrowing to the concurrent path (`_ConcurrentJobExecution._worker`), which had the same un-wrapped failing-recovery bug, with a new regression test.

tests/test_flow.py
davesh0812: Done in 210c266 — dropped the try/except entirely so a regression would surface as a failure, and removed the stale "poisoned today" comment (the flow terminates cleanly after the fix).

storey/flow.py
```diff
@@ -1213,11 +1220,22 @@ async def _worker(self):
...
+                                # error-stream/raise handling below rather than propagating raw, which
+                                # would poison an explicit-ack source. recovery_step._do (not
+                                # _do_and_recover) is used to avoid infinite recovery loops.
+                                if getattr(recovery_ex, "_raised_by_storey_step", None) is None:
+                                    recovery_ex._raised_by_storey_step = recovery_step
+                                ex = recovery_ex
+                        if not recovered:
                             if event._awaitable_result:
                                 none_or_coroutine = event._awaitable_result._set_error(ex)
                                 if none_or_coroutine:
```
royischoss, line 1246: `traceback.format_exc()` at line 1244 below returns the **original** exception's traceback, not the recovery step's. By the time control reaches that line, this inner `except Exception as recovery_ex:` block has already exited and Python has restored `sys.exc_info()` to the outer `ex` — but the local `ex` variable was reassigned to `recovery_ex`. Result: the dead-letter payload and the log line both read `f"<recovery_ex>\n<original traceback>"`, which is misleading when debugging a failing error handler.

The cleanest fix is to reuse the `_handle_unrecovered_error` helper added in this same PR (line 349) — call it from *inside* the inner except (where `sys.exc_info()` is still `recovery_ex`), and from the no-recovery branch. That mirrors `_do_and_recover` exactly, removes the `recovered` flag, and lets the awaitable-result + push_error + raise logic live in one place.

```suggestion
                        if recovery_step is not None:
                            event.origin_state = self.name
                            event.error = ex
                            try:
                                await recovery_step._do(event)
                            except Exception as recovery_ex:
                                # The recovery step (error handler) itself failed. Route it through
                                # _handle_unrecovered_error rather than propagating raw, which would
                                # poison an explicit-ack source. recovery_step._do (not
                                # _do_and_recover) is used to avoid infinite recovery loops.
                                if getattr(recovery_ex, "_raised_by_storey_step", None) is None:
                                    recovery_ex._raised_by_storey_step = recovery_step
                                await self._handle_unrecovered_error(event, recovery_ex)
                        else:
                            await self._handle_unrecovered_error(event, ex)
```

Applying this suggestion (bounded by the diff hunk) leaves lines ~1242–1249 below as dead code — please delete them manually as a followup so the worker no longer carries an inlined copy of the push_error/raise logic:
```python
                            if event._awaitable_result:
                                none_or_coroutine = event._awaitable_result._set_error(ex)
                                if none_or_coroutine:
                                    await none_or_coroutine
                            if self.context and hasattr(self.context, "push_error"):
                                message = traceback.format_exc()
                                if self.logger:
                                    self.logger.error(f"Pushing error to error stream: {ex}\n{message}")
                                self.context.push_error(event, f"{ex}\n{message}", source=self.name)
                            else:
                                raise ex
```

Quick repro of the underlying Python semantics that motivates the original concern:
```python
try:
    raise ValueError("outer")
except Exception:
    try:
        raise RuntimeError("inner")
    except Exception:
        pass
    print(traceback.format_exc())   # prints ValueError, not RuntimeError
```

Worth also extending `test_concurrent_execution_failing_recovery_step` to assert the captured message contains `"error handler failed"` so a regression on this is caught.

davesh0812: Great catch — fixed in bf7ed26. The implicit exception chaining made this subtle: the recovery step raises while the original exception is still being handled, so the late `format_exc()` returned the original traceback. Now `recovery_message = traceback.format_exc()` is captured inside the inner `except` (while `recovery_ex` is active) and used for the dead-letter payload/log. Also extended `test_concurrent_execution_failing_recovery_step` to assert the payload contains `RuntimeError: error handler failed` — verified it fails against the old late-capture and passes with the fix.

royischoss (approved):
LGTM

